PR #1, opened 2017-01-26
Break the main source file up into subfiles

mwachs5:
This doesn't change any content, but I feel like it will be easier to work with this document if the sections are in different files.


timsifive:
Personally I like one-big-file, because I can more easily navigate the document.

But if you get somebody else to comment that they prefer many files I'll happily accept this change.

mwachs5:
I don't feel too strongly about this and it's way out of date now. Closing.

PR #4, opened 2017-01-26
Unalias 'access' registers and fields for clarity

mwachs5:
I found the many uses of 'access' in the spec confusing. I renamed the bus-related registers and fields to "sb..." and renamed the register "access" to "accesscs" since that seemed to be the pattern for "control and status" registers.

I fixed a couple typos as well.

### DIFF
--- a/dm1_registers.xml
+++ b/dm1_registers.xml
@@ -58,9 +58,9 @@
         </field>
     </register>
 
-    <register name="Bus Access Control" short="buscs" address="0x02">
+    <register name="Bus Access Control" short="sbcs" address="0x02">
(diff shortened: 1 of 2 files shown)


dm1_registers.xml
```diff
@@ -58,9 +58,9 @@
         </field>
     </register>
 
-    <register name="Bus Access Control" short="buscs" address="0x02">
+    <register name="Bus Access Control" short="sbcs" address="0x02">
```
timsifive: Does sbcs stand for System Bus Control/Status? If so, can you fix the long name (or at least add System to the front)?

PR #5, opened 2017-01-29
reset: Add some comments

mwachs5:
I tried to explain why I find the reset part of the spec vague/underspecified. Most of this PR is comments/questions, so I don't expect it to be merged without some discussion.


riscv-debug-spec.tex
```diff
@@ -619,6 +642,34 @@ \subsection{JTAG Connector}
...
+% then tell the debugger to pull the 'RESET' signal.
+% The desired effect woudld be to be able to debug the
+% code that executes on external reset.
+%
+% Another common workflow would be to have been debugging for a while,
+% then you disconnect the debugger.
+% Another user connects, and has no idea what state the debug logic
+% is in. They could assert the TRSTn signal (or clock JTAG TAP into
+% the proper state) and then know that all the Debug Transport Module
+% (and Debug Module logic...?) is in a known state.
```
timsifive: I don't know how common this is. If the user doesn't know the debug state, do they know the CPU state? I would think they will want to reset everything. Perhaps we need a separate way to reset the Debug Module. (We can't rely on something JTAG-specific like a Test-Logic-Reset state in any case, although we could spec a DTM->DM reset signal.)

riscv-debug-spec.tex
```diff
@@ -358,8 +358,31 @@ \subsection{Debug Module Interface (DMI)}
...
+% somehow this debugger-driven reset signal can't impact them.
+% But I don't think that should be spelled out case-by-case in this spec,
+% so I want to focus on the actual requirement more than saying exactly
+% what this reset should do.
+%
+% Also, based on this, is it necessary to say something like Debug Module
+% registers are ONLY reset based on Power-On-Reset or explicit
+% Test-Logic-Reset, similar to JTAG
+% specification? Otherwise how do you debug what happens on e.g.
+% external reset (when someone hits reset button).
```
mwachs5: makes sense. I agree.

riscv-debug-spec.tex
```diff
@@ -619,6 +642,34 @@ \subsection{JTAG Connector}
...
 
+%MAW: Why do we only have 'RESET', but no 'TRST(n)' signal?
+% See http://openocd.org/doc/html/Reset-Configuration.html
+% Both are useful.
+% One is explicitly ``test logic reset'', which would imply
+% resetting the JTAG state machine and all associated test logic, in this
+% case the Debug Transport Module, synchronizers, and perhaps even
+% the Debug Module itself. It basically does the same thing as getting
+% the JTAG state machine into TEST_LOGIC_RESET state, which is
+% why it is optional.
```
mwachs5: Because some circuits may not have a POR circuit to actually reset the test logic.... so if I don't hook up a debugger, somehow that logic has to be reset (the JTAG standard says you can't use the system reset for this).

riscv-debug-spec.tex
```diff
@@ -619,6 +642,34 @@ \subsection{JTAG Connector}
...
+%
+% The other is a convenience which allows the debugger to essentially
+% hit the system reset button, which is basically mutually exclusive from
+% what TRST resets.
```
mwachs5: I agree. You definitely want this signal. I'm just wondering why you wouldn't want TRST as well, or at least to say where it would be on the header if someone wants to include it.

riscv-debug-spec.tex
mwachs5: I guess I would have to think about whether there is anything in the Debug Module that needs to be reset that couldn't just be done by writing to the registers. 

For example, say the debug module requested a mini debug program to execute, but this program gets stuck for some reason. We reset the core using the JTAG debugger's "RESET" pin, which doesn't reset the Debug Module(?) Now, we know the core is never going to finish the mini program execution, but Debug Module is sitting there waiting for it to do so.

I think my real question is just how to reset the Debug Module, if it's not reset with Test Logic Reset or System Reset.

mwachs5: Maybe we just need a control bit in the Debug Module to reset its registers and any state (DTM->DM reset signal would be OK too, but we don't really have a concept of signals like that in this spec). Writing 1 to it from any transport would reset the Debug Module. Then a common debugger use case may be to assert RESET, assert Debug Module Reset, deassert RESET, deassert Debug Module Reset, now everything is in known state.

Actually, this could have other uses.... it could be that the Debug Module is in an inactive (reset) by default on power-on, then only when a debugger writes to it to "activate" it is it active. Knowing that it is "active" would be useful. For example, you could debug across a sleep routine by preventing actual power gating when debug is active. This would allow you to keep the Debug Module on the same power region as the components it can debug.

riscv-debug-spec.tex
timsifive: Pins are expensive, and this one doesn't seem essential.
We copied the Cortex Debug Connector because it's compact and few pins. It doesn't have a TAP reset signal either. 1149.1 explicitly says TRST is an optional signal.

> TRST* is required when the test logic does not power-up in a known and controlled state, either because of the nature of the underlying technology or because there is no on-chip power-up reset generator.

That implies POR is fine for those circuits.

riscv-debug-spec.tex
timsifive: I reset bit like that makes a lot of sense. Although the DM must be reset during power-on to avoid accidentally interfering with regular execution. (At the least, the halt/reset signals it controls must be 0 on power on.)

riscv-debug-spec.tex
mwachs5: In this case, isn't ARM the one making all the circuits covered by the spec? So it's fine for them to say you don't need TRST.

timsifive: That quote was from IEEE 1149.1TM-2013

mwachs5: Right... when I read that sentence it tells me as a chip designer that I am allowed to require a TRST signal. Not that as a hardware debugger builder, I can choose to leave TRST off.

riscv-debug-spec.tex
mwachs5: agreed on those points

riscv-debug-spec.tex
timsifive: I agree that we could require it, but I don't know why we'd want to.

Earlier you made an argument based on the JTAG spec saying only TRST can reset the TAP (or at least that's what I understood). That doesn't sound right because clearly the spec says TRST is optional.

Does it make more sense to continue this discussion in e-mail? It's getting a bit confusing having to go through the pull request to see which thread has new comments.

dm1_registers.xml
```diff
@@ -15,7 +15,14 @@
             system. To perform a reset the debugger writes 1, and then writes 0
             to deassert the reset.
         </field>
-        <field name="0" bits="29:26" access="R" reset="0" />
+        <field name="dmactive" bits="29" access="R/W" reset="0">
+          This bit indicates whether the Debug Module is active. When 0,
+	  all DM registers and state keep their reset values.
+	  DM registers should not be otherwise reset after power-on,
+	  neither by the platform's system reset or Debug Transport
+	  reset signals.
```
timsifive: And when 1?

Also, can you keep the indents in this file consistent?

riscv-debug-spec.tex
```diff
@@ -608,6 +614,13 @@ \subsection{JTAG Connector}
 connector looks as it does in Table~\ref{tab:header}. The function of each pin
 is described in Table~\ref{tab:pinout}.
 
+This header does not include the optional JTAG TRST signal. If an implementation
+requires TRST to be driven at Power-On, this must be handled at the
+PCB level and is not exported to this header.
+
+TODO: If the above is too harsh of a requirement, we should pick a different header to
```
timsifive: Can you start a conversation about this on the mailing list?

dm1_registers.xml
mwachs5: Sorry about the whitespaces... you merged the last one before I fixed it... will see what's up with my editor. No need to merge this yet.

riscv-debug-spec.tex
mwachs5: email sent

PR #6, opened 2017-01-29
system_bus: be consistent and always call it 'System Bus'

mwachs5:
Even if some dislike the implications of the name, we should be consistent and clear in the spec. And it's easier to find-and-replace if we choose a different name.

### DIFF
--- a/dm1_registers.xml
+++ b/dm1_registers.xml
@@ -120,9 +120,9 @@
         <field name="halt1023:992" bits="31" access="R" reset="0" />
     </register>
 
-    <!-- =============== bus mastering =============== -->
+    <!-- =============== system bus mastering =============== -->
 
-    <register name="Bus Access Control" short="sbcs" address="0x03">
+    <register name="System Bus Access Control" short="sbcs" address="0x03">
         <field name="0" bits="31:20" access="R" reset="0" />
         <field name="sbaccess" bits="19:17" access="R/W" reset="2">
             Select the access size to use for system bus accesses triggered by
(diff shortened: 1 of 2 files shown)


dm1_registers.xml
```diff
@@ -138,23 +138,27 @@
 
 	    4: 128-bit
 
-            If an unsupported bus access size is written here, the DM may not
+            If an unsupported system bus access size is written here,
+	    the DM may not
```
timsifive: Awkward indent.

dm1_registers.xml
```diff
@@ -138,23 +138,27 @@
...
             perform the access, or may perform the access with any access size
         </field>
-        <field name="autoincrement" bits="16" access="R/W" reset="0">
-            When 1, the internal address value (used by the bus master) is
-            incremented by the access size (in bytes) selected in \Fsbaccess after
-            every bus access.
+        <field name="sbautoincrement" bits="16" access="R/W" reset="0">
+          When 1, the internal address value (used by the system bus master)
+	  is
```
timsifive: Ditto.

dm1_registers.xml
```diff
@@ -164,37 +168,39 @@
...
         </field>
     </register>
 
     <register name="System Bus Address 31:0" short="sbaddress0" address="0x04">
-        If \Fabussize is 0, then this register is not present.
+        If \Fsbasize is 0, then this register is not present.
 
-        When the bus master is busy, writes to this register will return error
+        When the syste bus master is busy,
+	writes to this register will return error
```
timsifive: There are lots of these. I'm not commenting on every one.

PR #9, opened 2017-02-02
Add prehalt and postresume to reg command.

timsifive:
This should allow for at least some basic quick access to read/write CSRs. With a larger program buffer and some scratch RAM it might be possible to do almost anything.

### DIFF
--- a/abstract_commands.xml
+++ b/abstract_commands.xml
@@ -5,8 +5,26 @@
         is accessible, then \Fsize matching the register's actual size must be
         supported.
 
+        \begin{steps}{Perform the following sequence of operations:}
+        \item Halt the hart, if \Fprehalt is set.
+        \item Execute the Program Buffer, if \Fpreexec is set.
+        \item Copy data from the specified register into {\tt data}, if \Fwrite is clear.
+        \item Copy data from {\tt data} into the specified register, if \Fwrite is set.
+        \item Execute the Program Buffer, if \Fpostexec is set.
+        \item Resume the hart, if \Fpostresume is set.
+        \end{steps}
+
         <field name="0" bits="31:24" />
-        <field name="0" bits="23:22" />
+        <field name="prehalt" bits="23">
+            When 1, halt the hart before performing the rest of the command.
+
+            If the hart is already halted, the command may fail.
+        </field>
+        <field name="postresume" bits="22">
+            When 1, resume the hart after performing the rest of the command.
+
+            If the hart is already running, the command may fail.
+        </field>
         <field name="size" bits="21:19">
             2: Access the lowest 32 bits of the register.
 


abstract_commands.xml
```diff
@@ -5,8 +5,26 @@
         is accessible, then \Fsize matching the register's actual size must be
         supported.
 
+        \begin{steps}{Perform the following sequence of operations:}
```
mwachs5: Question about these bits in general... is Debug Module required to implement all these? Or is that an option that can be discovered (e.g. by writing all 1's and seeing what sticks?)

abstract_commands.xml
```diff
@@ -5,8 +5,26 @@
...
+        \item Copy data from the specified register into {\tt data}, if \Fwrite is clear.
+        \item Copy data from {\tt data} into the specified register, if \Fwrite is set.
+        \item Execute the Program Buffer, if \Fpostexec is set.
+        \item Resume the hart, if \Fpostresume is set.
+        \end{steps}
+
         <field name="0" bits="31:24" />
-        <field name="0" bits="23:22" />
+        <field name="prehalt" bits="23">
+            When 1, halt the hart before performing the rest of the command.
```
mwachs5: Ah, this describes something different than what I understood. So actually what this does is very quickly, under the direction of the debugger, snatch something from the hart.

So this doesn't cover another possible case of interest, where a hart hits a breakpoint and quickly dumps a value to the debugger, or jumps to some other monitor code? Is that case covered only by usual breakpoint interrupt code?

abstract_commands.xml
timsifive: I would like them to always be supported, unless it's expensive.
(If they are always implemented, then perhaps it makes sense to make \Fhalt in \Rdmcontrol read-only.)

abstract_commands.xml
timsifive: Yes. This allows quickly accessing the hart. The use case is eg. PC sampling, or doing some kind of debugging on systems that shouldn't be halted because they have real-time requirements (but not so hard that they can't spare a hundred cycles here and there).

Quickly doing breakpoint->code->resume is an interesting idea, but I'm not sure how to implement it. Eg. what happens if all your harts hit the breakpoint at the same time? Making the DM handle them all in sequence starts to take a lot of time, and get a lot more complicated. What if the breakpoint hits again in the first hart when you're still going through the list? Can the debugger use the Program Buffer while something like this is set up? Maybe it's sufficient to handle those corner cases poorly. My inclination is to require the software to be instrumented if you want something like that. The serial ports could be used to get data out.

mwachs5: Ok. For some reason I thought we had this in the previous 0.11 implementation, but if not, simplicity is good. It seems the debugger could basically do something like this by using normal breakpoint exceptions and normal code, which is then basically the debug monitor situation. 

I need to think about whether this is complicated. I keep thinking that this command state machine is really complicated, but then I realize that it is basically what our ROM code was doing before, and all of it can be done in code vs. hardware (as desired) and I get less freaked out.

timsifive:
I don't even see it as a complicated state machine. It's basically linearly
moving from one step to the next. Then depending on how you structure it
you have side states to take actions, or the actions are just conditional
in these linear states. There's one way out to an error condition, but
otherwise I think the transitions are all straightforward, using signals
that exist anyway.

But it's been a long time since I did any state machines and I could be
missing something.

Tim

mwachs5:
It's not complicated, but it is a state machine, vs something that just dumbly does the single action you ask it to do. If it takes a while for a hart to halt, for example, you now have to introduce hardware that waits for that, vs. just reporting whether the hart is halted or not to the debugger, which is smart enough to understand what it can do once the hart is halted.

 I don't think this is really more complicated than the other 'preexec' stuff, but I think I should defer to the other people who are implementing this abstract interface in some other way than having processor execute code (because that's a cheap and easy way of moving through a state machine).

timsifive:
That's fair. Should I merge this into the main spec, or wait until next
week to solicit feedback, after more people have read the core spec?

Tim

mwachs5:
I don't know... seems like you should put it in otherwise no one would be able to comment on it, but maybe you should ask for feedback on the mailing list as well.

I'll update my picture branch.

What happens if `postresume` is set and there is an error? Does the thread stall or resume?

timsifive:
I've just pushed this into the main branch.

I decided that any error just aborts the whole operation. That's more
straightforward than some special behavior. It does mean that if something
goes wrong that you're interrupting the target code when you don't intend
to.

Tim

mwachs5:
Hmm I can imagine different scenarios you'd want error to do different things.

Anyway maybe "error behavior" should be a different high-level topic.

abstract_commands.xml
```diff
@@ -5,8 +5,26 @@
...
+
+            If the hart is already halted, the command may fail.
+        </field>
+        <field name="postresume" bits="22">
+            When 1, resume the hart after performing the rest of the command.
+
+            If the hart is already running, the command may fail.
```
mwachs5: Actually this doesn't make any sense. It seems that in the common case both prehalt and postresume would be set. So the hart would be running already in most situations where you set this, right?

timsifive:
On Fri, Feb 3, 2017 at 9:47 AM, Megan Wachs <notifications@github.com>
wrote:

> Hmm I can imagine different scenarios you'd want error to do different
> things.
>
So can I, but how much flexibility do we really need? As you pointed out,
the state machine is already getting larger.

mwachs5:
Right. I'd rather figure out whether people are OK with this whole state machine thing at all, then we can discuss errors seperately.

timsifive:
I attempted to make things more clear in
https://github.com/sifive/riscv-debug-spec/commit/e18a68d68ed060c9845e9a04019037875133f2c3

mwachs5:
This has been changed to a seperate "Quick Access" command

PR #12, opened 2017-02-07
Typos/errors in abstract command

mwachs5:
I fixed a typo and modified the error case for postresume because it doesn't make sense. The cases are:

-> Hart is running, but you've set both prehalt and postresume. That seems like a valid use case.
-> Hart was halted, but you're using this mechanism to get it out of halt. 
-> Hart is running, but you didn't set prehalt and you did set postresume. This seems like the only error case.

### DIFF
--- a/abstract_commands.xml
+++ b/abstract_commands.xml
@@ -7,7 +7,7 @@
         \item Copy data from the register specified by \Fregno into {\tt data}, if \Fwrite is clear.
         \item Copy data from {\tt data} into the register specified by \Fregno, if \Fwrite is set.
         \item Execute the Program Buffer, if \Fpostexec is set.
-        \item Resume the hart, if \Fpostresume is set. If the hart is already 
+        \item Resume the hart, if \Fpostresume is set. If the hart is already
         running and \Fprehalt is not set, the entire command fails.
         \end{steps}
 
--- a/abstract_commands.xml
+++ b/abstract_commands.xml
@@ -2,13 +2,13 @@
     <register name="Access Register">
         \begin{steps}{Perform the following sequence of operations:}
         \item Halt the hart, if \Fprehalt is set. If the hart is already
-            handled, the entire command fails.
+            halted, the entire command fails.
         \item Execute the Program Buffer, if \Fpreexec is set.
         \item Copy data from the register specified by \Fregno into {\tt data}, if \Fwrite is clear.
         \item Copy data from {\tt data} into the register specified by \Fregno, if \Fwrite is set.
         \item Execute the Program Buffer, if \Fpostexec is set.
-        \item Resume the hart, if \Fpostresume is set. If the hart is already
-            running, the entire command fails.
+        \item Resume the hart, if \Fpostresume is set. If the hart is already 
+        running and \Fprehalt is not set, the entire command fails.
         \end{steps}
 
         If any of these operations fail, \Fcmderr is set and none of the
--- a/abstract_commands.xml
+++ b/abstract_commands.xml
@@ -8,7 +8,7 @@
         \item Copy data from {\tt data} into the register specified by \Fregno, if \Fwrite is set.
         \item Execute the Program Buffer, if \Fpostexec is set.
         \item Resume the hart, if \Fpostresume is set. If the hart is already
-        running and \Fprehalt is not set, the entire command fails.
+            running and \Fprehalt is not set, the entire command fails.
         \end{steps}
 
         If any of these operations fail, \Fcmderr is set and none of the


abstract_commands.xml
```diff
@@ -2,13 +2,13 @@
...
         \item Halt the hart, if \Fprehalt is set. If the hart is already
-            handled, the entire command fails.
+            halted, the entire command fails.
         \item Execute the Program Buffer, if \Fpreexec is set.
         \item Copy data from the register specified by \Fregno into {\tt data}, if \Fwrite is clear.
         \item Copy data from {\tt data} into the register specified by \Fregno, if \Fwrite is set.
         \item Execute the Program Buffer, if \Fpostexec is set.
-        \item Resume the hart, if \Fpostresume is set. If the hart is already
-            running, the entire command fails.
+        \item Resume the hart, if \Fpostresume is set. If the hart is already
```
timsifive: Line ending in space. :-(

abstract_commands.xml
```diff
@@ -2,13 +2,13 @@
...
-            handled, the entire command fails.
+            halted, the entire command fails.
         \item Execute the Program Buffer, if \Fpreexec is set.
         \item Copy data from the register specified by \Fregno into {\tt data}, if \Fwrite is clear.
         \item Copy data from {\tt data} into the register specified by \Fregno, if \Fwrite is set.
         \item Execute the Program Buffer, if \Fpostexec is set.
-        \item Resume the hart, if \Fpostresume is set. If the hart is already
-            running, the entire command fails.
+        \item Resume the hart, if \Fpostresume is set. If the hart is already 
+        running and \Fprehalt is not set, the entire command fails.
```
timsifive: Why do you need the failure be conditional on \Fprehalt?

I imagined you could do a write-register-and-resume command, but if you do that while the hart's already running it should fail when it discovers that. I realize the phrasing is ambiguous. What I meant:
```If \Fpostresume is set: fail the command if the hart is already running, otherwise resume the hart.```

Does that make sense?

mwachs5: No, it doesn't make sense. The hart may already be running, but if you do prehalt, then it will be halted. So it is OK if you do prehalt & postresume together. Or are you saying you would check only at the end of the command's execution? That's weird when placed next to the bit's description, and it seems you'd rather decide at the beginning whether this command is legal (s.t. the only failures you get while running it are exceptions).

mwachs5: I guess the confusion comes from when do I check "if the hart is already running". In my mind, I'm checking the legality of all these bits right when you write this command or do an autoexec. perhaps you mean something like checking whether the hart is running when you get the the point where you make the decision. At any rate, it's ambiguous.

timsifive: Yes, I had imagined checking at the time the resume would happen. Eg. what do you do if the debugger writes a dret to the program buffer, or something else weird happens? Checking at the time of execution feels the most robust to me, and not complicated.

Checking beforehand is nicer in theory, but there may be more corner cases that you miss.

mwachs5:
If the debugger writes 'dret' then this is all irrelevant because it's going not going to be executing the program buffer anymore. Without the 'ebreak' debug module has no idea what's happening.

Didn't you say dret wasn't a real instruction, except as implementation detail? I thought that you could only resume by clearing HALT bit.

 I can imagine many schemes where putting 'dret' in the middle of a program buffer is just going to result in totally undefined behavior.

I'd say that an implementation is perfectly justified in doing the check before-hand, right?

And then it can also check afterwards if it is able to do so, but it may not be able to determine wonky paths out of debug mode like the one you describe.

timsifive:
dret was just an example of something wonky happening during program execution. Other examples might some interrupt that isn't masked properly, instructions with unexpected side effects, etc.
At least in our implementation dret would set the hart running during program buffer execution, messing everything up. In such a case checking just before attempting to resume makes sense.

Is it difficult to check later instead of earlier? It feels like the logic is no more complex, and it might catch some weird corner cases like this.

mwachs5:
In my implementation, "halted" is something determined by the debug module, not the hart. I suppose there is a scenario when the hart signals "Resuming" when not directed by the debugger, but that in and of itself would be the error case, and would be an error regardless of whether or not the 'postresume' bit was set.

Anyway, I have no problem with what you say about checking again. Sure, that can be an error. 

I just ask for permission to error out immediately before even starting if the requested command itself makes no sense... how can I resume if I know I'm not halted, and I'm not going to halt based on this command?

timsifive:
Erroring out immediately sounds fine to me. I'll update the spec to state that if an abstract command fails, any part of it may not be executed.

mwachs5:
Question on another scenario where this could be relevant, not directly related to this issue I guess.

Let's say someone writes a long-running loop into the Program Buffer and then does a command to execute the program buffer. 

What happens if someone writes the HALT bit to 0 at that time? 

Some possibilities (and keeping it open may make it easier to implement):
- Ignore the write and don't signal an error. HALT bit just stays set.
- Resume as soon as you can, no error, HALT bit eventually goes low.
- Some sort of error -- this I don't like, because it doesn't match the other error scenarios... the hart is still off doing who-knows-what.

timsifive:
I would think that resume-as-soon-as-you-can is what happens. The DM doesn't look at the halt bit while it's executing an abstract command.

abstract_commands.xml
asb: My reading of the list was that it was a sequential explanation of behaviour ("Perform the following sequence of actions"). 

The way I understand it, the failure case for line 6 is only encountered if abstract commands can be used without halting the core, and postresume was set but a command was used without halting the core. If the hart was already running and prehalt was set, that failure mode is already described in line 1.

I think currently the spec is giving a simple sequential model of things, and it's left as an exercise for the reader to map to hardware (as you suggest, there are advantages to determining if the command is valid or not immediately, and an implementer might choose to do this). I don't have a strong view on which way of describing behaviour should be preferred.

mwachs5:
Closing this PR as this has changed entirely and the typos are covered in #21

PR #14, opened 2017-02-09
Suggest a binary search for finding the number of harts

asb:
As debuggers are presumably written to work across a large variety of implementations, we should be explicit that a binary search is preferred when finding the number of harts the debug module can access.

That said, is there a reason not to just expose the number of addressable harts in a register somewhere?

### DIFF
--- a/riscv-debug-spec.tex
+++ b/riscv-debug-spec.tex
@@ -403,7 +403,8 @@ generally reset at power-up and if and only if
 Up to 1024 harts can be connected to the DM. The debugger can select any hart
 by writing its index to \Fhartsel. Hart indexes start at 0 and are continuous
 until the final index. This allows the debugger to enumerate all the harts
-attached to the DM by trying each index until \Fhartstatus is 3.
+attached to the DM by trying each index until \Fhartstatus is 3, although a
+binary search should be preferred.
 
 \subsection{Halt Control} \label{haltcontrol}
 


timsifive:
Typically a debugger that wants to figure out what's there, will want to know for each hart at least XLEN and supported extensions, so it can present to the user (who probably doesn't know what indexes correspond to which cores, and doesn't want to go grovel through data sheets to figure it out) some details about each hart. Ie. it would show:
```
0: RV32XYZ
1: RV64G
...
```

In that case the debugger ends up going through each hart anyway. I agree that binary search is an option, but I wouldn't say it's preferred necessarily.

We could just add the number to a register somewhere, but probing doesn't seem that onerous and I would like to keep DMI space somewhat compact.

asb:
I agree that probing harts can provide a more pleasant user experience.
However, as it stands getting that information requires at the very least
halting every hart and reading the misa csr, it could even require
executing test snippets like openocd does to detect xlen. Isn't that rather
invasive to do by default?

There's also the more minor point that this information only represents a
snapshot view, implementations may allow xlen and other fields in misa to
be changed. Doesn't mean it's not useful of course, just something users
and debuggers need to be aware of.

On 9 Feb 2017 21:03, "Tim Newsome" <notifications@github.com> wrote:

Typically a debugger that wants to figure out what's there, will want to
know for each hart at least XLEN and supported extensions, so it can
present to the user (who probably doesn't know what indexes correspond to
which cores, and doesn't want to go grovel through data sheets to figure it
out) some details about each hart. Ie. it would show:

0: RV32XYZ
1: RV64G
...

In that case the debugger ends up going through each hart anyway. I agree
that binary search is an option, but I wouldn't say it's preferred
necessarily.

We could just add the number to a register somewhere, but probing doesn't
seem that onerous and I would like to keep DMI space somewhat compact.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/sifive/riscv-debug-spec/pull/14#issuecomment-278772777>,
or mute the thread

timsifive:
On Thu, Feb 9, 2017 at 2:09 PM, Alex Bradbury <notifications@github.com>
wrote:

> I agree that probing harts can provide a more pleasant user experience.
> However, as it stands getting that information requires at the very least
> halting every hart and reading the misa csr, it could even require
> executing test snippets like openocd does to detect xlen. Isn't that rather
> invasive to do by default?
>

Yes, and the spec doesn't require it. But I predict most debuggers will
work this way.
This is, BTW, one real use case for quick access. It allows gathering of
this data without fully halting the target.

There's also the more minor point that this information only represents a
> snapshot view, implementations may allow xlen and other fields in misa to
> be changed. Doesn't mean it's not useful of course, just something users
> and debuggers need to be aware of.
>

That feels like a misfeature of the spec. Software should be able to
determine the actual limits of the hardware. But it's separate from this
issue, and probably separate from the debug spec altogether. We should
argue to have the privilege spec include complete introspection features.

PR #16, opened 2017-02-11
Allow size 4 for the program buffer

mwachs5:
Why is Program Buffer only allowed to be size 8?

In our previous implementation, for an XLEN=32 design, we only needed 7 words to implement the desired functionality for both program buffer and data transfer. With the latest "quick access" proposal we'll need 3 words for data. Is it possible to do what we need with only 4 words? (or 5 if we must implement ebreak at the end)?


timsifive:
With the latest quick access, we need 8 words to perform an arbitrary write.
On the call there was a suggestion to only allow 0, 1, or 8. That limitation makes writing the debugger easy, and nobody chimed in for other options, so that's what I wrote down. If you think 4 is important, I'm happy to add it.

PR #18, opened 2017-02-12
Flesh out and edit the introduction/background 

mwachs5:
I felt the intro/background was limited in the use cases covered by this document. I fleshed it out a bit, and did some general grammar edits. 

I think this PR could handle feedback from others as well.


riscv-debug-spec.tex
```diff
@@ -242,39 +259,80 @@ \subsubsection{Serial Ports}
...
+\item Debugging low-level software in the absence of an OS or other software.
+
+\item Debugging issues in the OS itself.
+
+\item Bootstrapping a system to test, configure, and program components before
+  there is any executable code path in the system.
+
+\item Accessing hardware on the system without a working CPU.
+
+\item Accesssing parts of a system which a CPU may not have software visibility
```
timsifive: Are we specifying anything that accommodates cache visibility?

riscv-debug-spec.tex
```diff
@@ -242,39 +259,80 @@ \subsubsection{Serial Ports}
...
+  (e.g. cache tag arrays).
+
+\item Efficiently communicating data between a running process and external hardware,
+  for example to download code and data to write a program to memory.
```
timsifive: I wouldn't consider downloading code as an example of communicating between a running process and external hardware. You're blowing the running process away when you do that.

riscv-debug-spec.tex
```diff
@@ -242,39 +259,80 @@ \subsubsection{Serial Ports}
...
+\item Analyzing low-level performance issues using profiling and sampling techniques.
+
+\item Providing a general transport for firmware running on a component to communicate with the
+  outside world.
+
+\end{itemize}
+
+In addition, even without a hardware debugging interface,
+architectural support in a RISC-V CPU can aid software debugging and
+performance analysis, by allowing hardware triggers and breakpoints.
```
timsifive: Remove comma, I think.

riscv-debug-spec.tex
```diff
@@ -242,39 +259,80 @@ \subsubsection{Serial Ports}
...
+debug agent runs on a component (eg.  triggered by a timer interrupt or breakpoint
+on a RISC-V core) which transfers data to or from the debugger
+without halting the component, only briefly interrupting its program flow.
+This functionality is essential if the component is controlling some real-time system (like a
+hard drive) where long timing delays could lead to physical damage.
+This requires additional software support (both on the system as well as on the debugger),
+and efficient communication channels between the component and the debugger.
+
+%MAW -- where does Quick Access fit in the above description? Is it fair to call
+% it run-mode debugging?
```
timsifive: I see Quick Access as a variant on halt mode debug, not run mode.

riscv-debug-spec.tex
```diff
@@ -242,39 +259,80 @@ \subsubsection{Serial Ports}
...
    \item An optional extension allows arbitrary instructions to be executed on
        a halted hart. That means no new debug functionality is needed when a
-       core has custom instructions or state, as long as there exist programs
+       core has additional or custom instructions or state, as
+       long as there exist programs
        that can move that state into GPRs.
    \item An implementation can choose to provide register access without
-       halting.
+     halting.
+   \item An implemenation can choose to provide the ability to automatically
```
timsifive: implementation

riscv-debug-spec.tex
```diff
@@ -287,16 +345,23 @@ \section{Supported Features}
...
    \item Optional serial ports can be used for communication between debugger
        and monitor, or as a general protocol between debugger and application.
-   \item The debugger doesn't need to know anything about the microarchitecture
+   \item The debugger does not need to know anything about the microarchitecture
        of the cores it is debugging.
 \end{enumerate}
 
+This document does not suggest a strategy or implementation
+for hardware test, debugging or error detection techniqes.
+Scan, BIST, etc are out of scope of this specification,
```
timsifive: etc.

riscv-debug-spec.tex
```diff
@@ -323,8 +388,7 @@ \section{System Overview} \label{overview}
 
 Each RISC-V core may implement a Trigger Module for each hart.  These can
 implement breakpoints, which cause a hart to halt spontaneously.  When that
-happens the DM notices because the hart will signal the DM that it's ready for
-an instruction.
+happens the the hart signals the DM that it is halted.
```
timsifive: the the

riscv-debug-spec.tex
```diff
@@ -348,13 +412,13 @@ \section{Debug Module (DM)} \label{dm}
     operations:}
 \item Provide access to a reset signal that allows debugging out of reset.
     (Required)
-\item Allow any individual hart to be halted. (Required)
+\item Allow any individual hart to be halted and resumed. (Required)
 \item Provide status on which harts are halted. (Required)
-\item Provide access to a halted hart's GPRs. (Required)
-\item Give the debugger necessary information about the implementation.
-    (Required)
+\item Provide read (and write?) access to a halted hart's GPRs. (Required)
```
timsifive: Yes, and write.

riscv-debug-spec.tex
```diff
@@ -348,13 +412,13 @@ \section{Debug Module (DM)} \label{dm}
...
+\item Give the debugger necessary information about the implementation. (Required)
 \item Provide access to other hart registers. (Optional)
 \item Force the hart to execute arbitrary instructions. (Optional)
+\item Single step? Is that required? What about seeing / modifying PC?
```
timsifive: Single step is a feature in the hart. It's accessed by setting a bit in a CSR before resuming. The DM doesn't know about it.

riscv-debug-spec.tex
mwachs5: No, but I was wondering if that was something that could be done with system bus master. I mean, you could memory map anything into that space... I guess we can discuss as part of System Bus discussion whether it is intended for this sort of thing or not.

riscv-debug-spec.tex
mwachs5: Good point. What would be a good way to describe it? I'm thinking of the use case of programming the Flash or similar, where you've downloaded some code and then download some data.

riscv-debug-spec.tex
mwachs5: Ok. I was wondering about the PC. PC is a CSR, right? But we aren't specifically requiring CSR access in this list, so wondering whether it is required.

mwachs5:
Yes will do another pass in the morning

riscv-debug-spec.tex
timsifive: But if it's on the system bus, presumably the software can also access it.

We do have a mechanism for debug-only access to stuff, which is the reserved register numbers in the abstract register command. How about:
\item Accessing custom registers that could be added to aid in hardware debug, system bringup, etc.

riscv-debug-spec.tex
timsifive: The way we program flash isn't a debug feature. It's a feature of the debugger, which can take advantage of its ability to download an arbitrary program and execute it. Any debug interface that allows for downloading and executing code will support that, which is any debug interface worth talking about.

riscv-debug-spec.tex
timsifive: To access the PC, you need to either implement the Program Buffer, or support accessing that specific using the abstract command.

riscv-debug-spec.tex
mwachs5: Hmm, still kind of unsure. I think Quick Access is a technique that can be applied to various use cases, like PC sampling, noninvasively setting triggers, and polling memory without system bus master. Not exactly debugging at all.

riscv-debug-spec.tex
mwachs5: Ok, I guess i was wondering if having *some* mechanism to read and/or write the PC is optional. Sounds like it is, though that seems like it would make debugging pretty hard.

timsifive: Some mechanism to read/write the PC is required.

riscv-debug-spec.tex
timsifive: I think of it as an optimization on regular halt-mode debugging, but you are right that it enables the things you mentioned. Without some kind of hardware quick access support, a debugger would have a major impact on the system it's polling.

PR #19, opened 2017-02-12
jtag_dtm: Clarifications, DBUS->DMI

mwachs5:
I did a pass over the JTAG DTM Section. I clarified some things that I think weren't clear, and renamed DBUS to DMI to be consistent with the rest of the doc.

### DIFF
--- a/jtag_registers.xml
+++ b/jtag_registers.xml
@@ -43,47 +43,60 @@
         The size of this register will remain constant in future versions so
         that a debugger can always determine the version of the DTM.
         <field name="0" bits="31:17" access="R" reset="0" />
-        <field name="dbusreset" bits="16" access="W1" reset="0">
-            Writing 1 to this bit resets the dbus controller, clearing any
+        <field name="dmireset" bits="16" access="W1" reset="0">
+            Writing 1 to this bit resets the DMI controller, clearing any
             sticky error state.
         </field>
         <field name="0" bits="15" access="R" reset="0" />
         <field name="idle" bits="14:12" access="R" reset="Preset">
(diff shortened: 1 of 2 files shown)


jtag_registers.xml
```diff
@@ -43,56 +43,69 @@
...
-            Writing 1 to this bit resets the dbus controller, clearing any
+        <field name="dmireset" bits="16" access="W1" reset="0">
+            Writing 1 to this bit resets the DMI controller, clearing any
             sticky error state.
         </field>
         <field name="0" bits="15" access="R" reset="0" />
         <field name="idle" bits="14:12" access="R" reset="Preset">
-            The number of cycles a debugger needs to send the target through
-            Run-Test/Idle after every dbus scan.
+            This is a hint to the debugger of the minimum number of cycles
```
timsifive: This is not just a hint. It's a minimum. Some implementations may only do something in the Run-Test/Idle state. This makes for a slower but simpler implementation of the TAP. And if there's some state machine there, eg. 1 state for read, and another for write, then you would always have to spend 2 cycles in Run-Test/Idle.

It can also be used, as you described, just to avoid seeing a busy state and that is, as you say, not super important.

riscv-debug-spec.tex
```diff
@@ -792,6 +796,11 @@ \subsection{JTAG Connector}
 
 \subsection{JTAG Registers}
 
+%MAW -- is it required that the DTM be the only functionality of the TAP?
```
timsifive: I assumed that the same TAP can also be used for basic scan chain functionality, which is why I left all that IR space for "standard" JTAG instructions. The 5-bit IR is a minimum.

PR #22, opened 2017-02-14
SystemBus: Clean up some formatting and error specification notes.

mwachs5:
One of the error cases (SB read underflow) seemed incorrectly described. Also added some formatting fixes.

### DIFF
--- a/dm1_registers.xml
+++ b/dm1_registers.xml
@@ -195,8 +195,11 @@
 
             3: There was some other error (eg. alignment).
 
-            4: The system bus master was busy when a one of the {\tt sbaddress} or
-            {\tt sbdata} registers was written.
+            4: The system bus master was busy when a one of the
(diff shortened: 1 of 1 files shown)


dm1_registers.xml
```diff
@@ -264,20 +267,27 @@
         If \Fsberror isn't 0 then accesses return error, and don't do anything
         else.
 
-        If the bus master is busy then accesses set \Fsberror, return error,
+        Writes to this register:
+
+        1. If the bus master is busy then accesses set \Fsberror, return error,
```
timsifive: Probably better to use an enumerate environment here.

mwachs5: Wasn't sure if that would work inside the XML, like the commentary

timsifive: Good point. It ends up inside a table cell, which does seem to have a lot of limits on it. Quick Googling doesn't tell me.

Issue #23, opened 2017-02-15
"Serial_select" description bug


zhutmost:
First thank you for working on it.

In ver 0.12, Serial_port_select is part of <dmcontrol>, but in 0.13, it is moved to <serial>.

The description in 7.11.17 is not correct now.

![image](https://cloud.githubusercontent.com/assets/20497382/22967182/6cf39704-f3a0-11e6-8088-5c2364ca450b.png)

timsifive:
I don't understand what exactly is wrong with the current description of serdata. Can you explain in a bit more detail?

mwachs5:
There is no serial field in dmcontrol anymore. It's in serstatus.

mwachs5:
And serstatus should probably be "serialcs" to match the fact that it holds control and status and all other such registers are called "cs"

mwachs5:
I'll make a  PR for this

zhutmost:
Thanks. 
Markdown cannot display angle brackets. Actually I typed `part of <dmcontrol>` but the contents in the angle brackets aren't displayed. 
I'm sorry I don't make it clear.

zhutmost:
And which branch is the main one? The master branch looks out of date.

timsifive:
The main branch is 0.13. That should be the one that shows up in github by default.

PR #25, opened 2017-02-15
Use consistent 'Control and Status' naming for CS registers.

mwachs5:
This addresses #23 , and also cleans up the naming of all Control and Status registers to be the same.


timsifive:
Thank you!

PR #27, opened 2017-02-22
Clarify that the halt state of all harts is maintained through reset

asb:
This addresses the question from me on the mailing list and in today's meeting. There were a few similar questions, so it seems worthwhile to make things more explicit.

### DIFF
--- a/riscv-debug-spec.tex
+++ b/riscv-debug-spec.tex
@@ -400,7 +400,8 @@ Transport Modules. The purpose of this feature is to allow debugging
 programs from the first instruction executed, so exactly what is affected
 by this reset is implementation dependent. Debug Module's own state and registers are
 generally reset at power-up and if and only if
-\Fdmactive in \Rdmcontrol is 0.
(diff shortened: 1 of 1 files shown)


riscv-debug-spec.tex
```diff
@@ -400,7 +400,8 @@ \subsection{Reset Control} \label{reset}
 programs from the first instruction executed, so exactly what is affected
 by this reset is implementation dependent. Debug Module's own state and registers are
 generally reset at power-up and if and only if
-\Fdmactive in \Rdmcontrol is 0.
+\Fdmactive in \Rdmcontrol is 0. This means that the halt state of all harts is
+maintained provided that \Fdmactive is 1, although breakpoints may be cleared.
```
mwachs5: Suggest you call this "triggers" or "trigger CSRs". "Breakpoints" isn't really a term used in the spec, and that makes it more clear.

asb: Thanks Megan, I've pushed an updated commit that refers to "trigger CSRs"

PR #47, opened 2017-03-22
All halt mode changed to debug mode to synchronize with the priv spec.

poweihuang17:
I've used the python script to change all halt mode into debug mode. 
Po-wei

### DIFF
--- a/riscv-debug-spec.tex
+++ b/riscv-debug-spec.tex
@@ -216,7 +216,7 @@ Sections \ref{abstractcommands}, \ref{abstractcs}, \ref{command},
 \subsubsection{Program Buffer}
 
 Sections \ref{programbuffer}, \ref{progbufcs}, \ref{progbuf0}, \ref{access
-register}, \ref{haltmode}, \ref{deb:regprogbuf}, \ref{deb:mrprogbuf}.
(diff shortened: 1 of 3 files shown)


riscv-debug-spec.tex
```diff
@@ -246,7 +246,7 @@ \subsubsection{Serial Ports}
 
 \section{Background}
 
-There are two forms of external debugging. The first is \emph{halt mode}
+There are two forms of external debugging. The first is \emph{debug mode}
```
mwachs5: I think this one should probably stay as "halt"

riscv-debug-spec.tex
```diff
@@ -514,15 +514,15 @@ \subsection{Program Buffer} \label{programbuffer}
 the debugger can write whatever program it likes, but the program must end with
 {\tt ebreak} or {\tt ebreak.c}.
 
-While these programs are executed, the hart does not leave Halt Mode (see
-Section~\ref{haltmode}).  If an exception is encountered during execution of
+While these programs are executed, the hart does not leave Debug Mode (see
+Section~\ref{debugmode}).  If an exception is encountered during execution of
 the Program Buffer, no more instructions are executed, the hart remains in Halt
```
mwachs5: missed one (due to line break I guess)

poweihuang17:
I've finished those two parts pointed out by Megan. Thanks for your help.

PR #68, opened 2017-05-11
Add missing SW Register File

megan-riscv-6thworkshop:
I haven't tested that this actually compiles, please test it before merging this PR. @palmer-dabbelt


palmer-dabbelt:
It's actually xml/sw_regitsers.xml, not xml/sw_registers.tex.  I fixed that it the merge.

PR #69, opened 2017-05-12
First draft of the security proposal.

donbmouse:
Hello,
This PR incorporates the security analysis of the debug specification as performed by Don A. Bailey of Lab Mouse Security. I would like to incorporate this analysis, the remediation strategies, and the authentication proposal (including the serial protocol) to the RISC-V debug specification for review and improvement. 

Thank you,
Don A. Bailey

### DIFF
--- a/security.tex
+++ b/security.tex
@@ -1,0 +1,251 @@
+\chapter{Debug Security}
+
+{\bf This part of the spec needs work before it's ready to be implemented,
+which is why it's in the appendix. It's left here to give a rough idea of some
+of the issues to consider.}
+
+\section{Rationale}
+
+Security is imperative within any computing ecosystem, but especially in the context of a debugging model for any given processing architecture. The RISC-V debugging architecture has been designed for flexibility, ease, and modularity. As such, ensuring security can be enforced as much as is reasonably possible is a critical part of the RISC-V debug specification. The following considerations illuminate this position:
+\begin{itemize}
+    \item Common debug authentication models considered inadequate.
+    \item A hart or core may never know it is being debugged.
+    \item With an optional System Bus in place, it may be impossible for a core/hart to know it is being debugged.
+    \item DM global reset can hold in reset every component in the system (including RTC?).
+    \item DTM may be accessible from many transports (USB, I2C, SPI, etc.).
+    \item Multiple DTMs can be used in parallel; up to use to ensure this doesn't happen.
+    \item The DMI may not be point-to-point, but may be implemented within TileLink, AMBA, or another robust interface.
+    \item Implementations where Program Buffer is implemented as a subset of main memory.
+    \item DM can be active regardless of whether the authenticated bit is unset.
+    \item If Triggers can use the Program Buffer, code may be executed from M-Mode through the abuse of triggers.
+    \item Since trace data {\bf may} be stored in RAM on the system bus, generated trace data could be used to subvert system security.
+\end{itemize}
+
+\subsection{Common Authentication Models}
+The security of a computing technology is often heavily pinned on its ability to control access to its incorporated hardware debugging interfaces. The most common access controls deployed are as follows:
+\begin{itemize}
+    \item Fuses that disable access to debug functionality
+    \item A static security token stored in the DM
+    \item A bootloader that disables DM access on startup and re-enables it through an authentication protocol
+\end{itemize}
+
+While fuses are the most concrete technology for disabling access to a debug module, they are bypassible and, more importantly, deny valid, authorized users access to a critical resource. While security analysts commonly recommend disabling access to debugging modules, it isn't a realistic practice. Hardware and software faults occur in production. Access to the debug module is often a critical path for realizing the source of faults in deployed systems. Furthermore, attacks against fuses have been a proven and cost-effective strategy for bypassing fuses. Glitching, for example, is one of the most well known and repeatedly proven strategies of attack. Thus, fuses, while often used, can be considered a technology that limits {\em legitimate} use while reducing the attack surface to only those individuals with the ability to perform attacks such as glitching. 
+
+Static security tokens, such as passwords, are problematic for less obvious reasons. During production, these keys cannot be made unique to each manufactured unit. This is because it is exceptionally challenging to alter the production process to create a static, unique key and associate it with one specific manufactured unit. The overhead for this process is often unmaintainable. As a result, solutions commonly end up sharing the same static key throughout a deployment. The result is that any individual capable of guessing the key, or any leak of the key, opens up the entire environment to adversarial debugging. As such, any static key (whether it be an encryption key used in challenge-response, or simply a binary number) should not be used in production. 
+
+The bootloader model is problematic for the reasons described in the paragraph above, but also incorporates the TOCTOU security tisk. TOCTOU (Time of Check, Time of Use) is a race condition where the attacker attempts to win a race with a computing system for access to a given resource. On processor reset, the processor's core will enable and configure internal modules (such as the debug module) prior to execution of the first instruction. This means that there is a period of time when the debug module is enabled and accessible prior to the bootloader disabling it in software. An attacker's advantage is the ability to automate the process and perform it an infinite number of times until they have successfully won the race. At this point, the attacker can simply extract the security key from system memory and will never have to race the bootloader again. 
+
+\subsection{Attestation}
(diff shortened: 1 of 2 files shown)


donbmouse:
I should note, this was built and tested against 0.13, even with the sw_registers build bug. It compiles and builds the PDF file correctly, incorporating the security chapter into Appendix E. I've tried to conform as closely as I can to the style of the rest of the document. 

Thanks!

security.tex
```diff
@@ -0,0 +1,251 @@
+\chapter{Debug Security}
+
+{\bf This part of the spec needs work before it's ready to be implemented,
+which is why it's in the appendix. It's left here to give a rough idea of some
+of the issues to consider.}
+
+\section{Rationale}
+
+Security is imperative within any computing ecosystem, but especially in the context of a debugging model for any given processing architecture. The RISC-V debugging architecture has been designed for flexibility, ease, and modularity. As such, ensuring security can be enforced as much as is reasonably possible is a critical part of the RISC-V debug specification. The following considerations illuminate this position:
+\begin{itemize}
```
mwachs5: (NB I wrote a lot of these comments before you gave your presentation, I just never hit 'SAVE' apparently)

security.tex
```diff
@@ -0,0 +1,251 @@
...
+		\multicolumn{2}{c}{\scriptsize 8} & \multicolumn{2}{c}{\scriptsize 24} \\
+	\end{tabular}
+\end{center}
+
+\begin{center}
+	\begin{xtabular}{|l|p{0.5\textwidth}|c|l|}
+	\hline
+    Field & Description \\
+	\hline
+	|Type| & The type of message being sent to the AM.\\
```
mwachs5: Does this spec define types?

security.tex
```diff
@@ -0,0 +1,251 @@
...
+    \item Pad the message, if necessary, to a multiple of 4 octets.
+    \item Determine the message type.
+    \item Construct the message type/length header word. 
+    \item Prepend the type/length header to the message.
+    \item Calculate the CRC32 checksum across the header and the message.
+    \item Append the CRC32 checksum to the message.
+\end{enumerate}
+
+\paragraph{Message Transmission}
+For each 32-bit word in the message:
```
mwachs5: It seems like it could be pretty inefficient if you have to check for OK after every word. Is that really necessary, other than just checking 'busy'?

One could imagine wanting to stream in a bunch of data, then just checking busy/error at the end (that's sort of the pattern in the rest of the spec)

security.tex
```diff
@@ -0,0 +1,251 @@
...
+
+The bootloader model is problematic for the reasons described in the paragraph above, but also incorporates the TOCTOU security tisk. TOCTOU (Time of Check, Time of Use) is a race condition where the attacker attempts to win a race with a computing system for access to a given resource. On processor reset, the processor's core will enable and configure internal modules (such as the debug module) prior to execution of the first instruction. This means that there is a period of time when the debug module is enabled and accessible prior to the bootloader disabling it in software. An attacker's advantage is the ability to automate the process and perform it an infinite number of times until they have successfully won the race. At this point, the attacker can simply extract the security key from system memory and will never have to race the bootloader again.
```
poweihuang17: For "The bootloader model is problematic for the reasons described in the paragraph above", what does this refer to?  I agreed with the TOCTOU, but I don't know what this sentence pointed to.

security.tex
```diff
@@ -0,0 +1,251 @@
...
+
+Static security tokens, such as passwords, are problematic for less obvious reasons. During production, these keys cannot be made unique to each manufactured unit. This is because it is exceptionally challenging to alter the production process to create a static, unique key and associate it with one specific manufactured unit. The overhead for this process is often unmaintainable. As a result, solutions commonly end up sharing the same static key throughout a deployment. The result is that any individual capable of guessing the key, or any leak of the key, opens up the entire environment to adversarial debugging. As such, any static key (whether it be an encryption key used in challenge-response, or simply a binary number) should not be used in production.
```
poweihuang17: One information about static security token, I forwarded it here. It seems that HDCP in HDMI use static token and has individual key for every IC. So I think it's possible for every module to has it own key. Moreover, isn't it possible by PUF ? 
Quote from my Taiwanese friend: 
"以HDMI的HDCP來說，必須向 HDCP委員會購買key，IC在量產測試時對每一顆IC燒錄各自獨立的 Key。"
Translation :
For HDMI of HDCP, we have to buy keys from HDCP, and give each IC a unique key in manufacturing and testing.

security.tex
poweihuang17: For this sentence: "At this point, the attacker can simply extract the security key from system memory and will never have to race the bootloader again. " ,
it bases on an assumption that the key is in system memory, but it needn't.

security.tex
```diff
@@ -0,0 +1,251 @@
...
+
+\subsection{Attestation}
+Attestation is a critical piece of the security puzzle for embedded systems, Internet of Things (IoT), and production server environments. In the context of debug security, attestation is challenging to enforce. How can running code {\em know} that it is being instrumented by a debugger? That challenge is augmented without adequate authentication and authorizatoin controls.
```
poweihuang17: Why the goal of the spec is not to allow a hart to know about debug mode? If the hart could know, it could let the hart knows it's possibly being eavesdropped.

mwachs5: Again, if that is desired, it can be done at a platform-specific way. But to enable debugging, need to ensure that the core behaves "the same" when debugger is attached as when it is not attached, as much as possible, and execute code in the same way.

security.tex
```diff
@@ -0,0 +1,251 @@
...
+
+\subsection{Attestation}
+Attestation is a critical piece of the security puzzle for embedded systems, Internet of Things (IoT), and production server environments. In the context of debug security, attestation is challenging to enforce. How can running code {\em know} that it is being instrumented by a debugger? That challenge is augmented without adequate authentication and authorizatoin controls. 
+
+With the System Bus option in use, a debugger can access system memory or any peripheral device enabled on the target system without having to explicitly control any given hart. The result is a bypass of attestation controls that would otherwise enable a hart to identify whether it has been forced into debug mode.
```
poweihuang17: Yes, so a lock should be presented at DTM and DM to block any attack at the beginning.

poweihuang17:
@timsifive It seems that this security proposal is completely dead. So, we just fall back to our original authentication method, right?

securitymouse:
Hi Guys,

This isn't dead if there is interest. I've been unavailable due to a family
situation. Ironically, that family situation has been resolved recently. If
there is interest, I am happy to revisit this issue, but would need to get
back up to speed on the current specification.

Thank you,

Don A. Bailey
Founder / CEO
Lab Mouse Security
@InfoSecMouse <https://twitter.com/InfoSecMouse>
https://www.securitymouse.com/

timsifive:
@securitymouse If you have time, then the logical step is to address the many comments made below, and to rebase this change onto the latest version of the spec. I don't think a lot has changed that would affect security, but of course I may be wrong.

securitymouse:
Ok sure! I am locked up closing out a large project for a client through
the weekend. But I will be able to take a look at this next week. Does that
work?

Thank you!

Don A. Bailey
Founder / CEO
Lab Mouse Security
@InfoSecMouse <https://twitter.com/InfoSecMouse>
https://www.securitymouse.com/

timsifive:
We're trying to push out a spec version for public comment in just under 3
weeks. That's the only time constraint I'm aware of. If you have some
obvious feedback next week I would expect we can get that in.

securitymouse:
Excellent. I'll do my best and update you if my timeline slips.

Don A. Bailey
Founder / CEO
Lab Mouse Security
@InfoSecMouse <https://twitter.com/InfoSecMouse>
https://www.securitymouse.com/

securitymouse:
Hey sorry for the late note. I came down with a bad flu and have been laid
up for a week now. I can't move quickly. My apologies. If this means the
security proposal gets delayed there's nothing I can do at the moment so
I'll have to accept it. If there are a couple more weeks, I would still
have time. I hope to be better in the next couple of days, so I will circle
back then.

Best,

Don A. Bailey
Founder / CEO
Lab Mouse Security
@InfoSecMouse <https://twitter.com/InfoSecMouse>
https://www.securitymouse.com/

timsifive:
Thanks for the update. Take care of yourself. We're shooting to have a
draft available for public review in 2 weeks. Even during the public review
period you can submit feedback, of course. Earlier is always better, but
life doesn't always allow for that.

PR #78, opened 2017-06-01
Add included tex files as dependencies.

timsifive:


Makefile
```diff
@@ -14,11 +14,23 @@ REGISTERS_TEX += sw_registers.tex
 REGISTERS_CHISEL += dm_registers.scala
 REGISTERS_CHISEL += abstract_commands.scala
 
+INCLUDES_TEX = introduction.tex
```
mwachs5: can't we just do *.tex?

timsifive: I didn't because at times the current directory contains some autogenerated tex files as well. That shouldn't obviously lead to problems, but it felt more right to be explicit.

If you still prefer *.tex I'll change it.

mwachs5: no strong preference, mostly just wondering

timsifive: So I can merge this?

mwachs5: sure

PR #79, opened 2017-06-01
Language cleanups, consistency and typo fixes.

timsifive:


debug_module.tex
```diff
@@ -52,18 +52,20 @@ \section{Debug Module Interface (DMI)} \label{dmi}
 \section{Reset Control} \label{reset}
 
 The Debug Module controls a global reset signal, which can
-reset, or hold in reset, every component in the platform,
+reset, or hold in reset, every component in the platform
 except for the Debug Module and Debug
 Transport Modules. The purpose of this feature is to allow debugging
 programs from the first instruction executed, so exactly what is affected
 by this reset is implementation dependent. The Debug Module's own state and registers are
-generally reset at power-up and if and only if
```
mwachs5: no, this 'and' is necessary. You have changed the meaning here. If it's unclear feel free to propose a different change

debug_module.tex
```diff
@@ -52,18 +52,20 @@ \section{Debug Module Interface (DMI)} \label{dmi}
...
+generally reset at power-up if and only if
 \Fdmactive in \Rdmcontrol is 0. This means that the halt state of all harts is
 maintained provided that \Fdmactive is 1, although trigger CSRs may be cleared.
 
 Due to clock and power domain crossing issues,
 it may not be possible to perform arbitrary DMI accesses across reset,
-even if the Debug Module state is maintained. It must always be possible to access
+even if the Debug Module state is maintained. The debugger must not perform
+a reset while it's expecting a DMI access to complete. It must always be
```
mwachs5: This is not correct. I guess it could be correc tot say "The debugger not perform a DMI access while the core is in reset", but then the second sentence contradicts this.

debug_module.tex
```diff
@@ -241,11 +243,11 @@ \section{Program Buffer} \label{programbuffer}
 While these programs are executed, the hart does not leave Debug Mode (see
 Section~\ref{debugmode}).  If an exception is encountered during execution of
 the Program Buffer, no more instructions are executed, the hart remains in Debug
-Mode, and \Fcmderr is set to 3({\tt exception error}).
+Mode, and \Fcmderr is set to 3 ({\tt exception error}).
 
 Executing the Program Buffer may clobber \Rdpc. If that is the case, it must be
 possible to read/write \Rdpc using an abstract command with \Fpostexec not set.
-The debugger must attempt to save \Rdpc between halting and
+The debugger must always save \Rdpc between halting and
```
mwachs5: There were a lot of discussion about this. How is debugger supposed to save dpc before executing program buffer if the only way to access DPC is through program buffer? So, that is why it should be "attempt to". 

And, why do we need to do this at all?

implementations.tex
```diff
@@ -13,17 +13,17 @@ \section{Abstract Command Based}
...
 
 This method uses the processor's existing pipeline
 and ability to execute from arbitrary memory locations to avoid
-modifications to a processors datapath.
+modifications to a processor's datapath.
 When \Fhaltreq is set, the Debug Module raises a special interrupt
 to the selected hart. This interrupt causes the
 hart to enter Debug Mode and jump to a 
-memory region that is serviced by the DM and execute a 'park loop'.
+memory region that is serviced by the DM and execute a "park loop".
```
mwachs5: is this kosher in LaTex? Don't they have to be ``park loop ''

introduction.tex
```diff
@@ -1,12 +1,12 @@
 \chapter{Introduction}
 \label{sec:intro}
 
-When a design progresses from simulation to a hardware implementation, a users's
+When a design progresses from simulation to hardware implementation, a users's
```
mwachs5: user's (that's my typo but since you're changing it anyway :) )

xml/dm_registers.xml
```diff
@@ -153,8 +148,8 @@
             Debug Module after power up, including the platform's system reset
             or Debug Transport reset signals.
 
-            A debugger should pulse this bit low to ensure that the Debug
-            Module is fully reset and ready to use.
+            When first connecting, a debugger should pulse this bit low to
```
mwachs5: I don't know if "should" is the right word here, or if it should only be done when first connecting.

debug_module.tex
timsifive: How about:
generally reset at power-up and when \Fdmactive in \Rdmcontrol is 0.
Should it be "while" to indicate a level-sensitive signal?

"if and only if" confuses me since it also happens at reset.

debug_module.tex
timsifive: Should we be more explicit about what is possible while reset is asserted? Eg.
While \Fndmreset is asserted, the only supported DM operation is accessing \Rdmcontrol. The behavior of other accesses are undefined.

Reading the original more closely, it's not clear to me what "reset" refers to. Initially I had assumed it refers to resetting of the DM itself.

I think being explicit about what specific operations are allowed when specific resets are asserted is the way to go, but I'm unsure what those are.

debug_module.tex
timsifive: I had imagined a minimal implementation would clobber dpc when executing ebreak from a program buffer/instructions stuffing. Thinking about this more, defining ebreak to not change dpc in Debug Mode is probably simpler and cleaner.

implementations.tex
timsifive: Depends on the quotes you like to see. :-) I'll use the fancy ones.

xml/dm_registers.xml
timsifive: The way I see it, the DM might be in some weird state when a debugger connects, so the debugger is responsible for fixing that by pulsing reset.
If the debugger wants to do it again later, that's fine. It's allowed to, but not required.
I believe this language reflects that.

debug_module.tex
mwachs5: Yes, that wording sounds better than the original. 'while' is also good.

debug_module.tex
mwachs5: So, in rocket-chip, you can't actually get across the CDC boundary while reset is asserted, because the clocks aren't running on the DM side (even though the DM values are maintained, your DMI access can't complete). It doesn't matter what the source of reset is (could be ndmreset, or me holding the reset button down).

DMCONTROL is on the JTCK domain so it's fine to read that one (which is necessary so that you can actually clear ndmreset).

All of this is very implementation specific, but I'm not sure how you would get around it, so wanted to specify the minimum that the implementation has to support.

debug_module.tex
mwachs5: Ah. yeah thats what rocket-chip implemented.

xml/dm_registers.xml
mwachs5: Right, but it doesn't need to be in the *spec* that debugger SHOULD do this (in the IEEE meaning of SHOULD). It's just something the debugger is allowed to do.

debug_module.tex
timsifive: I don't see any reason to access the DM while reset is asserted (except for custom extensions, and people implementing those could build their system to not have this limitation). So how about:
While \Fndmreset or any external reset is asserted, the only supported DM operation is accessing \Rdmcontrol. The behavior of other accesses are undefined.

debug_module.tex
timsifive: I'm reverting this change to "attempt to," and I'll make a separate change that removes the language altogether and changes 'ebreak' to only touch dpc when not in Debug Mode.

xml/dm_registers.xml
timsifive: I would like for it to be in the spec, so that compliant debuggers will perform the reset when connecting, and will work with targets that need that reset. (I can't think of any unhappy side effects pulsing dmactive.)

mwachs5: I don't see a reason while Debugger A couldn't connect and put things in a certain state, then debugger B comes along and hooks up. There is no reason to *require* that B reset the state. Many debuggers may want to do this, but it shouldn't really be a requirement of the spec. 

The spec should state what the hardware and software agree on. Forcing the software to do certain things except what is required by the hardware shouldn't really be in the spec.

debug_module.tex
mwachs5: The obvious case is when debugger writes ndmreset bit. Then it HAS to write ndmreset bit back to 0. Unless you're expecting it to auto-clear?

But yes your suggested text addresses this use case.

xml/dm_registers.xml
timsifive: I agree that you can switch debuggers and preserve *hart* state, but resetting the DM doesn't affect the system that's being debugged at all, right? It just affects which harts are currently selected, default system bus access size etc. These are all things that any debugger would configure anyway.

Hardware *may* require that a debugger perform a reset when first connecting. By requiring the software to do it we make everything a little more consistent.

debug_module.tex
timsifive: I agree with you about how to use ndmreset. My last proposal is a bit more explicit than the original text. Should I go with that?

mwachs5: 👍

debug_module.tex
mwachs5: sounds good

xml/dm_registers.xml
mwachs5: I still think this should simply state that "A debugger may pulse this bit low to get the debug module into a known state". It doesn't have anything to do with first connecting.

debug_module.tex
timsifive: Another reason the debugger might want to save dpc:
An implementation may alias dpc to pc, and pc might change when that implementation executes a Program Buffer. Should we just not allow that?

xml/dm_registers.xml
timsifive: I'll go with that for now.

debug_module.tex
mwachs5: That's what I thought you put it in there for. I don't know of any implementations that have a problem with it (despite discussion at several meetings / on the mailing list). Current implementation of OpenOCD does not do this.

xml/dm_registers.xml
```diff
@@ -65,19 +65,14 @@
...
 
-            01: There is a Debug Module and it conforms to version 0.11 of this
+            1: There is a Debug Module and it conforms to version 0.11 of this
             specification.
 
-            10: There is a Debug Module and it conforms to version 0.13 of this
+            2: There is a Debug Module and it conforms to version 0.13 of this
             specification.
-
-            11: Reserved for future use.
```
mwachs5: Should we leave "others: Reserved for future use"?

timsifive: I'm tempted to put some general language somewhere that says any values not explicitly listed in the spec are reserved for future use.

timsifive:
Megan, I think all these items are waiting on your final OK. Can you take a look?

introduction.tex
```diff
@@ -62,6 +62,9 @@ \subsection{Register Definition Format}
         \hline
         R/W1 & Read/Write. Only writing 1 has an effect.  \\
         \hline
+        R/W1C & Read/Write. Writing every bit high clears the field. Writing
```
mwachs5: Do you want this to be Read/Write-1-Clear ?

For multi-bit fields like cmderr, it is actually a bitwise OR, it shouldn't necessarily check that all bits are cleared. For example, you may read a certain error code and write back that error code to "acknowledge" it, without acknowledging other errors.

timsifive: Hmm... good point. But other fields (eg. cmderr) contain a multi-bit number. If you clear them bit-by-bit there's a race. Eg. the register contains 2. Debugger writes 2 to clear it, but error 3 happens just before. Now the next time the register is read it contains 1.
Actually in the case of cmderr that can't happen since the errors are sticky.

I think it makes sense to define fields that are really bitfields themselves as individual fields. Then this description is correct.

Thoughts?

mwachs5: Yeah, but I think the "don't unintentionally clear errors that you didn't actually observe" is the exactly desired behavior. 

Someone else suggested that you have to write exactly what is in the field in order to clear it.

Otherwise you have to go to 1-bit-per-error type, which is what I think you're suggesting.

Both of these are backwards incompatible.

timsifive: I think we're not talking about exactly the same thing. Let's take cmderr, which is currently a 3-bit field. It can contain 0 (no error), as well as several other values for error codes (1, 2, 3, 4, 7).
An alternate encoding would be to use 5 bits, one for each of the possible errors. I had thought that is what you were talking about, but now it sounds like that's not the case. I am not proposing that.

So let's stick with the 3-bit encoding. What we're discussing is how these errors should be cleared. Let's say the error code is 2. There are 3 options I can think of:
1. Write 7 (every bit high)
2. Write 2 (only the bits that are set in the error)
3. Write 0 (the value that you want to see in the register)

3 was discarded because it's too easy to accidentally clear an error.
I don't like 2 because of the race I described, but I don't know of an actual case where it's a problem since all (?) our error codes are sticky.
1 is my preferred solution.

But I don't feel very strongly about that. I just think it needs to be documented, which was the whole point of this change. :-)

mwachs5: Well there is one more option which is what is actually implemented in Spike/Rocket-Chip:

4. Each bit is treated like an unrelated W1C field. 

Since our error codes are sticky, the race doesn't really exist, and it's simpler if the update logic only needs to consider 1 bit.

PR #82, opened 2017-06-05
Add stepie to dcsr.

timsifive:
This allows single step with or without interrupts disabled.

### DIFF
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -4,8 +4,8 @@
         \begin{commentary}
             \Fintdisable is intended to be used in conjunction with \Fstep.
             Usually it is necessary to disable interrupts so that single step
-            debugging is not constantly interrupted by jumps to the exception
-            handler. Occasionally a user will want to step into the exception
+            debugging is not constantly interrupted by jumps to the interrupt
+            handler. Occasionally a user will want to step into the interrupt
             handler, and for those cases the interrupt behavior is controllable
             here. Having the debugger use mstatus.ie for this functionality is
             not desirable because it has side effects when stepping through
@@ -35,7 +35,7 @@
         <field name="intdisable" bits="11" access="R/W" reset="0">
             0: Interrupts are handled as usual.
 
-            1: Interrupts are disabled. This bit supersedes mstatus.ie.
+            1: Interrupts are disabled. This bit supersedes mstatus.mie.
 
             This feature may not exist in implementations from early in the
             specification process.  The debugger must read back the value it
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -1,6 +1,16 @@
 <registers name="Core Debug Registers" prefix="CSR_">
     These registers are only accessible from Debug Mode.
     <register name="Debug Control and Status" short="dcsr" address="0x7b0">
+        \begin{commentary}
+            \Fintdisable is intended to be used in conjunction with \Fstep.
+            Usually it is necessary to disable interrupts so that single step
+            debugging is not constantly interrupted by jumps to the exception
+            handler. Occasionally a user will want to step into the exception
+            handler, and for those cases the interrupt behavior is controllable
+            here. Having the debugger use mstatus.ie for this functionality is
+            not desirable because it has side effects when stepping through
+            code that accesses mstatus itself.
+        \end{commentary}
         <field name="xdebugver" bits="31:30" access="R" reset="Preset">
           0: There is no external debug support.
 
@@ -22,7 +32,15 @@
             When 1, {\tt ebreak} instructions in User/Application Mode enter
             Debug Mode.
         </field>
-        <field name="0" bits="11" access="R" reset="0" />
+        <field name="intdisable" bits="11" access="R/W" reset="0">
+            0: Interrupts are handled as usual.
+
+            1: Interrupts are disabled. This bit supersedes mstatus.ie.
+
+            This feature may not exist in implementations from early in the
+            specification process.  The debugger must read back the value it
+            writes to check whether the feature is supported.
+        </field>
         <field name="stopcount" bits="10" access="R/W" reset="Preset">
 
             0: Increment counters as usual.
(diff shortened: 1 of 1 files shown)


xml/core_registers.xml
```diff
@@ -1,6 +1,16 @@
 <registers name="Core Debug Registers" prefix="CSR_">
     These registers are only accessible from Debug Mode.
     <register name="Debug Control and Status" short="dcsr" address="0x7b0">
+        \begin{commentary}
+            \Fintdisable is intended to be used in conjunction with \Fstep.
```
mwachs5: Why did you put this commentary here rather than with the actual bit?

xml/core_registers.xml
```diff
@@ -22,7 +32,15 @@
             When 1, {\tt ebreak} instructions in User/Application Mode enter
             Debug Mode.
         </field>
-        <field name="0" bits="11" access="R" reset="0" />
+        <field name="intdisable" bits="11" access="R/W" reset="0">
+            0: Interrupts are handled as usual.
+
+            1: Interrupts are disabled. This bit supersedes mstatus.ie.
```
mwachs5: There is no such bit. I think you mean MSTATUS.mie ? What about xSTATUS.Yie, for x, y = M, U, S

mwachs5: I think you can disregard my second question. Since Debug mode is similar to M-Mode, saying that this supersedes MSTATUS.mie is sufficient

xml/core_registers.xml
```diff
@@ -22,7 +32,15 @@
...
+
+            This feature may not exist in implementations from early in the
```
mwachs5: I would just say this is a WARL bit and may be hardcoded depending on the implementation. There is no guarantee that it will ever be implemented.

mwachs5: And it should probably be called "IE" since the polarity is an enable vs a disable then

xml/core_registers.xml
```diff
@@ -1,6 +1,16 @@
 <registers name="Core Debug Registers" prefix="CSR_">
     These registers are only accessible from Debug Mode.
     <register name="Debug Control and Status" short="dcsr" address="0x7b0">
+        \begin{commentary}
+            \Fintdisable is intended to be used in conjunction with \Fstep.
+            Usually it is necessary to disable interrupts so that single step
+            debugging is not constantly interrupted by jumps to the exception
+            handler. Occasionally a user will want to step into the exception
+            handler, and for those cases the interrupt behavior is controllable
```
zhonghochen: Interrupt handler is more precise than exception handler.

xml/core_registers.xml
mwachs5: For consistency we could call it DCSR.die (though somewhat misleading name!)

xml/core_registers.xml
timsifive: Trying to keep the contents of the cell small. Not sure what happens if you try to put the commentary in one, but it doesn't feel right.

xml/core_registers.xml
timsifive: The way I wrote the bit, it doesn't solely apply to single step. It seemed like it'd be simpler to implement that way.

I made it a disable so that existing implementations which have a 0 here will just be compliant.

We can revisit either of those decisions.

xml/core_registers.xml
timsifive: Good point. I'm fixing that.

xml/core_registers.xml
mwachs5: The spec previously said that interrupts were always disabled during single step/debug mode. And this bit was a 0 before. So wouldn't the 0 = interrupts disabled be the backwards compatible way? What am I missing?

xml/core_registers.xml
mwachs5: sounds good

xml/core_registers.xml
mwachs5: Also, I can't think of a case where you would actually want interrupts during debug mode. Everything about how MEPC etc would get updated is unclear. You don't want to be getting interrupts in the middle of executing program buffer, for example. 

During single step is a different story, because you leave debug mode during single step.

Do you have an application in mind when you'd actually want interrupts enabled *during debug mode*?

aswaterman:
It's more complicated than "supersedes MIE" - it's really, if this bit is 0, don't take any interrupts; and if this bit is 1, and you would have taken the interrupt if not single-stepping, then take it.  This is a function of privilege mode x, and mstatus.xie, not just MIE.

xml/core_registers.xml
timsifive: No, and this bit doesn't enable interrupts during debug mode. During debug mode interrupts are still always disabled. This bit can be used to disable interrupts while running regular code (not in debug mode), whether that be while single stepping, or running freely.

Clearly I need to work on the description, but I'm not sure how. In my reading of the text it doesn't mention anything about changing interrupt behavior during Debug Mode. What can I add to make it more clear?

mwachs5: The statement that interrupts are masked during Debug Mode:
https://github.com/riscv/riscv-debug-spec/blob/0512f5d9e9f80b392b91ca32c17f493544820302/core_debug.tex#L17

I think this is a statement we don't want to change.

Then there are 2 separate functionalities we are discussing:

1) The statement here says that interrupts are disabled while single-stepping. I envisioned that this bit would only be an "IE during single stepping":
https://github.com/riscv/riscv-debug-spec/blob/1fbbe6e955e4cbfd478c1a07d33e2106bf65670e/xml/core_registers.xml#L69

2) An entirely different functionality, which creates another bit "DIE" which when 0 would disable interrupts, regardless of whether in debug mode or not. That is a completely separate issue from 1 with a larger scope.

mwachs5: The reason I say that these are different functionalities is that if you want this spec to be backwards compatible, you need two bits, not 1, to get the currently specified behavior.

mwachs5:
I renamed this bit to "stepie" and noted that it may be hard-wired to 0. The debugger is fully capable of emulating the decision & action to take on interrupt when this bit is not implemented.

This doesn't address the other item of "die" bit. I'm not sure of the use case for that bit.

timsifive:
Looks good to me.

Issue #71, opened 2017-05-15
define 'dret' again


mwachs5:
'dret' was removed since it's now an "implementation detail". But if we are to have any sort of reusable components, we should define the dret instruction, even if it is optional to implement.

PR #93, opened 2017-06-13
Define DRET instruction

mwachs5:
Although from the SW debugger perspective it is not necessary to define dret, we should still reserve the opcode. DRET is needed for some instruction feeding approaches, and also defning it here will alow the same debug module hardware to be used to debug different core implementations.


mwachs5:
Addresses #71

PR #102, opened 2017-06-30
Remove Trace

mwachs5:
As discussed in WG, the trace section is not part of the current ratifiable spec under discussion. Remove it to avoid confusion.


mwachs5:
Good idea. I created 'before_trace_removal' tag.

PR #104, opened 2017-07-03
Move Serial Ports proposal to Appendix

mwachs5:


timsifive:
If we removed trace from the appendix because it confused people, does that mean serial should also be removed? The serial spec is more complete than trace, but afaik nobody is implementing it and it's also not a hard requirement for the first round of ratification.

mwachs5:
It's a reasonable suggestion. But by that logic we should probably just remove the entire "future ideas" section. Let's consider this an intermediate step for now and we could talk about more things to remove entirely for ratification.

poweihuang17:
Yes, I agree. Without ratification, it's hard to get open source programmer involved. So, we better focus on the most important part of ratification and remove those unnecessary stuff. I guess because debugging tools from ARM and Lauterbach usually include all these stuff, so we put it in the spec before.  But trace and serial ports are really independent concepts and unnecessary at this stage.

Issue #100, opened 2017-06-28
Does reading one CSR mean that you should be able to read any CSR? (ditto Write, ditto FPR)


mwachs5:
We should clarify, if you have one CSR abstract command which fails, does that mean all CSRs would also fail

mwachs5:
https://github.com/riscv/riscv-openocd/pull/79 Seems to make this assumption since it caches the result of the first time it tries to read/write CSRs/FPRs. Is that the conclusion?

timsifive:
It's just something that I implemented until the spec makes clear what I should do. Using 2 bits per CSR seemed a little excessive.

In any case, what I implemented is safe as long as a program buffer exists to fall back on.

PR #107, opened 2017-07-10
Clarify implications of CSR read, write, halt

mwachs5:
To remove the ambiguity about whether e.g.  reading a given CSR implies anything about other CSRs.


mwachs5:
Closes #100

PR #108, opened 2017-07-10
clarify DCSR.cause

mwachs5:
There was still some references to "halt mode" vs. "debug Mode". The causes in DCSR.cause should specify "why was debug mode entered", not really "why did hart halt".


mwachs5:
NB this doens't change functionality, just removes some ambiguity in meaning

PR #109, opened 2017-07-11
Use LL instead of L for 64-bit constant suffix.

timsifive:
See https://github.com/riscv/riscv-openocd/pull/81


mwachs5:
I wanted the updated text in debug_defines.h so I snuck in the merges ahead of yours.

PR #111, opened 2017-07-21
Clarify dpc contents.

timsifive:
There was some confusion about this. See
https://groups.google.com/a/groups.riscv.org/forum/?utm_medium=email&utm_source=footer#!msg/debug/o9buamtml2s/a4r4qwBTFAAJ

### DIFF
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -90,8 +90,12 @@
     </register>
 
     <register name="Debug PC" short="dpc" address="0x7b1">
-        Upon entry to debug mode, \Rdpc is written with the
-        virtual address of the instruction that encountered the exception.
+        Upon entry to debug mode, \Rdpc is updated with the virtual address of
+        the next instruction to be executed.  {\tt ebreak} instructions are an
+        exception to this. When an {\tt ebreak} instruction is executed, \Rdpc
+        is set to the address of the {\tt ebreak} instruction. (Another way to
+        think of this is that {\tt ebreak} instructions act like illegal
+        instruction exceptions.)
 
         When resuming, the hart's PC is updated to the virtual address stored in
         \Rdpc. A debugger may write \Rdpc to change where the hart resumes.
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -91,11 +91,32 @@
 
     <register name="Debug PC" short="dpc" address="0x7b1">
         Upon entry to debug mode, \Rdpc is updated with the virtual address of
-        the next instruction to be executed.  {\tt ebreak} instructions are an
-        exception to this. When an {\tt ebreak} instruction is executed, \Rdpc
-        is set to the address of the {\tt ebreak} instruction. (Another way to
-        think of this is that {\tt ebreak} instructions act like illegal
-        instruction exceptions.)
+        the next instruction to be executed. The behavior is described in more
+        detail in Table~\ref{tab:dpc}.
+
+        \begin{table}[H]
+        \centering
+        \caption{Virtual address in DPC upon Debug Mode Entry}
+        \label{tab:dpc}
+        \begin{tabular}{|l|L|}
+        \hline
+        Cause &amp; Virtual Address in DPC \\
+        \hline
+        {\tt ebreak} &amp; Virtual address of the {\tt ebreak} instruction \\
+        \hline
+        single step &amp; Virtual address of the instruction after the single
+        instruction that was executed \\
+        \hline
+        trigger module &amp; If \Ftiming is 0, the virtual address of the
+        instruction which caused the trigger to fire. If \Ftiming is 1, the
+        virtual address of the next instruction to be executed at the time that
+        debug mode was entered \\
+        \hline
+        halt request &amp; Virtual address of the next instruction to be
+        executed at the time that debug mode was entered \\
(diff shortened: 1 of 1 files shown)


xml/core_registers.xml
```diff
@@ -90,8 +90,12 @@
     </register>
 
     <register name="Debug PC" short="dpc" address="0x7b1">
-        Upon entry to debug mode, \Rdpc is written with the
-        virtual address of the instruction that encountered the exception.
+        Upon entry to debug mode, \Rdpc is updated with the virtual address of
+        the next instruction to be executed.  {\tt ebreak} instructions are an
```
mwachs5: I think this is less clear. "The instruction that encountered the exception" is what it is. It's only relevant that it is "the next instruction to be executed" when you are EXITING debug mode.

I think the confusion is that entry to debug mode is not always caused by an exception. "Halt Request" is not really defined as an exception. It acts like an interrupt. We should just use the same language as the Priv spec uses for explaining how to update MEPC when there is an interrupt.

timsifive:
For reference, this is what the definition of mepc says:

> When a trap is taken into M-mode, mepc is written with the virtual address of the instruction that encountered the exception. Otherwise, mepc is never written by the implementation, though it maybe explicitly written by software.

xml/core_registers.xml
mwachs5: This is what Priv ISA says about $mepc:

"When a trap is taken into M-mode, mepc is written with the virtual address of the instruction that encountered the exception."

But then in the NMI section it says this: "The mepc register is written with the address of the next instruction to be executed at the time the NMI was taken,"

So I think we should break out this description into the "entry into debug mode caused by halt request (acts like NMI case)" and "entry into debug mode caused by breakpoint exception" (acts like exception), and use same phrasing.

mwachs5:
ah our comments crossed

timsifive:
Maybe we just need a table:

| Cause | DPC |
| --- | --- |
| ebreak | Address of ebreak instruction |
| single step | Address of the instruction after the one that was stepped over |
| trigger module | Address of the next instruction to be executed |
| halt request | Address of the next instruction to be executed |

mwachs5:
If that's the correct table then I guess we even disagree... trigger module should have the address of the instruction which caused the exception, shouldn't it? If I set a breakpoint to hit on a certain instruction, I expect the dpc to have that address. So I think there is only two cases, as I tried to clarify with #112

mwachs5:
Also, to be clear, DPC always has "address of next instruction to be executed". That isn't very helpful, the question is what should that instruction be.

timsifive:
Trigger module is tricky, because it has its own separate timing section, which indicates whether the trigger should happen before/after the instruction is committed. In my interpretation before means that the instruction never executed so that it should still be executed.

I agree that "address of the next instruction" isn't as helpful as it should be.

mwachs5:
Hmm, good point about the read data match trigger. In that case it acts more like an interrupt which is raised by the instruction which encountered the match, so it doesn't take effect until the next instruction.

mwachs5:
Here was my suggested edit:
"  Upon entry to debug mode due to a halt request, 
 +	\Rdpc is updated with the virtual address of
 +        the next instruction to be executed. Upon entry to debug
 +	mode due to a breakpoint exception, \Rdpc
 +        is set to the address of the instruction which caused the exception.
"

timsifive:
So there you address halt request and ebreak, but not triggers or single step. (They could perhaps be one of those you mentioned, but it's not obvious what is what.) With that in mind, I think a table where we explicitly address each cause makes most sense. Updated below:

Cause | DPC
-- | --
ebreak | Virtual address of the ebreak instruction
single step | Virtual address of the instruction after the one that was stepped over
trigger module | If \Ftiming is 0, the virtual address of the instruction which caused the trigger to fire. If \Ftiming is 1, the virtual address of the next instruction to be executed at the time that debug mode was entered
halt request | Virtual address of the next instruction to be executed at the time that debug mode was entered

mwachs5:
Ok, seems reasonable. But I think SingleStep may have a few cases that we at least need to at least consider. It may already be obvious, but just making sure it's clear and the spec reflects the intent.

For example, let's say our $mtvec is set to 0xDEAD_BEEF (some bad non-legal address).

In debug mode, we set $dpc to 0x0 (some non-executable address).
Then we set $dcsr.step and resume.

When we enter debug mode due to single step, we will immediately get an exception and immediately enter debug mode, because you're supposed to do that when you get an exception in single step. $mepc should be updated with 0x0. Presumably $dpc holds 0xDEAD_BEEF?

I feel like this needs to be clear since "after the one that was stepped over" doesn't really apply if you never even completed the instruction you stepped over.

Nit, I don't really like the phrase "stepped over", as it implies you didn't do the actual stepped instruction. But if this is standard phrasing wording, so be it.

timsifive:
That's a nice corner case for single step. It feels a little bit beyond this section, though. Do you agree we should add a Single Step section to RISC-V Debug to explain this case?

mwachs5:
why ddi you delete all the "virtual" statements? I think those were added earlier to clarify that dpc holds virtual address.

timsifive:
I deleted the "virtual" statements in the description of each case, because the header in the table already states the address is virtual.

PR #112, opened 2017-07-21
Change DPC wording

mwachs5:
Make it more clear the difference between "halt request" (interrupt-like) and "breakpoint exception" (exception-like) entry to debug mode, and how it impacts $dpc update.


timsifive:
Can we keep this conversation in PR #111? That way some future curious soul will have a better chance of finding one discussion that explains our thought process.

mwachs5:
Sure.

PR #125, opened 2017-08-02
DMI does not have an Error response

### LINKED ISSUE #116: What are DMI Errors "returned only by System Bus Error"
Why don't system bus accesses also just set an error flag?

They already do?

This is a historical artifact of when lots of things retruned errors, and this is the only one left. Can we get rid of it?

mwachs5:
Closes #116 

There really is no reason to have DMI return error responses, as in all other error cases a sticky error status bit is set. The only one left here was for System Bus access. This PR removes this last remaining case.


mwachs5:
Oops I apparently did this on top of the DPC branch. Let me rebase and force push.

Issue #124, opened 2017-08-02
Fix Travis Build (sympy package not found)


mwachs5:
Can you take a look at why Travis build is failing? For some reason the sympy package is no longer found, maybe Travis switched to some new version of python that doesn't include it by default?

PR #126, opened 2017-08-02
Fix and speed up build.

timsifive:
Making a pull request so Travis will build it. Not done yet.


timsifive:
This fixes Issue #124.

Issue #87, opened 2017-06-05
Clarify actual current requirements on JTAG Header


mwachs5:
Right now there are statements about the current supply the JTAG header must supply. NOt sure we are ready to make such a strong requirement

timsifive:
Can I close this?

PR #128, opened 2017-08-02
Don't require a target to provide 25mA on VCC.

timsifive:
Issue #87.


timsifive:
@mwachs5 Can you review this?

Issue #113, opened 2017-07-26
Configuration String-- is it in the ISA spec anymore?


mwachs5:
Debug spec talks about configuration string, but not sure that  is part of the ISA Spec at all anymore. We should update debug spec to align.

PR #131, opened 2017-08-09
Configuration String -> Device Tree

timsifive:
Issue #113.


mwachs5:
Does the term "Device Tree" have meaning? Or should this be "DTB"?

mwachs5:
Talked to some ISA spec folks, and this wording is fine.

PR #145, opened 2017-09-13
Incorporate some of Nikhil's feedback

timsifive:


debug_module.tex
```diff
@@ -102,24 +108,21 @@ \section{Selecting Harts} \label{selectingharts}
...
 the \Rdmcontrol register.
-When a debugger wants to halt a single hart it selects it in \Fhartsel
-and sets \Fhaltreq, then waits
-for \Fallhalted to indicate the hart is halted before clearing \Fhaltreq
-to 0. Setting \Fhaltreq has no effect on a hart which is already halted.
-
-To resume, the debugger selects the hart with \Fhartsel,
-and sets \Fresumereq. This action sets the hart's
-{\tt resumeack} bit to 0. Once the hart has resumed, it
+After selecting harts, it sets \Fhaltreq, then waits for \Fallhalted to
```
mwachs5: I think the tense here (this isnt' really a comment on your change as it was already like this) is strange for a spec. It says "a debugger does this", rather than what a debugger and/or hardware must do. MUST the debugger wait for \Fallhalted? MUST it set \Fhaltreq to 0?

For example, for reset, we don't always immediately set \Fhaltreq to 0 right away.

debug_module.tex
```diff
@@ -102,24 +108,21 @@ \section{Selecting Harts} \label{selectingharts}
...
+To resume, the debugger sets \Fresumereq. This action sets each selected hart's
+{\tt resumeack} bit to 0. Once a hart has resumed, it
 sets its {\tt resumeack} bit to 1. Thus, the
-debugger should wait for \Fallresumeack to indicate the
-hart has resumed before clearing \Fresumereq to 0.
-If a debugger reads \Fallresumeack
-and \Fallhalted in the same cycle, the hart must have resumed
-and then halted again.
+debugger should wait for \Fallresumeack to indicate all
+hart have resumed before clearing \Fresumereq to 0.
```
mwachs5: harts

debug_module.tex
timsifive: The debugger only asserts haltreq when it feels like it. It doesn't have to wait for allhalted. It might just assert briefly and be happy with whichever harts are now halted.

The only MUST in there might be that haltreq needs to be deasserted before interacting with the hart in Debug Mode. Is that a requirement?

Does the whole sentence be prefixed with "Typically, ..."?

debug_module.tex
timsifive: Fixed.

PR #156, opened 2017-09-28
Combine {all,any}{halted,running} into runstate.

timsifive:
This is an incompatible change, and needs the version number to be incremented.

Based on feedback by Rishiyur Nikhil.


timsifive:
I don't feel strongly about it.

Issue #148, opened 2017-09-20
clarify ndmreset vs. hartreset


timsifive:
Does ndmreset reset harts as well?

mwachs5:
it resets "everything" but the debug module. Otherwise what is the point of it?

mwachs5:
I believe hartreset is specifically to reset only the selected harts.

mwachs5:
Closed by #157

PR #157, opened 2017-09-29
Clarify ndmreset.

timsifive:
Inspired by isse #148

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -143,11 +143,11 @@
         <field name="0" bits="15:2" access="R" reset="0" />
         <field name="ndmreset" bits="1" access="R/W" reset="0">
             This bit controls the reset signal from the DM to the rest of the
-            system. To perform a system reset the debugger writes 1,
+            system. The signal should reset every part of the system, including
+            every hart, except for the DM.
+            To perform a system reset the debugger writes 1,
             and then writes 0
-            to deassert the reset. This bit must not reset the Debug Module
-            registers. What it does reset is platform-specific (it may
-            reset nothing).
+            to deassert the reset.
         </field>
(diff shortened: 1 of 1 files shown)


xml/dm_registers.xml
```diff
@@ -143,11 +143,11 @@
         <field name="0" bits="15:2" access="R" reset="0" />
         <field name="ndmreset" bits="1" access="R/W" reset="0">
             This bit controls the reset signal from the DM to the rest of the
-            system. To perform a system reset the debugger writes 1,
+            system. The signal should reset every part of the system, including
+            every hart, except for the DM.
```
mwachs5: what about the debug transport? should we just say "the debug logic, including the Debug Module and Debug Transport Module"

timsifive: I'm ambivalent about that. The reset could or could not reset the transport. We should be clear about it, though. Sounds like you would prefer for the transport not to be reset. But what if the transport is something used by the reset of the system as well (like some shared PCI bus)? I'm inclined to add something like:
"This may include the debug transport. The debugger must be prepared to reestablish the connection to the Debug Module immediately after writing 1 to ndmreset."

mwachs5: Ok. Well I guess it gets back to the "how do you then clear ndmreset" if the transport is reset.

timsifive: Right. I forgot about that. I'll add something explicit to the spec about that.

timsifive: Or is this an argument to make the register handle the entire reset sequence and clear itself?

timsifive:
Is this good to go?

mwachs5:
I think so. we may want to also answer the "how long does it have to be asserted 1 for" question (to which the answer should be no requirement, its up to the system to stretch or synchronize the reset as needed)

Issue #151, opened 2017-09-20
is there one haltreq signal, or is there one bit per hart?


timsifive:
The spec is not entirely clear whether haltreq directly connects to only the currently selected harts, or if it modifies a bit which persists even when the currently selected harts changes. The former seems easier to implement in hardware. The latter would help resetting multicore systems that don't support hawindowsel.

Based on Nikhil's observation:
> Section 4.3, para 1
> "If the halt signal is asserted when a core comes out of reset, the
> core must enter Debug Mode before executing any instructions, ..."
>
> Clarify:
>
> "If 'dmcontrol.haltreq' is asserted when a core comes out of reset,
>  then all harts selected by dmcontrol.hasel/hartsel must enter Debug
>  Mode before executing any instruction ..."
>
> and
>
> "If 'dmcontrol.haltreq' is asserted when a hart comes out of reset,
>  then if it is selected by dmcontrol.hasel/hartsel, it must enter
>  Debug Mode before executing any instruction ..."

mwachs5:
The latter is *required* for resetting multicore systems that don't support hawindowsel. Let's clarify the intention in the text.

timsifive:
What happens on a 2-hart system, when you:
1. Select hart 0, and write haltreq
2. Select hart 1, and clear haltreq
3. Select both harts, and read haltreq

I'm tempted to make haltreq, resumereq and hartreset write-only.

mwachs5:
Yes, it was already sort of ambiguous whether they were readable.

PR #158, opened 2017-10-05
haltreq, resumereq, hartreset are per-hart bits

timsifive:
We need that to halt all harts out of reset on a system that does not
support hawindow.

Issue #151.


debug_module.tex
```diff
@@ -110,8 +110,10 @@ \section{Selecting Harts} \label{selectingharts}
 
 \section{Run Control} \label{runcontrol}
 
-The Debug Module can halt selected harts and allow them to run again using
-the \Rdmcontrol register.
+For every hart, the Debug Module contains 3 bits: halt request, resume request,
+and hart reset. (The hart reset bit is optional.) These bits all reset to 0. A
```
mwachs5: they only reset to 0 when debug logic is reset (maybe this is obvious, maybe not)

timsifive: Obvious to me. Same reset as is mentioned in every register description. I don't think it's necessary to list what signal does the reset, since our descriptions of the reset signals already list what they reset.

debug_module.tex
```diff
@@ -110,39 +110,21 @@ \section{Selecting Harts} \label{selectingharts}
...
-resume. Debuggers should not depend on this behavior and should not
-set \Fresumereq for running harts.
-
-To halt or resume multiple harts at the same time, the debugger first
-sets the hart's bits in the hart array mask register, then
-follows the same procedure but with \Fhasel set to 1.
-Depending on the desired operation, the debugger might consider the {\tt any*}
-versions of the status instead of {\tt all*}.
+For every hart, the Debug Module contains 3 bits: halt request, resume request,
+and hart reset. (The hart reset bit is optional.) These bits all reset to 0. A
```
mwachs5: period after parens

timsifive: Huh? You mean like this:
and hart reset. (The hart reset bit is optional). These bits all reset to 0. A

I thought it goes inside because it's an entire sentence. (https://english.stackexchange.com/a/11129)

mwachs5: ah, you are correct. I rescind my comment

debug_module.tex
```diff
@@ -110,39 +110,21 @@ \section{Selecting Harts} \label{selectingharts}
...
+debugger can write them for the currently selected harts through \Fhaltreq,
+\Fresumereq, and \Fhartreset in \Rdmcontrol.
+
+When a running hart sees a halt request, it responds by halting and asserting
+its halted signal. The halted signals of all selected harts are reflected in
```
mwachs5: how about the halt summary registers?

It's a little strange to say above that "each hart has 3 bits" and now you are actually talking about a 4th. Did you consciously leave it off the list above? Or, maybe you should say above "there are 3 conceptual control bits", since below you go on to discuss even more bits/signals.

timsifive: I'm differentiating between bits (eg. FFs) and signals. Halt request is a bit set by the debugger, that persists even if a different hart is selected. Halted is a signal from a hart to the DM. It might be driven by a FF, or by some logic, etc. Is that confusing? How can I make it clear?

mwachs5: I think it's not clear to me what you are trying to clarify :) The fact that the control signals are in flip flops or not is also just as much an implementation detail as the status signals.

I think you should just say "3 control signals" instead of "3 bits" and it would be clear to me.

timsifive: The control signals *have* to be in flip-flops, otherwise their value is lost when a different hart is selected.

The halt summary is probably not in flip-flops. It's probably just logic on the halted signals, which is sampled when the read happens.

mwachs5: but all of this detail doesn't need to be in the spec. The fact that they are flops or not is really not necessary to clarify. Isn't a "halted signal" also probably in a flop? 

What is the higher level confusion we are trying to clear up?

timsifive: The part that has to be in the spec is that if a debugger sets haltreq on one hart, and then selects a different hart, that the first hart still sees a hartreq. There has to be state in the DM to track that.

mwachs5: Not really. Maybe "haltreq" creates a message that gets sent by the Debug Module over some crazy network-on-chip, which responds with some packet later back to the Debug Module that tells the Debug Module it actually halted. The debug Module doesn't *have* to store the fact that it requested halt in a flop (unless we want to be able to read the haltreq bit when we switch back, which maybe we do, but that doesn't seem to be the thrust of this PR).

mwachs5: But I think what is confusing in the spec and my example is that if a halt is requested, the hart halts,  then that hart gets a reset, then the hart should STILL be halted. So in my crazy NoC example, cores would need to know that they must stay halted as long as they don't get a "ok to not halt message" which is presumably which what an implementation would send when haltreq is written to 0.

timsifive: I agree that your example would be a valid implementation.

I've attempted to clarify that the bits are just state, and that there are additional signals from each hart to the DM.

debug_module.tex
```diff
@@ -110,39 +110,22 @@ \section{Selecting Harts} \label{selectingharts}
...
-versions of the status instead of {\tt all*}.
+For every hart, the Debug Module contains 3 bits of state: halt request, resume
+request, and hart reset. (The hart reset bit is optional.) These bits all reset
+to 0. A debugger can write them for the currently selected harts through
+\Fhaltreq, \Fresumereq, and \Fhartreset in \Rdmcontrol. In addition the DM
+receives halted, running, and resume ack signals from each hart.
+
+When a running hart sees a halt request, it responds by halting and asserting
+its halted signal. The halted signals of all selected harts are reflected in
+the \Fallhalted and \Fanyhalted bits. \Fresumereq is ignored by running harts.
```
mwachs5: Shouldn't the \Fresumereq and \Fhaltreq "is ignored by" be swapped? This one seems like it would go with the halt and the other seems like it shoudl go with the resume.

xml/dm_registers.xml
```diff
@@ -103,27 +105,30 @@
...
-        <field name="resumereq" bits="30" access="R/W" reset="0">
-            Resume request signal for all currently selected harts. When set to 1,
-            each selected hart will resume if it is currently halted.
+        <field name="resumereq" bits="30" access="W" reset="-">
+            Writes the resume request bit for all currently selected harts.
+            When set to 1, each selected hart will resume if it is currently
+            halted.
 
-            This bit is ignored while \Fhaltreq is set.
+            The resume request bit is ignored while the halt request bit is
```
mwachs5: This is a backwards compatible change by the way (not from this actual PR but I think this was added recently), and could be hard to implement. It could conflict with possible implementations of quick access. Wasn't this undefined before?

debug_module.tex
```diff
@@ -110,39 +110,22 @@ \section{Selecting Harts} \label{selectingharts}
...
-may cause a hart which halts in the future to immediately
-resume. Debuggers should not depend on this behavior and should not
-set \Fresumereq for running harts.
-
-To halt or resume multiple harts at the same time, the debugger first
-sets the hart's bits in the hart array mask register, then
-follows the same procedure but with \Fhasel set to 1.
-Depending on the desired operation, the debugger might consider the {\tt any*}
-versions of the status instead of {\tt all*}.
+For every hart, the Debug Module contains 3 bits of state: halt request, resume
```
mwachs5: can we change this to say "the Debug Module controls 3 conceptual bits of state"? I think if we said that I would be fine with it.

debug_module.tex
```diff
@@ -110,39 +110,22 @@ \section{Selecting Harts} \label{selectingharts}
...
+request, and hart reset. (The hart reset bit is optional.) These bits all reset
+to 0. A debugger can write them for the currently selected harts through
+\Fhaltreq, \Fresumereq, and \Fhartreset in \Rdmcontrol. In addition the DM
+receives halted, running, and resume ack signals from each hart.
+
+When a running hart sees a halt request, it responds by halting and asserting
```
mwachs5: can we replace "sees" with "receives"

xml/dm_registers.xml
mwachs5: sorry I meant "incompatible"

debug_module.tex

debug_module.tex

debug_module.tex
timsifive: Yes. Thanks.

xml/dm_registers.xml
timsifive: I don't understand how this affects anything. It's saying that the debugger must clear haltreq before (or at the same time as) settings resumereq. Does it sound like something different is meant?

mwachs5: It says "the resume request is ignored while...". That is a stronger HW requirement than "Setting both resumereq and haltreq at the same time leads to undefined behavior".

timsifive: Good point. I think I like this better since it forces sane debugger behavior.

Issue #160, opened 2017-10-11
Implicit ebreak indicator


mwachs5:

mwachs5:
Sometimes (currently when progbufsize = 1) you don't have to write an ebreak. But space in the program buffer just for an ebreak is expensive. Yet, some implementations may not be able to control code execution out of the program buffer. So suggestion is that we add a bit which when 1 implies "ebreak is implicit" and debugger doesn't need to write it at the end of the program buffer (it still could with no ill effects). If this bit is 0, software must write the ebreak.

PR #162, opened 2017-10-11
Add \Fimpebreak, to support of implicit ebreak.

timsifive:
Issue #160.


debug_module.tex
```diff
@@ -242,18 +242,24 @@ \section{Program Buffer} \label{programbuffer}
...
 execute it exactly once with the Access Register Abstract Command,
-
 setting the \Fpostexec bit in \Rcommand.
-If \Fprogbufsize is 1, the Program Buffer may only hold a single instruction.
-This can be a 32-bit
+The debugger can write whatever program it likes (including jumps out of the
+Program Buffer), but the program must end with
+{\tt ebreak} or {\tt ebreak.c}. To save hardware, an implementation may support
+an implied {\tt ebreak} that is executed when a hart runs off the end of the
+Program Buffer. With this feature, a Program Buffer of just 2 32-bit words can
```
mwachs5: shouldn't we say "as indicated by \Fimpebreak" or something similar?

xml/dm_registers.xml
```diff
@@ -17,7 +17,14 @@
         time.  Eg. in a multi-hart system some might temporarily be powered
         down, or a system might support hot-swapping harts.
 
-        <field name="0" bits="31:21" access="R" reset="0" />
+        <field name="0" bits="31:23" access="R" reset="0" />
```
mwachs5: "reset" should not be 0, it should be "whatever it is".

xml/dm_registers.xml
```diff
@@ -17,7 +17,14 @@
...
+        <field name="impebreak" bits="22" access="R" reset="Preset">
+            If 1, then there is an implicit {\tt ebreak} instruction that will
+            be executed if execution runs off the end of the Program Buffer.
```
mwachs5: can we make this a bit more clear... If program buffer word [\Fprogbufsize -1] is executed. To make it clear what is "the end" of program buffer.

xml/dm_registers.xml
timsifive: It's 0 for the 0 part of the register. For the impebreak it's "Preset" (see one line down).

debug_module.tex
timsifive: Good idea.

xml/dm_registers.xml
timsifive: I started down that route, but gave up when I couldn't think of a good phrasing for:
After the last instruction in the Program Buffer is executed, and does not change the PC in any way besides normal incrementing...

Suggestions?

xml/dm_registers.xml
mwachs5: ah yep.

xml/dm_registers.xml
mwachs5: Hmm, how about just saying there is an implicit {\tt ebreak} at the (nonexistent) ProgBuf location ProgBuf[\Fprogbufsize]. That's really what this is saying.

timsifive:
How's this?

PR #163, opened 2017-10-11
Remove hmode reference, to fix build.

timsifive:


mwachs5:
I think we should change it to \dmode is 1. Sorry about the accidental push!

timsifive:
:+1:

Issue #155, opened 2017-09-28
Run/Halt Debug State Machine needs a legend


timsifive:
This state machine is color coded, but there's no description of what the colors mean. I assume they correspond to various optional functionality.

PR #164, opened 2017-10-11
Add legend to the  Abstract Command State Machine Diagram

mwachs5:
Also updated some states and transitions to reflect the "transfer" bit


mwachs5:
addresses #155

Issue #165, opened 2017-10-16
Request for "has been reset" indicator


mwachs5:
(From Yung-Ching Hsiao) The current debug spec is lacking the capability to report the occurrences of resets (for the selected hart). Examples where this might be useful: While the external debugger is waiting for some breakpoint to happen: 

* watchdog fires and triggers reset. 
* Accidental touch of the reset button (pilot errors). 

The external debugger software could deal with these situations more gracefully if such information is available to it. I wonder if something like dmstatus.ANYRESET makes sense to everybody, or could we allow dmstatus.ANYUNAVAIL to be a sticky bit (and require W0 to dmstatus to clear it?).

mwachs5:
I think this is a good feature request. I think we would want to distinguish between "is currently unavailable" and "has been unavailable since last cleared". I wonder if we want to overload "unavailable" or do we really mean "has been reset"?

timsifive:
I lean towards anyreset/allreset bits in dmstatus, a clearreset bit in dmcontrol to clear the state. Whenever a hart resets for any reason it sets it reset signal, and the DM notices this and sets a sticky bit. Later the debugger comes along and clears it by writing clearreset.

How does that sound?

mwachs5:
I think the general approach is good, but some different namings may be less confusing. I'm copying your text below with some different names:

I lean towards *anyhasreset/allhasreset* bits in dmstatus, a *hasresetack* bit in dmcontrol to clear the state. Whenever a hart resets for any reason it sets it reset signal, and the DM notices this and sets a sticky bit. Later the debugger comes along and clears *the reset ack for any selected harts* by writing *hasresetack*.

PR #168, opened 2017-10-24
Add 'has reset' status and control

mwachs5:
Addresses #165


debug_module.tex
```diff
@@ -76,6 +75,12 @@ \section{Reset Control} \label{reset}
 operation is accessing \Rdmcontrol. The behavior of other accesses is
 undefined.
 
+When harts are reset, they should set a sticky {\tt hasreset} state bit.
```
timsifive: Should or must?

debug_module.tex
```diff
@@ -76,6 +75,12 @@ \section{Reset Control} \label{reset}
...
+The {\tt hasreset} state bits can be read for selected harts in the \Fanyhasreset
+and \Fallhasreset in \Rdmstatus. This bit should be set whether the
```
timsifive: "regardless of what caused the reset to happen"?

timsifive: Also, "These bits"

debug_module.tex
```diff
@@ -76,6 +75,12 @@ \section{Reset Control} \label{reset}
...
+hart reset under the direction of the debugger or not. It can be cleared
```
timsifive: "They can be cleared"

debug_module.tex
```diff
@@ -76,6 +75,12 @@ \section{Reset Control} \label{reset}
...
+by writes to \Fhasresetack in \Rdmcontrol.
```
timsifive: by writing to

xml/dm_registers.xml
```diff
@@ -17,17 +17,16 @@
         time.  Eg. in a multi-hart system some might temporarily be powered
         down, or a system might support hot-swapping harts.
 
-        <field name="0" bits="31:23" access="R" reset="0" />
-        <field name="impebreak" bits="22" access="R" reset="Preset">
+        <field name="0" bits="31:24" access="R" reset="0" />
+        <field name="impebreak" bits="23" access="R" reset="Preset">
```
timsifive: Why are you moving impebreak?

xml/dm_registers.xml
```diff
@@ -17,17 +17,16 @@
...
             If 1, then there is an implicit {\tt ebreak} instruction at the
             non-existent word immediately after the Program Buffer. This saves
             the debugger from having to write the {\tt ebreak} itself, and
             allows the Program Buffer to be one word smaller.
 
             This must be 1 when \Fprogbufsize is 1.
         </field>
-        <field name="0" bits="21" access="R" reset="0" />
-        <field name="dmerr" bits="20:18" access="R/W1C" reset="0">
+        <field name="dmerr" bits="22:20" access="R/W1C" reset="0">
```
timsifive: I wouldn't move this, unless you know it hasn't been implemented anywhere yet.
(FWIW, OpenOCD doesn't check these bits.)

xml/dm_registers.xml
```diff
@@ -37,6 +36,12 @@
 
             7 (other): An access failed for another reason.
         </field>
+	<field name="allhasreset" bits="19" access="R" reset="-">
```
timsifive: allhavereset?

xml/dm_registers.xml
```diff
@@ -37,6 +36,12 @@
...
+          This field is 1 when all currently selected harts have been reset since the last write to its \Fhasresetack.
+        </field>
+        <field name="anyhasreset" bits="18" access="R" reset="-">
```
timsifive: anyhavereset?

xml/dm_registers.xml
```diff
@@ -37,6 +36,12 @@
 
             7 (other): An access failed for another reason.
         </field>
+	<field name="allhasreset" bits="19" access="R" reset="-">
+          This field is 1 when all currently selected harts have been reset since the last write to its \Fhasresetack.
```
timsifive: ... since the last time reset was acknowledged for them.

Referring to the last \Fhasresetack write seems confusing regarding which harts were/are selected.

xml/dm_registers.xml
```diff
@@ -37,6 +36,12 @@
...
+        </field>
+        <field name="anyhasreset" bits="18" access="R" reset="-">
+          This field is 1 when any currently selected hart has been reset since the last write to its \Fhasresetack.
```
timsifive: ... since the last time reset was acknowledged for it.

xml/dm_registers.xml
mwachs5: because dmerr moved it up if you really want 3 bits for it. I guess I missed the addition of dmerr, not sure what the point of it is, and also it seems weird to have it cross a byte boundary.

xml/dm_registers.xml
mwachs5: yeah, I agree it was awkward as I was writing it. I think this parallels the same things you are trying to address with #158

xml/dm_registers.xml
mwachs5: I know of nowhere that it has been implemented, wasn't even aware of its addition... We can discuss in meeting tomorrow

xml/dm_registers.xml
mwachs5: I guess the problem is that neither "have" nor "has" works with both "any" or "all" grammatically.

"Do you know if any hart has reset in the past 5 minutes?"
"Do you know if all harts have reset in the past 5 minutes?"

`has` is fewer letters...

timsifive: Any harts have reset in the past 5 minutes.
All harts have reset in the past 5 minutes.

But that doesn't work in question form. The other bits solve this by removing the verb altogether, although that creates the question of whether reset is a verb. If we're not going to be able to be 100% clear, maybe just go with "allreset" and "anyreset" to parallel the halted bits. I don't feel strongly about this since I can't find something that really works across the board.

xml/dm_registers.xml
timsifive: I don't care about dmerr. It'd be nice if impebreak doesn't move (or if it does that we're certain about its final location with this change), because the small program buffer OpenOCD pull request uses it.

debug_module.tex
```diff
@@ -75,11 +75,11 @@ \section{Reset Control} \label{reset}
...
 
-When harts are reset, they should set a sticky {\tt hasreset} state bit.
+When harts are reset, they must set a sticky {\tt hasreset} state bit.
 The {\tt hasreset} state bits can be read for selected harts in the \Fanyhasreset
-and \Fallhasreset in \Rdmstatus. This bit should be set whether the
-hart reset under the direction of the debugger or not. It can be cleared
-by writes to \Fhasresetack in \Rdmcontrol.
+and \Fallhasreset in \Rdmstatus. These bits should be set regardless of the cause of
+the reset. 
+They can be cleared writing 1 to \Fhasresetack in \Rdmcontrol.
```
timsifive: "cleared by writing"

xml/dm_registers.xml
```diff
@@ -37,10 +37,10 @@
             7 (other): An access failed for another reason.
         </field>
 	<field name="allhasreset" bits="19" access="R" reset="-">
-          This field is 1 when all currently selected harts have been reset since the last write to its \Fhasresetack.
+          This field is 1 when all currently selected harts have been reset but not acknowledged.
```
timsifive: I'd go with
"This field is 1 when all currently selected harts have been reset but the reset has not been acknowledged."

Otherwise it's not clear to me what's acknowledged.

debug_module.tex
```diff
@@ -76,6 +75,13 @@ \section{Reset Control} \label{reset}
 operation is accessing \Rdmcontrol. The behavior of other accesses is
 undefined.
 
+When harts have been reset, they must set a sticky {\tt havereset} state bit.
+The conceptual {\tt havereset} state bits can be read for selected harts in
+\Fanyhavereset and \Fallhavereset in \Rdmstatus.
+These bits should be set regardless of the cause of the reset.
```
timsifive: must be set

xml/dm_registers.xml
```diff
@@ -37,6 +28,22 @@
...
+	<field name="0" bits="23" access="R" reset="0" />
+        <field name="impebreak" bits="22" access="R" reset="Preset">
+            If 1, then there is an implicit {\tt ebreak} instruction at the
+            non-existent word immediately after the Program Buffer. This saves
+            the debugger from having to write the {\tt ebreak} itself, and
+            allows the Program Buffer to be one word smaller.
+
+            This must be 1 when \Fprogbufsize is 1.
+        </field>
+	<field name="0" bits="21:20" access="R" reset="0" />
```
timsifive: The diff looks like indenting is not consistent between eg. this line and the previous one.

xml/dm_registers.xml
```diff
@@ -140,8 +147,14 @@
 
             Writes apply to the new value of \Fhartsel and \Fhasel.
         </field>
-        <field name="0" bits="28:27" reset="0" access="R" />
-         <field name="hasel" bits="26" access="R/W" reset = "0">
+	<field name="haveresetack" bits="28" access="R/W" reset="0">
```
timsifive: "ackreset"? "haveresetack" sounds like a bit that indicates whether "reset ack" is supported or not.

mwachs5: how about 'ackhavereset'? I do like tying the bits together since there are a lot of resets one could be acking

xml/dm_registers.xml
mwachs5: some are tabs and some are spaces. The whole file is inconsistent.

xml/dm_registers.xml
timsifive: Perfect!

xml/dm_registers.xml
timsifive: :-( Maybe something to fix in a separate PR then.

mwachs5: woops I already went through and pushed it to this one.

timsifive: No worries. At least it's fixed now. :-)

PR #171, opened 2017-11-08
Explain cache coherency wrt to system bus access

timsifive:
I suspect we'll have to do something for systems that don't implement
the Program Buffer, but let's get this ball rolling at least.


debug_module.tex
```diff
@@ -299,6 +299,13 @@ \section{System Bus Access} \label{systembusaccess}
 involving a hart, regardless of whether Program Buffer is implemented.
 The System Bus Access block uses physical addresses.
 
+If a hart has a cache, the data in the cache may not be consistent with the
+memory as accessed directly through the system bus. It is up to the debugger to
+have each hart write back its caches before performing a system bus read, and
+to invalidate the caches after a system bus write. The simplest way to do this
+is by executing {\tt fence} instructions using the Program Buffer. If there is
```
mwachs5: "fence" instructions do not necessarily write back caches, or anything else of that nature. The RISC-V ISA doesn't even mention caches, so I would hesitate to mention them here. I agree we need some sort of statement like that, but I don't think we should claim here that a "fence" does anything specific with regards to caches.

timsifive: Good point. I've removed the reference to fence instructions.

debug_module.tex
```diff
@@ -299,6 +299,14 @@ \section{System Bus Access} \label{systembusaccess}
 involving a hart, regardless of whether Program Buffer is implemented.
 The System Bus Access block uses physical addresses.
 
+If a hart has a cache, the data in the cache may not be consistent with the
```
mwachs5: I suggest leaving out all mention of "caches" and keeping it more abstract.

Depending on the microarchitecture, the data read and/or written by the debugger may not always be consistent with that observed by the hart(s). It is up to the debugger to enforce consistency if the implementation does not. This specification does not define a standard way to do this, as it is implementation/platform specific. Possibilities may include using the SystemBus and/or Program Buffer to write to special memory-mapped locations, or executing special instructions via the Program Buffer.

Issue #169, opened 2017-10-30
Are "conceptual" bits cleared on dmactive = 0


mwachs5:
we clarified the spec to explain that there are hart request, resume request, etc bits.  Also there are hasreset bits.

Do we need to clarify the impact of dmactive on these bits?

It seems like hart request, resume request, would be cleared to 0 when dmactive is 0, but that wouldn't really change the fact of whether the hart is halted or running or not. So maybe these are less confusing.

For hasreset, should those bits persist regardless of dmactive? This would also clarify that the first time a debugger hooks up, and sets dmactive to 1 for the first time, hasreset could be 0 even though the hardware was reset.

timsifive:
It doesn't matter to me what happens to those bits when dmactive goes low, so I'm inclined to spec whatever is simplest to implement. If a debugger cares about these bits, it should not let dmactive go low.

mwachs5:
Also the spec already says that the other bits aren't really defined when dmactive is 0. I'm fine with that.

mwachs5:
Closed by #172

Issue #120, opened 2017-07-26
Should we declare that ndmreset assertion time must be synchronized


mwachs5:
Clarify that that implementer needs to make sure that this reset is synchronized to their system, and asserted as long as possible.

Or is there a reset handshake -- we poll to see if it is out of reset.

timsifive:
Brainstormy thoughts:
1. We could add a register that indicates in real time what the ndmreset/reset pulse widths should be.
2. If there is some kind of system ready signal internal to a system, it would be cool to expose that in dmstatus.

timsifive:
Nikhil's feedback raises this issue as well:
>  - Resets have no handshake: the debugger asserts and then deasserts
>       reset after some fixed number of cycles (even though different
>       implementations may take varying number of cycles to perform
>       resets).

timsifive:
From Nikhil:
> Section 3.10, Section on dmcontrol: hartreset, ndmreset and dmactive
> "the debugger writes 1, and then writes 0 to deassert the reset"
>
> Why? The debugger does not know how long a reset takes, so it's dicey
> to count time or cycles.  Why not make these R/W1 fields, instead of
> R/W, and let the debugger poll for hartresetack, ndmresetack and
> dmactiveack, respectively, either at these same bit positions, or in
> separate bits in the dmstatus register?

mwachs5:
To answer the question of "Why do we have to write 0 to deassert the reset" --- asserting reset may kill all the clocks in the system, except the debug/test interface clock. It's not obvious what logic would deassert that reset. We could say "after 1 debug module interface clock tick it is automatically reset" but that introduces the idea of a Debug module interface clock, which is pretty low-level implementation detail. Saying "you have to write 0 to deassert reset" is clear and straightforward to implement.

But we should still clarify that the hart(s) may not ACTUALLY come out of reset for some time later, again this is also described in #165

mwachs5:
Closed by #172

PR #172, opened 2017-11-16
Reset clarifications 

mwachs5:
* Add explanation about how \Fndmreset and actual system reset are related (addresses #120)
* Also clarify that havereset bits may or may not be ~~are not~~ affected by dmactive low (addresses #169)


timsifive:
Why aren't the havereset bits affected by dmactive going low? I wouldn't want to mandate that they are, but requiring them to persist limits implementation options.
Is it to somehow track resets that necessarily also reset the debug module?

mwachs5:
so you're saying just leave it undefined what happens on dmactive = 0?

timsifive:
Yes. Or explicitly state that the those bits might be reset when the debug module is reset.

mwachs5:
What do you mean "when the debug module is reset". The concept of the debug module reset isn't clearly stated anywhere, unless you're referring to dmactive.

timsifive:
I'm trying to imagine when you might want to have havereset bits persist even when dmactive goes low. The only scenario I can think of is when the debugger resets the target (including debug module) completely using eg. a dedicated JTAG pin.

I still think it's better to let the implementation do whatever they like with havereset values when dmactive is low.

Issue #121, opened 2017-07-26
add Single Step section that details some corner cases


timsifive:
> For example, let's say our $mtvec is set to 0xDEAD_BEEF (some bad non-executable address).
> 
> In debug mode, we set $dpc to 0x0 (some non-executable address).
> Then we set $dcsr.step and resume.
> 
> When we exit debug mode, we will immediately get an exception and immediately enter debug mode, because you're supposed to do that when you get an exception in single step. $mepc should be updated with 0x0. Presumably $dpc holds 0xDEAD_BEEF?
> 
> I feel like this needs to be clear since "after the one that was stepped over" doesn't really apply if you never even completed the instruction you stepped over.
> 

--Megan

mwachs5:
I think we did have some discussion on this already, need to revisit the text to see if this is clear

timsifive:
I can't find a Single Step section in the current version of the document (except in the implementation section). I feel it should be added to the RISC-V Debug section.

mwachs5:
I think this is the bit I was talking about:
https://github.com/riscv/riscv-debug-spec/blob/c1e61b0530085a6ed19b738de1585621decd26d1/xml/core_registers.xml#L73

mwachs5:
I would say that that linked text doesn't make it clear what $dpc would be set to in that case (or in all cases)

timsifive:
I agree, but I don't think that is a good spot to delve into the corner cases of single step, just because it's in a table and formatting wise it will look wrong.

PR #179, opened 2017-11-22
Clarify some single step corner cases.

timsifive:
Issue #121. There's too much emphasis on what not in this change, but
I'm not sure how to rephrase it to be positive.


core_debug.tex
```diff
@@ -42,10 +42,14 @@ \section{Single Step}
 A debugger can cause a halted hart to execute a single instruction and then
 re-enter Debug Mode by setting \Fstep before setting \Fresumereq.
 
-If the instruction that is executed causes an exception, Debug Mode is re-entered
-immediately after the PC is changed to the exception handler and and the
+If executing or fetching that instruction causes an exception, Debug Mode is
+re-entered immediately after the PC is changed to the exception handler and the
 appropriate {\tt badaddr} and {\tt cause} registers are updated also.
```
mwachs5: this looks good, but drop the last 'also'?

core_debug.tex
```diff
@@ -42,10 +42,14 @@ \section{Single Step}
...
 
+If executing or fetching the instruction causes a trigger to fire, Debug Mode
+is re-entered immediately after that trigger has fired. Whether the instruction
```
mwachs5: Do we need to clarify that DCSR.cause is set to "trigger" in that case?

core_debug.tex
timsifive: Yes.

core_debug.tex
timsifive: Done.

core_debug.tex
```diff
@@ -37,6 +37,25 @@ \section{Load-Reserved/Store-Conditional Instructions}
...
 sc} pairs.
 
+\section{Single Step}
+
+A debugger can cause a halted hart to execute a single instruction and then
+re-enter Debug Mode by setting \Fstep before setting \Fresumereq.
+
+If executing or fetching that instruction causes an exception, Debug Mode is
+re-entered immediately after the PC is changed to the exception handler and the
+appropriate {\tt badaddr} and {\tt cause} registers are updated.
```
mwachs5: I just remembered that "*badaddr" is now "*tval" in the latest priv spec, so we should use that name instead

Issue #180, opened 2017-11-23
Access to unimplemented Debug Module Registers 


zhonghochen:
https://github.com/riscv/riscv-debug-spec/blob/ce1b359329ba87c3302df496885ff3781cb4c053/debug_module.tex#L363-L364

Since dmerr is added, would it be better to update the description? 
Writing unimplemented Debug Module DMI Registers updates dmerr to 1.

mwachs5:
I agree that these two sections are somewhat inconsistent. But I don't see a good use of dmerr so far, and one way of "implementing" debug module registers is to return 0 and ignore writes if the functionality doesn't exist. This is pretty clearly spelled out for every register that is optional. So I am not sure what functionality dmerr is adding other than reserving space for future TBD  uses.

zhonghochen:
Could dmerr be optional? Is there any usage for justifying dmerr is worth to be implemented? 
Debugger requires an extra step for clearing dmerr after confirming a feature is not implemented.

mwachs5:
I agree with the above... requiring the debugger to both attempt to read the register, checking dmerr, and clearing dmerr, seems like much more work for both SW HW implementers than just reading the register and seeing if it has it's "unimplemented" value as defined in the spec.

mwachs5:
What if we change the text from:

```

timsifive:
Per Issue #190, we're about to remove dmerr.

PR #181, opened 2017-11-24
dmerr is set on invalid DM access.

timsifive:
Issue #180


timsifive:
Rishiyur wanted dmerr. I'm ambivalent. But if we have it, we should use it.

The intent is to set it both on reads and writes. How can we rephrase this to be more clear?

timsifive:
That's also fine. We can leave the Issue/PR in case it comes up in the
future.

PR #184, opened 2017-11-28
Remove dmerr

mwachs5:
This is a proposed alternative to #181, to just remove dmerr.


timsifive:
I'm fine removing this field.

It was added per discussion subject:"Detailed comments on latest draft of debug spec":
> >   > Sec 3.1, para 2
> >   > "It supports read and write operations."
> >   >
> >   > There used to be a qualification in an earlier draft which has been
> >   > removed in current draft: "It supports read and write operations,
> >   > which may return an error. (Errors are only returned by the optional
> >   > System Bus Access Block)."
> >   >
> >   > What happens if the debugger attempts to read or write an
> >   > unimplemented Debug Module address?  There seems no way to signal this
> >   > kind of error.
> 
> >   Right. This was changed to remove the requirement on the internal bus
> >   that it has to support error reporting. I'm adding a line stating that
> >   unimplemented registers in the DM must read 0, and writing them has no
> >   effect.
> 
> I agree it's good for the internal bus not to have to support error
> reporting.  I was implying that there ought to be a DM register that
> records this, that can be queried with a later read.
> 
> I'm adding a dmerr field. We'll see if people like it.
>

mwachs5:
Reopening following today's discussion. The spec intends that registers can deterministically be determined to exist by either reading them directly and if they return 0 they do not exist, or by inferring whether they exist by reading other registers. In all cases this is defined already in the spec.

It's also unclear how to get rid of dmerr and whether it is blocking.

For that reason we would like to remove it entirely.

timsifive:
Aided by confusion, I ended up pushing my own change to this branch, which is identical to yours except that it also clarifies:
For each register it is possible to check whether it is implemented by checking
that it's not 0 (eg. \Rsbcs), or by checking bits in another register (eg.
\Fprogbufsize).

debug_module.tex
```diff
@@ -363,4 +363,8 @@ \section{Debug Module DMI Registers} \label{dmdebbus}
 When read, unimplemented Debug Module DMI Registers return 0. Writing them has
 no effect.
 
+For each register it is possible to check whether it is implemented by checking
+that it's not 0 (eg. \Rsbcs), or by checking bits in another register (eg.
```
mwachs5: maybe clarify that you just have to read it "it's not 0 -> reading it returns 0"

timsifive:
This is Issue #190

mwachs5:
For some reason I can't click "approve", but LGTM

Issue #70, opened 2017-05-15
Support more than 1024 harts


mwachs5:
Feedback from the workshop

timsifive:
This might actually be easy:
If a system has more than 1024 harts, the Debug Module must support hawindowsel and hawindow. hawindowsel will be as wide as is required to be able to access every hart.

mwachs5:
So 32*32 is the actual limit then?

mwachs5:
oh wait no its 2^32 * 32. That seems... OK.

sorear:
Possibly relevant: the current [1024-way read mux for halt flags](https://github.com/freechipsproject/rocket-chip/blob/9804bdc34e81ae6e1674a7c0102480f167763443/src/main/scala/devices/debug/Debug.scala#L888) was responsible for [most](https://gist.github.com/sorear/eca052c170c7b754b9f0f06af968388e) of the non-TLMonitor rocket-chip firrtl output when I last did a build a couple months ago.

timsifive:
Continuing this thought, do we need hartsel at all?
When there is just one hart, hawindowsel would be tied to 0, and hawindow tied to 1.
On systems with 32 harts or less hawindowsel would still be tied to 0. hawindow would use its full 32 bits. That's a bit more expensive than the 5 bits otherwise required by hartsel, but if you have that many harts I think you can spare those bits.

The only issue is whether it should be possible to set multiple bits at once or not. hawindow could be RAWL (did I get that right?) to deal with that.

mwachs5:
it is nice to be able to set halt/resume on a particular hart without ambiguity/race.

(that being said, it causes implementation pain to allow hartsel being in the same register as haltreq)

timsifive:
> it is nice to be able to set halt/resume on a particular hart without ambiguity/race.

I agree, but I'm not sure what exactly this comment applies to. If the debugger wants to do that, it should only set a single bit in hawindow.

mwachs5:
right, but that takes two seperate writes

timsifive:
That's true. Is that the only benefit of the hartsel method? That you can change harts and request halt/resume all in one write? That doesn't really seem worth having a separate hart selection mechanism for.

PR #189, opened 2017-12-15
Make \Fhaltsel WARL.

timsifive:
This allows a single-core system to wire all these bits to 0, and lets
eg. a 4-core system only use 2 bits instead of 10.

I had thought this is the way it was, but clearly not.

Since this is a backwards incompatible change, I'm incrementing the
version number.


mwachs5:
Also, may I suggest that we create a "next" branch if we're going to bump the version number. I don't really want to bump the version number for every sentence that we change.

timsifive:
Because "This allows a single-core system to wire all these bits to 0, and lets
eg. a 4-core system only use 2 bits instead of 10."

As stated, this is backwards incompatible.

Currently software checks anynonexistent to check for existence. After this change software will also have to read back hartsel.

mwachs5:
Fair enough, but again I don't think it makes sense to make this a PR into the "0.13" branch

xml/dm_registers.xml
```diff
@@ -175,6 +175,11 @@
         <field name="hartsel" bits="25:16" access="R/W" reset="0">
```
mwachs5: should we use the standard RISC-V "WARL" label here vs "R/W"?

xml/dm_registers.xml
```diff
@@ -175,6 +175,11 @@
...
             The DM-specific index of the hart to select. This hart is always part of the
             currently selected harts.
+
+            When the written value does not specify a hart that exists in the
+            system, this field may take on any value (valid or invalid). Unless
```
mwachs5: what is the meaning of an "invalid" value? Do you mean "it can take on the value of a nonexistent hart"?

timsifive:
I'll start a thread about branching/versioning on the mailing list.

It's not quite a WARL, because an "invalid" value may be read. Here's the use case in my head:
You have a system with 3 harts. The simplest hardware implementation is to assign 2 bits to hartsel, and let the debugger deal with the fact that it can select a non-existend hart. So then you get:
write 0 -> read 0
write 1 -> read 1
write 2 -> read 2
write 3 -> read 3 (and anynonexistent is 1)
write 4 -> read 0
write 5 -> read 1
...

mwachs5:
I think you are saying that "valid == implemented". That's not necessarily true, it could be "valid hart to select, but I'll tell you it's not implemented".

timsifive:
Good point. But in that case, legal and valid are meaningless term for this register. Any bit pattern is legal/valid. I'll fix the "valid or invalid" language to reflect this.

timsifive:
Can I get another review on this?

mwachs5:
Our goal is to get out the current version for ratification before bumping the version. I don't think this change is critical enough to bump the version.

timsifive:
Would you accept the change if we just change the hartsel language without incrementing the version number?

timsifive:
I should clarify that a debugger written to the new language (read back hartsel after writing it) will work fine with hardware built with the current language. Debuggers written for the old language (don't read back hartsel) may not work right with hardware written to the new language.

mwachs5:
Exactly. So debuggers which previously supported "version 0.13" may do wrong things (like halting the wrong hart and not realize it). This isn't something where some functionality is just not available, it will actually do the wrong thing.

timsifive:
Do you agree that this change improves the spec?

mwachs5:
I think there are still some unanswered questions... what if I write hartsel = 3 and observe 2? Should I also check hartsel = 4, or stop? (somewhat looking ahead to the changes proposed for #70)

timsifive:
Good point. I've updated the relevant section in debug_module.tex (which needed it in any case).

timsifive:
Pretty much completely rewrote this. I think it's much less ambiguous now.

debug_module.tex
```diff
@@ -98,7 +98,8 @@ \section{Selecting Harts} \label{selectingharts}
 
 A debugger can enumerate all the harts
 attached to the DM by selecting each hart starting from 0
-until \Fanynonexistent in \Rdmstatus is 1.
+until either \Fanynonexistent in \Rdmstatus is 1, or until the value it reads
+back from \Fhartsel does not match what it wrote there.
```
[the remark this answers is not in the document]
timsifive: It sounds like you're considering letting people tie eg. bits 1 and 9:3 to 0, leaving as the only valid values 0, 1, 4, 5. An implementation could trivially rewire the bits so that bits 9:2 are tied to 0 instead, which makes the spec simpler as well as the debugger implementation.

Did I misinterpret what you said?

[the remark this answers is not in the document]
mwachs5: That was what my comment was saying... I think your change made it more clear that it must be the LSB which are wired.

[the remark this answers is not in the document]
timsifive: I think I see what you're saying.

If we allow tieing arbitrary bits to 0, then we can add more range to hartsel by expanding the register on the LSB side (where we have space), and in old implementations those are simply detected as tied to 0, since that's what the spec says. Is that right?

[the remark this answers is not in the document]
mwachs5: I'm not advocating changing the location of hartsel[0] in the register, even if hartsel[10] ends up at a lower location in DMCONTROL register

debug_module.tex
```diff
@@ -96,9 +96,11 @@ \section{Selecting Harts} \label{selectingharts}
 selects a hart, and then subsequent halt, resume, reset, and debugging
 commands are specific to that hart.
 
-A debugger can enumerate all the harts
-attached to the DM by selecting each hart starting from 0
-until \Fanynonexistent in \Rdmstatus is 1.
+To enumerate all the harts, a debugger must first determine {\tt HARTSELBITS}
```
mwachs5: can we call this `HARTSELLEN` to be consistent with RISC-V `XLEN`, `ILEN`

xml/dm_registers.xml
```diff
@@ -120,6 +120,13 @@
         This register controls the overall debug module
         as well as the currently selected harts, as defined in \Fhasel.
 
+        {\tt HARTSELBITS} (the size of \Fhartsel) is chosen by an
+        implementation depending on its needs.  It must be at least 0 and at
+        most 10. The maximum number of harts a DM can support is $2^{\tt
+        HARTSELBITS}$. A debugger can discover {\tt HARTSELBITS} by writing all
+        ones to \Fhartsel (assuming the maximum size) and reading back the
+        value to see which bits were actually set.
+
```
mwachs5: We may want to clarify a few points:

1) "can support" should perhaps be "can select", or "can give status on". It's unclear that this is not the same thing as how many harts are *actually* existent in the system.
2) Writing a hartsel value greater than 2^HARTSELBITS can result in aliasing to a hart that is supported, which may have unintended effects. So an implementation SHOULD determine HARTSELBITS as part of the probing sequence.

debug_module.tex
timsifive: Yes. Done.

xml/dm_registers.xml
timsifive: 1. How about "can control?"
2. Done.

Issue #150, opened 2017-09-20
figure out sbsingleread, sbautoread


timsifive:
There are 4 use cases, and I don't think they're best served by what we have:
1. Read from consecutive addresses. We'd like to:
   1. Write address (triggering read)
   2. Read data (triggering read+postincrement), ...
2. Read from non-consecutive addresses.
   1. Write address (triggering read)
   2. Read data
   3. Repeat
3. Write to consecutive addresses.
   1. Write address
   2. Write data (triggering write+postincrement), ...
4. Write to non-consecutive addresses.
   1. Write address
   2. Write data (triggering write)
   3. Repeat

The bits we need to achieve that:
1. postincrement
2. Trigger read on address write (addresstrigger?)
3. Trigger read on data read and write on data write (datatrigger?)

Does that sound right?

mwachs5:
Those sound like reasonable use cases. How much overhead is introduced by the current solution?

timsifive:
In our current spec:

1. Read from consecutive addresses.
   1. Write address
   2. Set sbautoread+sbautincrement
   3. Read data N-1 times (triggering increment+read)
   4. Disable sbautoread
   5. Read data
2. Read from non-consecutive addresses.
   1. Write address
   2. Set sbsingleread
   3. Read data
   4. Repeat...
3. Write to consecutive addresses
   1. Set sbautoincrement
   2. Write address
   3. Write N words to data (triggering write+increment)
4. Write to non-consecutive addresses.
   1. Write address
   2. Write data (triggering write)
   3. Repeat

So the reading is a bit more clunky than necessary right now.

mwachs5:
In the proposed solution I think there is something missing, you still have to write some control bits which would say whether you want the auto read & increment behavior to happen, right?

I think you should perhaps keep the current bits autoread and autowrite names, and just slightly change their functionality -- for read when it is set, you want the read to happen on address write or data read. For autowrite, you just want the write to happen on data writes.

timsifive:
Yes. The "proposed solution" was really a wishlist, and doesn't include the setup. Each one of them (probably) will require a single setup write at the start.

There is no autowrite bit in the current spec. It's the standard behavior, and could be in the new version as well.

The current bits are: sbsingleread, sbautoincrement, sbautoread
Proposed bits are: sbautoincrement, sbreadonaddr, sbreadondata

sbautoread is the same as sbreadondata.

[...]

PR #203, opened 2018-01-09
Get rid of sbsingleread in favor of sbreadonaddr

timsifive:
Now writing sbcs ever triggers a bus access, and sbreadonaddr is more
similar to sbreadondata (renamed). This is a backwards incompatible
change. We do have extra bits in sbcs that we could use for versioning.

Issue #150


timsifive:
Per today's meeting, I added a version field.

xml/dm_registers.xml
```diff
@@ -508,15 +508,40 @@
...
+            the system address bus is that wide).
+        </field>
+    </register>
+
     <register name="System Bus Access Control and Status" short="sbcs" address="0x38">
-        <field name="0" bits="31:21" access="R" reset="0" />
-        <field name="sbsingleread" bits="20" access="W1" reset="0">
-            When a 1 is written here, triggers a read at the address in {\tt
-            sbaddress} using the access size set by \Fsbaccess.
+        <field name="version" bits="31:29" access="R" reset="1">
```
mwachs5: shouldn't this be called `sbversion` to avoid alias issues with `\Fversion` macros (maybe not an actual compile issue because we're currently using `versionlo` and `versionhi`). All the other sb-related bits seem to be nicely prefixed with `sb`

mwachs5: Also, why is it 3 bits? vs byte-aligned 4 bits?

xml/dm_registers.xml
```diff
@@ -508,15 +508,40 @@
...
+        <field name="sbbusy" bits="21" access="R" reset="0">
+            When 1, indicates the system bus master is busy. This bit goes high
+            immediately when a read or write is requested for any reason, and
+            does not go low until the access is fully completed.
+
+            To avoid race conditions, debuggers must not try to clear \Fsberror
+            until they read \Fsbbusy as 0.
+        </field>
+        <field name="sbreadonaddr" bits="20" access="R/W" reset="0">
+            When a 1, every write to \Rsbaddresszero automatically triggers a
```
mwachs5: When a 1 -> When 1

xml/dm_registers.xml
```diff
@@ -508,15 +508,40 @@
...
+            system bus read at the new address.
         </field>
         <field name="sbaccess" bits="19:17" access="R/W" reset="2">
             Select the access size to use for system bus accesses triggered by
-            writes to the {\tt sbaddress} registers or \Rsbdatazero.
+            writes to \Rsbaddresszero or \Rsbdatazero.
```
mwachs5: only writes? What about reads to \Rsbdatazero?

xml/dm_registers.xml
```diff
@@ -626,7 +651,7 @@
         \begin{steps}{Writes to this register:}
             \item If the bus master is busy then accesses set \Fsberror, and
             don't do anything else.
-            \item Start a bus write of {\tt sbdata} to {\tt sbaddress}.
+            \item Start a bus write of the new value of {\tt sbdata} to {\tt sbaddress}.
             \item If \Fsbautoincrement is set, increment {\tt sbaddress}.
```
mwachs5: Should we add "sets `sbbusy` until the bus is not busy?

xml/dm_registers.xml
```diff
@@ -636,7 +661,7 @@
             \item ``Return'' the data.
             \item Mark the register stale.
```
mwachs5: there is a lot of abstract discussion about the register being marked stale, but as a concept it's never really introduced (I think). It's sort of confusing because it's not like there is a "stale" bit in a register that can actually be read.

xml/dm_registers.xml
```diff
@@ -586,19 +611,19 @@
         When the system bus master is busy,
         writes to this register will set \Fsberror.
 
-        If \Fsberror is 0 and \Fsbautoread is set then the system bus
+        If \Fsberror is 0 and \Fsbreadonaddr is set then the system bus
```
mwachs5: Should this say "... is set then a *write* to this register will cause the system bus master to start a read at the newly written address.

`\Faddress` is the wrong macro. It should be `\Fsbaddresszero`.

xml/dm_registers.xml
timsifive: Yes, sbversion is better.

3 bits because how many versions do we really anticipate? I'd rather save the other bits for some future use. We can always widen the field later.

xml/dm_registers.xml
timsifive: Fixed.

xml/dm_registers.xml
timsifive: This language is really old. It controls the size of every system bus access. Fixing...

xml/dm_registers.xml
timsifive: Yes. Fixing...

xml/dm_registers.xml
```diff
@@ -636,7 +661,7 @@
             \item ``Return'' the data.
             \item Mark the register stale.
             \item If \Fsbautoincrement is set, increment {\tt sbaddress}.
-            \item If \Fsbautoread is set, start another system bus read.
+            \item If \Fsbreadondata is set, start another system bus read.
```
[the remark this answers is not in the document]
timsifive: Yes. Fixing...

xml/dm_registers.xml
timsifive: Yes. Fixing.

xml/dm_registers.xml
timsifive: Thinking about this more, I don't think we need the stale bit at all. Instead we can just look at the busy bit instead.

xml/dm_registers.xml
```diff
@@ -580,9 +580,7 @@
             3: There was some other error (eg. alignment).
 
             4: The system bus master was busy when one of the
-            {\tt sbaddress} or {\tt sbdata} registers was written,
-            or \Rsbdatazero was read when it had stale data.
-
+            {\tt sbaddress} or {\tt sbdata} registers was written.
```
mwachs5: don't you still mark error if sbdata was read when busy?

xml/dm_registers.xml
```diff
@@ -611,10 +609,9 @@
         When the system bus master is busy,
         writes to this register will set \Fsberror.
 
-        If \Fsberror is 0 and \Fsbreadonaddr is set then the system bus
-        master will start
-        to read after updating the address from \Faddress. The access size is
-        controlled by \Fsbaccess in \Rsbcs.
+        If \Fsberror is 0 and \Fsbreadonaddr is set then a write to this register
+        will cause the system bus master to start
+        a read after updating the address from \Faddress.
```
mwachs5: \Faddress is the wrong macro

debugger_implementation.tex
```diff
@@ -158,19 +158,18 @@ \subsection{Using System Bus Access} \label{deb:mrsysbus}
     \hline
     Op & Address & Value & Comment \\
     \hline
-    Write & \Rsbaddresszero & address & \\
+    Write & \Rsbcs & \Fsbaccess$=2$, \Fsbreadonaddr, \Fsbreadondata, \Fsbautoincrement &
```
mwachs5: In the generated PDF the 'Value' column overflows.

debugger_implementation.tex
```diff
@@ -317,10 +316,10 @@ \subsection{Using System Bus Access} \label{deb:mrsysbus}
     \hline
     Op & Address & Value & Comment \\
     \hline
-    Write & \Rsbaddresszero & address & \\
-    \hline
     Write & \Rsbcs & \Fsbaccess$=2$, \Fsbautoincrement & Turn on autoincrement \\
```
mwachs5: In the generated PDF the Value column overflows

xml/dm_registers.xml
```diff
@@ -650,7 +664,8 @@
         \begin{steps}{Writes to this register start the following:}
             \item Set \Fsbbusy.
             \item Perform a bus write of the new value of {\tt sbdata} to {\tt sbaddress}.
-            \item If \Fsbautoincrement is set, increment {\tt sbaddress}.
+            \item If the write succeeded and \Fsbautoincrement is set,
+            increment {\tt sbaddress}.
             \item Clear \Fsbbusy.
```
mwachs5: you actually only set/clear busy if it wasn't already busy, right? This is a little unclear from this formation of the description.

mwachs5: I guess the sentence above sort of goes with this. Maybe you should say, "If the bus master is busy then accesses set \Fsberror. Otherwise, {steps}"

xml/dm_registers.xml
timsifive: Yes. Fixing.

debugger_implementation.tex
timsifive: Fixed.

debugger_implementation.tex
timsifive: Fixed.

xml/dm_registers.xml
timsifive: Appears to already be fixed.

timsifive:
Just above is the line "If the bus master is busy then accesses set
\Fsberror, and don't do anything else." I'm not sure how to make it any
more clear.

mwachs5:
For sbaddress, registers, are you allowed to read them while busy? sbaddress0 says "accesses to this register will set sberror and don't do anything else". sbaddress1 says "writes to this register will set sbserror and don't do anything else".

I think it's probably fine to read them at any time?

mwachs5:
(I just did a pass over the PDF again... the Table which lists all the registers is too long again. Do we really need to list out all the Abstract Data / Program Buff registers in that table?

timsifive:
On Thu, Jan 18, 2018 at 10:09 AM, Megan Wachs <notifications@github.com>
wrote:

> For sbaddress, registers, are you allowed to read them while busy?
> sbaddress0 says "accesses to this register will set sberror and don't do
> anything else". sbaddress1 says "writes to this register will set sbserror
> and don't do anything else".
>
> I think it's probably fine to read them at any time?
>
There's no real need for a debugger to read sbaddress registers while busy,
so I'm inclined to disallow it. This might allow for more flexible hardware
implementation (although I cannot think of exactly how it helps things).

mwachs5:
in general I think it's harder to make things fail than  to just allow them, and at any rate it was inconsistent. So I changed it so only writes were not allowed, but I'm fine either way as long as all the address words are treated the same.

timsifive:
On Thu, Jan 18, 2018 at 10:14 AM, Megan Wachs <notifications@github.com>
wrote:

> (I just did a pass over the PDF again... the Table which lists all the
> registers is too long again. Do we really need to list out all the Abstract
> Data / Program Buff registers in that table?
>
Not really need, although it's nice to be explicit and complete. (I really
wish LaTeX would have tables that had good default behavior.)

timsifive:
On Thu, Jan 18, 2018 at 12:53 PM, Megan Wachs <notifications@github.com>
wrote:

> in general I think it's harder to make things fail than to just allow
> them, and at any rate it was inconsistent. So I changed it so only writes
> were not allowed, but I'm fine either way as long as all the address words
> are treated the same.
>
Sounds good.

xml/dm_registers.xml
```diff
@@ -603,7 +606,7 @@
         If \Fsbasize is 0, then this register is not present.
 
         When the system bus master is busy, writes to this register will set
-        \Fsberror and don't do anything else.
+        \Fsbbusyerror and don't do anything else.
 
         \begin{steps}{If \Fsberror is 0 and \Fsbreadonaddr is set then writes
```
mwachs5: I believe we need to  augment these sentences (wherever they occur) to include \Fsbusyerror being 0 as well

timsifive: Yes. Fixed.

timsifive:
This fixes #150.

Issue #198, opened 2018-01-03
Expand \Rdata0, \Rprogbuf0 descriptions to cover the other registers


mwachs5:
There is no explicit defintion of registers other than \Rdata0 and \Rprogbuf0. Nowhere do we actually say that the other dataN/progbufN registers exist or not based on datacount/progbufcount. We should upate the 3.11.11 and 3.11.12 descriptions to cover this.

PR #207, opened 2018-01-09
Explicitly list data[1-10] and progbuf[1-15]

timsifive:
The table still *just* fits on the page.

Issue #198


mwachs5:
Can we update the text that goes with progbuf0 and data0 to generalize to progbufN and dataN? I think that was the real spirit of this PR (though I don't mind adding all these too)

mwachs5:
Also to say that these registers don't exist based on progbufsize and datasize

timsifive:
Updated. How's this?

Issue #194, opened 2018-01-03
Clarify what CSRs can be used when HARTINFO.dataccess is 0


mwachs5:
Spec says:
```
datasize: If dataaccess is 0: Number of CSR registers dedi-
cated to shadowing the data registers.

dataaddr: If dataaccess is 0: The number of the first CSR
dedicated to shadowing the data registers.
```

Does this need to clarify what registers are legal for this? RISC-V Priv spec allocates 0x7B0-0x7BF as "debug mode only CSRs", 3 of which are already defined as $dpc and $dcsr and $dscratch0 (note that $dscratch0 doesn't really need a standardized location given the current spec). Currently there is space for a system which decided it wanted to put all 12 data words in that region, but if additional functionality like #187 is actually added, it could conflict with this usage. So we may want to define a space that is "safe" to use for this.

Or should the registers used for this come out of some other space like the "non-standard R/W CSRs" region.

timsifive:
Per yesterday's meeting, Megan will discuss this with the ISA group.

mwachs5:
The "non standard R/W CSRs" can be used as well. I'll update the text to reflect this

mwachs5:
I think the whole point of the "non standard CSRs" is that they are usable for implementation-specific purposes. So stating it again seems a bit redundant.

Issue #187, opened 2017-12-13
Debug "Undelegate" CSRs (Exception Trigger)


mwachs5:
There was a suggestion to add debug-mode CSRs that are roughly analogous to `mideleg`, `medeleg`, which would "undelegate" certain exception types into debug mode. If that exception/interrupt type is encountered, it would trap to the debugger like an `ebreak`. This would require specing out additional CSRs.

This behavior can be emulatabled in software, especially if the debugger is comfortable modifying the trap handler code, but is also an easy to add optional feature (like other delegate CSRs, these could be tied to 0... though other `*deleg` CSRs must actually exist even if tied to 0, and requiring these to exist would be backwards incompatible.)

mwachs5:
This issue is being addressed by a new trigger type instead: see #266

mwachs5:
Resolved by #266

PR #211, opened 2018-01-16
hartinfo: Clarify which CSRs should be used

mwachs5:
Addresses #194.

I thought about clarifying that the "Debug mode only " CSRs should be used, but since these may get allocated for other standard purposes (e.g. #187 ) it's probably not the best to make that the official suggestion.


xml/dm_registers.xml
```diff
@@ -221,8 +221,10 @@
         <field name="dataaccess" bits="16" access="R" reset="Preset">
             0: The {\tt data} registers are shadowed in the hart by CSR
             registers. Each CSR register is XLEN bits in size, and corresponds
-            to a single argument, per Table~\ref{tab:datareg}.
-
+            to a single argument, per Table~\ref{tab:datareg}. These
+	    implementation-specific CSRs should be located in the ``Non-standard''
```
timsifive: It looks like these lines are indented less than the surrounding block.

I would add parens indicating which CSRs are the non-standard range. That saves somebody 5 minutes of digging, and is unlikely to change.

mwachs5: @aswaterman These are optional, and I was thinking we didn't need to standarize their location (since the whole point of this register is to describe where they are).

After thinking about this more, it's really only needs a max of 3 CSRs (I was thinking it would be up to 12, but these CSRs should be XLEN wide, not 32). So I think we should just standardize that if you're going to do it this way its the last 3 Debug CSRs which are XLEN bits wide. Which would render this field in hartinfo sort of pointless.

I'm not sure how you would use this information when the CSR width can change based on writable $misa. But this is already unclear in the spec as written -- how many CSRs do you actually need to access all the $dataN registers.

Will fix the indentation

mwachs5: Ah, I realized that it does say each CSR corresponds to a single argument, so it is pretty clear how that corresponds to XLEN-wide CSRs.

PR #213, opened 2018-02-02
Add support for 2^20 harts, up from 2^10

timsifive:
The main difference is to remove the halt region in favor of 4
registers. This is slightly slower to determine which hart is halted if
their number is between 32 and 1024, but it allows for many more harts
overall, and the time to figure out which one is halted is O(log(N)).

This addresses Issue #70.


xml/dm_registers.xml
```diff
@@ -11,11 +11,18 @@
 
         Harts are nonexistent if they will never be part of this system, no
         matter how long a user waits. Eg. in a simple single-hart system only
-        one hart exists, and all others are nonexistent.
+        one hart exists, and all others are nonexistent. Debuggers may assume
+        that a system has no harts with ids higher than the first nonexistent
```
mwachs5: This is a little misleading since I think you are mixing "id" with "sel value". "ID" was used pretty consistently to mean "the value you get when you read $mhartid" which could even go in descending order or something.

You could say "debuggers may assume than any harts selected by a hartsel value higher than the first nonexistent hartsel will all be nonexistent."

xml/dm_registers.xml
```diff
@@ -11,11 +11,18 @@
...
+        one.
 
         Harts are unavailable if they might exist/become available at a later
-        time.  Eg. in a multi-hart system some might temporarily be powered
-        down, or a system might support hot-swapping harts.
+        time, or if there are other harts with higher ids than this one. Eg.
```
mwachs5: again, mixing ID with sel value is confusing.

xml/dm_registers.xml
```diff
@@ -11,11 +11,18 @@
...
+        in a multi-hart system some might temporarily be powered down, or a
+        system might support hot-swapping harts. Systems with very large number
+        of harts may permanently disable some during manufacturing, leaving
+        holes in the otherwise continuous hart id space. In order to let the
```
mwachs5: samezies.

xml/dm_registers.xml
```diff
@@ -105,15 +112,18 @@
...
-        implementation depending on its needs.  It must be at least 0 and at
-        most 10. The maximum number of harts a DM can control is $2^{\tt
-        HARTSELLEN}$. A debugger should discover {\tt HARTSELLEN} by writing
-        all ones to \Fhartsel (assuming the maximum size) and reading back the
+\label{hartsel}
+\index{hartsel}
+        Throughout this document we refer to \Fhartsel, which is \Fhartselhi
+        combined with \Fhartsello. The awkward ordering is to preserve
+        backwards compatibility. While the spec allows for 20 \Fhartsel bits,
+        an implementation may choose to implement fewer than. The actual width
```
mwachs5: fewer than -> fewer than that.

mwachs5:
This PR should also update this sentence:
https://github.com/riscv/riscv-debug-spec/blob/6d9bccc4cc7c0f82c229bc7fb75b134c2e1939ea/debug_module.tex#L22

mwachs5:
In a system which implements the "Hart Array" mechanism for selecting harts, this seems to overlap with that functionality, if all harts are selected with that mechanism. The problem is you would have to keep changing the hart-array to query "any" harts, then switch back to a smaller subset in order to control them.

I am wondering if there is a more elegant way to combine the two functions (reporting state of single/several/all harts at once, and controlling single/several/all harts at once).

mwachs5:
This proposal could be made basically backwards-compatible for existing systems of <32 harts. Keep most of the changes in the PR, but (I believe) if `haltsummary3` is placed where the old `haltsummary` was, and `haltsummary0` is placed at 0x40, then there is really no change for small number of hart systems. Yes, existing systems don't really implement `haltsummary1` or `haltsummary2`, but debuggers interacting with systems of small numbers of hart wouldn't need to go through all the layers of indirection anyway.

timsifive:
The backwards-compatible version (#218) is better now that we're in the official review period.

Issue #141, opened 2017-09-13
table order can be confusing


timsifive:
From Nikhil:
> Sec 1.2.2, para 1, para 2, table, graphic
> ".. A simple graphic ... After the graphic follows a table ... allowed accesses are listed in Table 1.1"
>
> The order of the paragraphs, graphic and example table is confusing.
>
> The example graphic and example table should be moved up between paras 1 and 2.

LaTeX puts tables where it wants. I'd love it if some expert could improve this.

PR #214, opened 2018-02-05
Move Reg Access Abbrev table after sample register

timsifive:
Fixes issue #141.


introduction.tex
```diff
@@ -48,15 +48,21 @@ \subsection{Structure}
...
+All register definitions in this document follow the format shown below.  A
+simple graphic shows which fields are in the register. The upper and lower bit
+indices are shown to the top left and top right of each field. The total number
+of bits in the field are shown below it.
 
 After the graphic follows a table which for each field lists its name,
 description, allowed accesses, and reset value. The allowed accesses are listed
 in Table~\ref{tab:access}.
 
+Names of registers and their bits are hyperlinks to their definition. If you're
```
mwachs5: "and their bits" -> "and their fields"

mwachs5: suggest: "Names of registers and their fields are hyperlinks to their definition, and are indexed on page \pageref{index}".

timsifive: Done.

PR #216, opened 2018-02-09
Make hart selection consistent, support 2^20 harts

timsifive:
Get rid of the old hart window, halt summary, and hartsel mechanisms. Instead use a new aggregate selection mechanism. Now there is only one mechanism (no separate hartsel and window), that supports up to 2^20 harts on a DM.

Fixes Issue #70.

This change is not backwards compatible.


mwachs5:
I was working on a PR based off to this one previously that made it backwards-compatible. I will make it a separate PR

timsifive:
Sounds good. I made this PR to have something concrete to discuss instead
of hand-wavy e-mails.

timsifive:
Don't want to make such a big incompatible change now we're in the official review period.

PR #218, opened 2018-02-14
Modification of > 1024 hart proposal, maintains backwards compatibility

mwachs5:
This is a modification of #213 which moves a few registers to maintain backwards compatibility with the current spec for systems of small numbers of harts. To address #70.

I considered keeping Allens' original proposal to window into the Hart Halted region, but agree this hierarchical approach is better.


timsifive:
I'd make haltsum show up in two places: 0x40 for backwards compatibility, and 0x32 so targets can implement a 6-bit address space for the DM, which reduces the number of bits scanned in all operations by 2 percent.

timsifive:
Shall I click 'merge'?

Issue #222, opened 2018-02-27
Authentication: What's the relationship to DMACTIVE/DMSTATUS?


mwachs5:
A few questions about what registers can be read/written if AUTHENTICATED is 0.

-- Can we write DMACTIVE ? Must we write DMACTIVE in order to perform authentication?
-- Can we read DMSTATUS.VERSION, and other fields? But not the other fields of DMSTATUS?

Probably some other things that are a bit unclear.

mwachs5:
Also, if authenticated = 1, can the system be reset (either externally or using ndmreset) without corrupting authentication status? If not, it is hard to have an authenticated debug session across reset.

mwachs5:
Closed by #223

PR #223, opened 2018-02-28
Clarify what happens when \Fauthenticated is clear

timsifive:
Proposed solution to Issue #222


debug_module.tex
```diff
@@ -382,9 +382,18 @@ \section{Security}
...
+clear, the DM must not interact with the rest of the platform, nor expose
+details about the harts connected to the DM. All DM registers should read 0,
+while writes should be ignored, with the following mandatory exceptions:
+\begin{steps}{}
+    \item \Fauthenticated in \Rdmstatus is readable.
+    \item \Fauthbusy in \Rdmstatus is readable.
+    \item \Fversion in \Rdmstatus is readable.
+    \item \Fdmactive in \Rdmcontrol is readable and writable.
+    \item \Rauthdata is readable and writable.
+\end{steps}
```
mwachs5: This looks good... one other question, is \Fauthenticated = 0 similar to a reset/dmactive = 0? I mean, does clearing the authentication reset the other debug module registers (or *might* it)?

timsifive: No. I think of authentication as orthogonal to DM reset, or any other DM operation. If you're authenticated you can perform operations on the DM. If you're not authenticated then you cannot. In general I feel things don't have side effects unless they are explicitly mentioned. Do you think it's helpful to add:

DM reset is completely independent of DM authentication.

mwachs5: I ask because I thought we said something about authentication and reset elsewhere in the spec, but I may be mistaken. Let me grep a bit.

timsifive:
dmactive says:
            0: The module's state, including authentication mechanism,
            takes its reset values (the \Fdmactive bit is the only bit
which can
            be written to something other than its reset value).

Maybe that's what you were thinking of? I think that resetting the DM
should reset the authentication as well, if only so we can get out of some
wedged state.

Tim

mwachs5:
Ah yes, that was what I was thinking of. So, you have to first set dmactive, then you can authenticate, and then authentication lasts until Debug Module itself is reset (e.g. by power-on) or by writing dmactive to 0. That makes sense.

mwachs5:
given that, I don't think you need to actually add anything to the PR, since it was just my recollection that was wrong.

PR #224, opened 2018-03-05
Clarify multiple DMs/harts

timsifive:
Added \Rnextdm to allow a debugger to discover multipled DMs, if they exist.

I wasn't sure whether to rewrite every DM reference to possible plural, so I took the lazy route and didn't.


overview.tex
```diff
@@ -15,10 +15,15 @@ \chapter{System Overview} \label{overview}
 OpenOCD, which may include a hardware driver) to communicate with Debug
 Transport Hardware (eg.  Olimex USB-JTAG adapter).
 The Debug Transport Hardware connects the Debug Host to the Platform's Debug
-Transport Module (DTM).  The DTM provides access to the Debug Module (DM) using
-the Debug Module Interface (DMI).
+Transport Module (DTM).  The DTM provides access to one or more Debug Modules
+(DMs) using the Debug Module Interface (DMI).
 
-The DM allows the debugger to halt any hart in the platform. Abstract commands
+Each hart in the system is controlled by exactly one DM. Harts may be
```
mwachs5: system -> platform, since that seems to be the term we are using.

overview.tex
```diff
@@ -15,10 +15,15 @@ \chapter{System Overview} \label{overview}
...
+heterogeneous. There is no further limit on the hart-DM mapping, but usually
+all harts in a single core are controlled by the same DM. In most systems there
```
mwachs5: why do we have to talk about cores at all? Does the RISC-V ISA spec talk about cores?

overview.tex
```diff
@@ -15,10 +15,15 @@ \chapter{System Overview} \label{overview}
...
+will only be one DM that controls all the harts in the system.
```
mwachs5: system -> platform

xml/dm_registers.xml
```diff
@@ -429,6 +429,14 @@
     <register name="Device Tree Addr 2" short="devtreeaddr2" address="0x1b"/>
     <register name="Device Tree Addr 3" short="devtreeaddr3" address="0x1c"/>
 
+    <register name="Next Debug Module" short="nextdm" address="0x1d">
+        If there is more than one DM accessible on this DMI, this register
```
mwachs5: oh, for some reason I assumed these would be chained at the DTM level, not the DebugModule level. I guess this works too and is more efficient than JTAG chaining.

mwachs5: So this suddenly means that all these addresses are actually offsets from some base, and the first one happens to be 0? I don't know if you made that clear.

debug_module.tex
```diff
@@ -23,16 +23,17 @@ \chapter{Debug Module (DM)} \label{dm}
...
 or use a more full-featured bus like TileLink or the AMBA Advanced Peripheral
 Bus. The details are left to the system designer.
 
 The DMI uses between 7 and 32 address bits.  It supports read and write
 operations.  The bottom of the address space is
-used for the DM. Extra space can be used for custom debug devices, other cores,
-additional DMs, etc.
+used for the first (and usually only) DM. Extra space can be used for custom
+debug devices, other cores, additional DMs, etc. If there are additional DMs
+on this DMI, the DMI address of the next DM is given in \Rnextdm.
```
mwachs5: I think you mean the "base address of the next DM on the DMI address space is given in..."

overview.tex
timsifive: Done. Thanks.

overview.tex
timsifive: I added this because on the mailing list and the call there was some confusion about whether there can only be one DM per core or not. I agree that the DM doesn't care at all whether harts are on the same core or not.

overview.tex
timsifive: Done.

xml/dm_registers.xml
timsifive: More DMs can be added at the DTM level also, but this seems like a trivial extension that can be more efficient, and is also DTM-agnostic.

I'll some blurb about a base address somewhere.

debug_module.tex
timsifive: Indeed.

timsifive:
Can you take a look at my updated PR?

PR #225, opened 2018-03-06
travis: add 'make chisel' target to regression

mwachs5:
This currently fails, once the issues are fixed we should add this to regression.


mwachs5:
FYI @rmac-sifive

mwachs5:
@timsifive now that this has been rebased on #218 and no longer actually modifies any spec code, can I get a review?

PR #230, opened 2018-03-10
Remove implemented features from Future Ideas.

### LINKED ISSUE #227: Update "future ideas"
Some ideas are already covered in the current spec:

"Future Ideas, 5 and 9 are already mentioned in the main part of the spec".

timsifive:
Fixes #227.


future.tex
```diff
@@ -15,7 +15,6 @@ \chapter{Future Ideas}
       be halted/run/stepped simultaneously. When a hart hits a breakpoint, all
       the other harts in the same group also halt within a few clock cycles.
    \item DTMs are specified for protocols like USB, I2C, SPI, and SWD.
-   \item Core registers can be read without halting the processor.
    \item The debugger can communicate with the power manager to power cores up
       or down, and to query their status.
    \item Serial ports can raise an interrupt when a send/receive queue becomes full/empty.
```
mwachs5: maybe we should delete this one since there is nothing about Serial Ports in the spec at all, while we're cleaning this up anyway

mwachs5:
Yes, that does seem more like a "notes from the WG" than something that belongs in the spec

timsifive:
Because I'm lazy, and we already have a bunch of LaTeX infrastructure built. Especially the serial registers are much easier this way. I'd have to rewrite them or registers.py to also generate markdown.

timsifive:
What's the verdict here?

timsifive:
Can you comment on this? Is it OK to keep it as LaTeX?

mwachs5:
Sure, the rationale makes sense.

PR #231, opened 2018-03-10
Clarify user responsibilities when debugging lr/sc

### LINKED ISSUE #228: Clarify the LR/SC behavior
We've gotten this question multiple times now, so I think it's worth a commentary at least:

The spec says:

> 4.2 Load-Reserved/Store-Conditional Instructions
> 
> The reservation registered by an lr instruction on a memory address may be lost when entering Debug Mode or while in Debug Mode. This means that there may be no forward progress if Debug Mode is entered between lr and sc pairs.

The question is what are you supposed to do when debugging lr/sc pairs then?

(from @timsifive):

> This doesn't limit where breakpoints can be set, or what code can be stepped through. It does allow an implementation to lose a reservation whenever Debug Mode is entered, so if you have a breakpoint in between lr/sc, or are just single stepping, then it's possible you will never see the reservation occur. To guarantee that you see it, you must let the code run (single stepping doesn't count) so the entire sequence from lr to sc (inclusive) executes without debug interruption. (There may be other system conditions that interfere with the reservation, of course.)

Can we add a version of this explanation as a commentary?

timsifive:
Fixes #228.


core_debug.tex
```diff
@@ -37,6 +37,16 @@ \section{Load-Reserved/Store-Conditional Instructions}
 may be no forward progress if Debug Mode is entered between {\tt lr} and {\tt
 sc} pairs.
 
+\begin{commentary}
+    This is a behavior that debug users must be aware of. If they have a
+    breakpoint set between a {\tt lr} and {\tt sc} pair, or are stepping
+    through such code, they may never leave it.  Fortunately in general use
```
mwachs5: "they may never leave it" -> "the sc may never succeed".

core_debug.tex
```diff
@@ -37,6 +37,16 @@ \section{Load-Reserved/Store-Conditional Instructions}
...
+    there will be very few instructions in such a sequence, and anybody
+    debugging it will quickly notice that the reservation is not occurring.
+    The solution in that case is to set a breakpoint on the first instruction
+    after {\tt sc} and run to it.
```
mwachs5: "after the sc"

timsifive:
Can you review this again? It seems that addressing your feedback invalidated your approval.

PR #235, opened 2018-03-14
hawindowsel can be smaller, depends on # of harts

### LINKED ISSUE #232: hawindow and hawindowsel may be limited depending on number of supported harts
Eg. hawindowsel doesn't need to be more than 1 bit wide if the DM only supports 64 harts.

Address this once #218 has merged.

timsifive:
Fixes #232

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -268,15 +268,8 @@
     <!-- =============== Hart Array Registers =============== -->
 
     <register name="Hart Array Window Select" short="hawindowsel" address="0x14">
-
-      This register selects which of the 32-bit portion of the hart array mask register
-      is accessible in \Rhawindow.
-
-      The hart array mask register provides a mask of all harts controlled by
-      the debug module. It contains one bit for every hart this DM may address.
-      A hart is part of the currently selected harts if
-      the corresponding bit is set in the hart array mask register and
-      \Fhasel in \Rdmcontrol is 1, or if the hart is selected by \Fhartsel.
+      This register selects which of the 32-bit portion of the hart array mask
+      register (see Section~\ref{hartarraymask}) is accessible in \Rhawindow.
 
       <field name="0" bits="31:15" access="R/W" reset="0" />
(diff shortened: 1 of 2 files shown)


xml/dm_registers.xml
```diff
@@ -287,6 +292,9 @@
       The position of the window is determined by \Rhawindowsel. I.e. bit 0
       refers to hart $\Rhawindowsel * 32$, while bit 31 refers to hart
       $\Rhawindowsel * 32 + 31$.
+
+      Bits that are outside the hart array mask for this DM are not writable,
```
mwachs5: do we need to explicitly say that the hart array mask register may not be any larger than the number of harts in the system? Otherwise it's a little unclear what it means to be "outside the hart array mask". Is it a legal implementation to allow, say, the bottom 13 harts to be selected with hart array, but not the rest of the 200 harts on my system?

timsifive: The hart array mask register explicitly may be larger than the number of harts in the system. That allows people to have a DM module that supports 64 harts that they can just plug into a system that has <65 harts.

I guess it would be legal to not have all your harts be selectable, but that doesn't seem like something anybody would want to do. Maybe it's best not to allow it.

timsifive: I've clarified this.

mwachs5: Maybe you should just remove this sentence again, since you clarified above that some bits of the register can be tied to 0. I still think "outside the hart array mask" is unclear. Maybe you could change this sentence to say:

"Since some bits in the hart array mask register may be constant 0, not all bits of this register are always R/W, depending on the current value of hawindowsel".

timsifive: Done (and then edited just a tiny bit).
This is ready for review again. :-)

mwachs5:
This is clear to me at this point. Are there still questions about this behavior?

timsifive:
I'm merging it. We can always clarify it more later if that's necessary.

PR #236, opened 2018-03-14
Clarify abstract command errors.

### LINKED ISSUE #234: CMDERR 4 vs 2 clarification
from Ben Horgan:
3.11.7: abstractcs – when is cmderr 4 used rather than 2? The reference to unsupported/supported when halted is confusing.

timsifive:
Fixes #234.


xml/dm_registers.xml
```diff
@@ -316,14 +316,13 @@
...
-            supported when it is halted.
+            2 (not supported): The requested command is not supported,
+            regardless of whether the hart is running or not.
 
             3 (exception): An exception occurred while executing the command
             (eg. while executing the Program Buffer).
 
-            4 (halt/resume): An abstract command couldn't execute because the
+            4 (halt/resume): The abstract command couldn't execute because the
             hart wasn't in the expected state (running/halted).
```
mwachs5: expected -> required

mwachs5: Is there a clear precedence here? So what if in an implementation that doesn't support Abstract CSR access at any time gets such a request when the hart is running. Which type of error should be reported? That may be more to the point of what #234 was asking

timsifive: I think it's clear that it should be 2, but I'm open to a better way to phrase it.

PR #237, opened 2018-03-14
Be more precise about processor vs hart

### LINKED ISSUE #226: use consistent wording: harts not processor/core
As pointed out on `isa-dev`:

"The language in the spec seems a bit changeable. It talks about ‘halting the processor’ sometimes rather than halting a hart."

timsifive:
Fixes #226

### DIFF
--- a/core_debug.tex
+++ b/core_debug.tex
@@ -7,10 +7,10 @@ of the rest.
 
 \section{Debug Mode} \label{debugmode}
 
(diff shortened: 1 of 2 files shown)


timsifive:
I just looked for every use of "processor" and there weren't that many.

mwachs5:
I think there were a lot more "core"

timsifive:
So there are. Let me take a look.

PR #243, opened 2018-03-16
Change dcsr.prv reset value to 3

timsifive:
This will help avoid a trap where an implementation might end up with
prv 0 when entering debug out of reset. Even in that case prv should be
set from the privilege mode of the processor, which should be machine
mode, but specifying this as the actual value makes things a little
simpler.


mwachs5:
I guess. It's sort of like the reset value doesn't matter, but this is fine with me.

timsifive:
I agree the value doesn't matter, but putting this in the field is less confusing.

PR #245, opened 2018-03-19
Nonexistent/unavailable harts are not halted.

timsifive:


timsifive:
I agree, but Alan asked in e-mail so it wasn't clear enough.

PR #249, opened 2018-03-26
Explain why progbufsize=1 is special

timsifive:


debug_module.tex
```diff
@@ -244,6 +244,13 @@ \section{Program Buffer} \label{programbuffer}
 instruction, or a compressed instruction in the lower 16 bits accompanied by a
 compressed {\tt nop} in the upper 16 bits.
 
+\begin{commentary}
+    The slightly inconsistent behavior with a Program Buffer of size 1 is to
+    accommodate hardware designs that prefer to stuff instructions directly
+    into the pipeline when halted, instead of having the Program Buffer exist
+    in the address space somewhere.
```
mwachs5: so is this specifically saying it MUST be a NOP?

So is it not a legal implementation to build a program buffer that is only 4-bytes of actual memory based RAM? I guess that wasn't even obvious to me when writing the spec, since can't one do quite a bit (e.g. read a CSR) with a 32-bit program buffer?

I guess maybe the confusion is what does it mean "the program buffer may hold only a single instruction". I read that as "the program buffer is maybe only capable of holding a single instruction", but in your commentary I'm realizing maybe you meant "the program buffer shall hold only a single instruction". What does it mean then if the debugger tries to write 2 compressed instructions? Is Debug Module supposed to quash the upper one? 

What I'm getting at is who are these restrictions imposed on, the HW implementation or the debugger?

timsifive: You can build a program buffer that is 4 bytes of RAM, but the debugger can not write 2 compressed instructions into it.

The debugger must not write 2 compressed instructions. If it does it's out of spec and behavior is undefined. (I'd expect both to be executed, or only the one in the lower 16 bits to be executed, but the debugger mustn't do that.)

timsifive:
Yes, when progbufsize=1 then the Program Buffer is allowed to only hold a single instruction. Because that is allowed, the debugger must not assume it can write 2 instructions.

mwachs5:
so as a hardware designer though, I am *allowed* to write a program buffer that doesn't care about this, right? So what you are saying is this is ONLY a restriction on the debugger, not a requirement on the hardware.

timsifive:
Correct.

timsifive:
Thanks. I've incorporated that.

Issue #252, opened 2018-04-09
SB Access: should there be an error code for unsupported SB Access size?


mwachs5:
Currently the spec calls out SB error code 3 to be "some other error, e.g. alignment". What about the error case when you try to do an unsupported sbaccess size? Does that fall under "some other error"? It seems like a common enough case that it should be specifically tied to some value, even if it is also 3.

mwachs5:
Oh I see you tried to clarify this here: https://github.com/riscv/riscv-debug-spec/pull/203/commits/b8eb62a53e9616e078d71ee264cf73528ee87ef2

But that is not really in a place that someone might look for it when implementing SBERROR logic

mwachs5:
But then it is also unclear the priority of these errors. What if I send a misaligned access to an inaccessible address, which error type should I return? Either is fine?

timsifive:
I'm updating \Fsberror to also mention unsupported access size for 3.

mwachs5:
Resolved by #254

PR #254, opened 2018-04-10
Clarify SBA unsupport access size error.

timsifive:
Maybe solves #252.

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -620,14 +620,19 @@
             While this field is non-zero, no more system bus accesses can be
             initiated by the debug module.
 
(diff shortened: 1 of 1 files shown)


timsifive:
They're definitely arbitrary. I don't really care what the division is. In general it's nice to give users specific errors, though. What do you think the divisions should be?

timsifive:
How about:

0: There was no bus error.
1: There was a timeout.
2: A bad address was accessed.
3: Alignment error
4: Unsupported size
7: Other. (An implementation may report Other for all error conditions.)

We've got the bits. Let's use them.

mwachs5:
I like Other's description. Or you could say "an implementation may report "Other" for Any error condition"

Issue #242, opened 2018-03-14
How to enforce entering debug mode on reset


mwachs5:
Example: Debugger wants to trigger on some bad scenario, so it places  a breakpoint on the "bad scenario handler". Let's say bad scenario generally happens shortly after reset. Then core gets reset (by someone pressing reset button, or SW reset... essentially something out of debugger's control). Debugger will notice the hart got reset because of `hasreset`, but as the spec is currently written the HWBP set up to trigger on bad scenario could be lost and run past before debugger notices.

timsifive:
Simplest solution:
Add a reset-proof halt-on-reset bit to DCSR in each hart. When 1, the hart will enter debug mode immediately out of reset.

timsifive:
I'd like to get this feature in before we ratify.

I think either bit 27 in dmcontrol, or bit 3 in DCSR. Preference? It feels like it belongs more in dmcontrol so it's easier to think of it as not being reset when the hart is.

mwachs5:
If it's in `dmcontrol`, it's global. What does that mean then w.r.t. each hart?

timsifive:
Good point. If it's in dmcontrol, there would have to be set-halt-on-reset
and clear-halt-on-reset bits. DCSR seems more clean from that point of
view. Is there any issue with saying this one bit persists across reset?

timsifive:
That makes me think the bits have to be in DM then, so some way of
accessing them will probably be dmcontrol. Do you agree?

PR #255, opened 2018-04-11
resethaltreq: Proposal for forcing a hart into debug mode out of reset

mwachs5:
Addresses #242 

Adds an optional bit in DMCONTROL analogous to `haltreq`, but is `resethaltreq`. This bit is not self-clearing, but only has impact when the hart's reset is deasserted.


mwachs5:
I'm sketching out this implementation in rocket-chip. An immediate problem I see is that the debugger will have to make sure not accidentally clear this down as it's doing all its other operations, since it applies to all selected harts, but it's not clear how to read it back (the same problem exists for `haltreq` and `resumereq`, but those aren't usually long-lived bits, and they are specified as W-only).

mwachs5:
I see in the discussion for #242 you mentioned having 2 separate "set" and "clear" bits. I guess that's why.

debug_module.tex
```diff
@@ -140,6 +143,16 @@ \section{Run Control} \label{runcontrol}
 (How this is implemented is not further specified. A few
 clock cycles will be a more typical latency).
 
+The debugger can provide an optional \Fresethaltreq bit. Writing to this bit
```
timsifive: Debug Module, I assume

mwachs5:
So I started to edit this to be set and clear but then I couldn't figure out how to make it optional. Are we ok with making this non-optional?

timsifive:
I prefer to make it optional because if you have a lot of harts, routing all the signals could be a pain. Simplest way is to add a bit to dmstatus that, when 1, indicates the feature is supported.

mwachs5:
@timsifive can I get an updated review? I think I did the intended set/clr changes.

debug_module.tex
```diff
@@ -140,6 +143,19 @@ \section{Run Control} \label{runcontrol}
 (How this is implemented is not further specified. A few
 clock cycles will be a more typical latency).
 
+The debug module can implement optional halt-on-reset bits for each hart,
+which it indicates by setting \Fhasresethaltreq to 1.
+This means the debugger implements the \Fsetresethaltreq and \Fclrresethaltreq bits.
```
timsifive: DM, not debugger

debug_module.tex
```diff
@@ -140,6 +143,19 @@ \section{Run Control} \label{runcontrol}
 (How this is implemented is not further specified. A few
 clock cycles will be a more typical latency).
 
+The debug module can implement optional halt-on-reset bits for each hart,
```
timsifive: Debug Module is capitalized

debug_module.tex
```diff
@@ -140,6 +143,19 @@ \section{Run Control} \label{runcontrol}
...
+which it indicates by setting \Fhasresethaltreq to 1.
+This means the debugger implements the \Fsetresethaltreq and \Fclrresethaltreq bits.
+Writing 1 to \Fsetresethaltreq sets the halt-on-reset request bit for each
+selected hart.
+When a hart's halt-on-reset request bit is set, the hart will immediately enter
+debug mode on the next deassertion of its reset. This is true regardless of
+whether or not the hart's reset is caused by the debugger's
```
timsifive: DM's

Issue #229, opened 2018-03-09
Clarify: Are NMI Masked during Debug Mode


mwachs5:
I think the answer is "no": you'd handle it like an exception in Debug Mode. I need to go read the priv spec to see how it defines NMI, but either way this should be clarified in the spec.

mwachs5:
Closed by #256

PR #256, opened 2018-04-11
NMI Clarification

mwachs5:
Addresses #229


xml/core_registers.xml
```diff
@@ -71,7 +71,10 @@
 
             Other values are reserved for future use.
         </field>
-        <field name="0" bits="5:3" access="R" reset="0" />
+        <field name="0" bits="5:4" access="R" reset="0" />
+        <field name="nmip" bits="3" access="R" reset="0">
+            When set, there is a Non-Maskable-Interrupt (NMI) pending for the hart.
```
timsifive: Can you add something like:
Since NMIs indicate a hardware error condition, reliable debugging may no longer be possible once this bit becomes set.

Issue #240, opened 2018-03-14
Suggestion: Allow Debug Mode to be not just M mode


mwachs5:
Spec says in Debug Mode, all instructions operate as if in M mode, and MPRV is ignored. Suggestion is to add WARL bits (presumably in DCSR) to change the prv mode in which Debug Mode operates.

markuslauterbach:
Allowing the debugger to display and edit memory from different "privilege perspectives" is an often requested feature. I second that proposal.

timsifive:
@markuslauterbach by a different privilege perspective, are you interested in anything beyond the address translation? Part of me feels the address translation is well defined in the spec, and could just as easily be done by the debugger. That would be slower due to the memory reads required, but keeps the hardware simpler.

mwachs5:
Other things that might matter:
* Certain CSRs/ instructions are not accessible in different modes (for example, `mstatus` register is not accessible in anything but M-mode, so if we allow changing the priv level, does that mean we can't access `mstatus` register in Debug Mode). This is *more* restrictions than provided by the normal `MPRV`, and seems like a good way to block yourself into a corner from which you can't recover from. The omnipotent debugger can just not do those things.  So let's not make that part of the spec.

* In addition to translation there are Phyiscal memory protections and other access protections, which I think is one thing @markuslauterbach was referring to. Let's say you change the hypothetical "DPRV" to "U" and then access a memory address that the PMP settings don't allow in U-Mode. The core would get an exception, which in Debug Mode doesn't transfer the hart to the usual exception handler after updating all the `mepc`/mcause registers, it just sets the "Exception" code in `ABSTRACTCS.error`. So then debugger would have to know why it got that exception anyway.

My point is that the debugger is going to have to be pretty smart about the memory system, either way, so I tend to agree with @timsifive that the SW can and should handle this rather than adding HW that probably has a bunch of corner cases.

 Another corner case... probably we don't want anything but Debug Mode to be able to access the Debug Module/Program Buffer space. but now we have to punch a hole through for "User Mode Debug Mode" and make sure that it can it access it as well? I don't even know how that works with address translation. Does it mean that HW must enforce that debug module space is always mapped to the same place in the different memory modes?

yungchinghsiao:
[...]

PR #265, opened 2018-04-26
mprven: Add a bit to enable MPRV to take effect in debug mode

mwachs5:
Addresses #240


aswaterman:
lgtm

timsifive:
Can we just remove the restriction that mprv is ignored in Debug Mode? That makes Debug Mode more like M Mode, and we don't need an extra bit at all.

mwachs5:
That would be a major backwards incompatible change

core_debug.tex
```diff
@@ -13,7 +13,12 @@ \section{Debug Mode} \label{debugmode}
 \begin{steps}{When executing code from the Program Buffer, the hart stays
     in Debug Mode and the following apply:}
 \item All operations are executed at machine mode privilege level, except that
-    \Fmprv in \Rmstatus is ignored.
+    \Fmprv in \Rmstatus may be ignored according to \Fmprven.
+    \begin{commentary}
+      In general, debugger SW is expected to be able to virtualize all the effects of \Fmprv.
```
timsifive: simulate instead of virtualize?

timsifive: "the debugger" instead of "debugger SW"

PR #266, opened 2018-04-27
Specify an Exception Trigger

timsifive:
I tried to get all the sources from the Syntacore proposal in, but did
not add the various timing options. We do get interrupt sources now as
well as the exception sources.


xml/hwbp_registers.xml
```diff
@@ -32,6 +32,9 @@
             3: The trigger is an instruction count trigger. The remaining bits
             in this register act as described in \Ricount.
 
+            4: The trigger is an exception trigger. The remaining bits
+            in this register act as described in \Ricount.
```
mwachs5: is \Ricount the right pointer here? Should be \Retrigger?

xml/hwbp_registers.xml
```diff
@@ -289,4 +292,85 @@
...
+
+        <field name="type" bits="XLEN-1:XLEN-4" access="R" reset="2" />
+        <field name="dmode" bits="XLEN-5" access="R/W" reset="0" />
+        <field name="0" bits="XLEN-6:24" access="R" reset="0" />
+        <field name="source" bits="23:10" access="R/W" reset="0">
+            For each bit that is set, the corresponding interrupt/exception
+            will cause this trigger to fire if bit corresponding to the current
+            mode is also set.
+
+            0: Software interrupt
```
mwachs5: I think we need to keep this parallel to the actual $mip/$mie/$mdeleg registers. Can we use \tdatatwo as the mask?

The ISA spec allows additional interrupt sources, and maybe they'll shove something in between some bits that are currently reserved, it seems brittle to re-assign the bit numbers here. Then we can be more generic and say "the coresponding bit in *ip *ie"

mwachs5: And perhaps we just want two separate trigger types, one for interrupt vs exception, or maybe we have bits in this register that say whether the mask applies to exceptions and/or interrupts

xml/hwbp_registers.xml
```diff
@@ -289,4 +292,85 @@
...
+            10: Environment call
+
+            11: Instruction page fault
+
+            12: Load page fault
+
+            13: Store/AMO page fault
+        </field>
+        <field name="m" bits="9" access="R/W" reset="0">
+            This trigger only fires in M mode if this bit is 1.
```
mwachs5: why the "this trigger only fires in only" vs "this trigger fires in"? What if all 3 bits are set, the logic makes no sense

xml/hwbp_registers.xml
```diff
@@ -289,4 +292,85 @@
...
+        </field>
+        <field name="0" bits="8" access="R" reset="0" />
```
mwachs5: thanks for keeping this

xml/hwbp_registers.xml
```diff
@@ -289,4 +292,85 @@
...
+            If this optional bit is implemented, the hardware sets it when this
+            trigger matches. The trigger's user can set or clear it at any
+            time. The trigger's user can use this bit to determine which
+            trigger(s) matched.  If the bit is not implemented, it is always 0
+            and writing it has no effect.
+        </field>
+        <field name="action" bits="4:0" access="R/W" reset="0">
+            Determines what happens when this trigger matches.
+
+            0: Raise a breakpoint exception. (Used when software wants to use the
```
mwachs5: if these actions are going to be common among various types of triggers, perhaps we can just make a table of these actions and refer to it from all the trigger types

xml/hwbp_registers.xml
timsifive: I started out with that, but then it turned out a good portion of the ones in *cause are "... from user mode", "... from machine mode" etc. Since we already have S/U/M bits that seemed like unnecessary duplication, and then we need extra language to explain that "external interrupt from user mode" won't actually cause a trigger to fire unless the U bit is also set. And there's a break exception which seems pointless to have here so I excluded it completely.

We could put all the bits in data2, using the low 16 bits for exceptions and high 16 bits for interrupts. Or even put interrupts in data3.

What do you think?

xml/hwbp_registers.xml
timsifive: Sounds like "only" is a poor word choice there. The intent is that if you want the trigger to fire in M mode, this bit has to be 1. Otherwise it won't. If you also want it to fire in S mode, you also need to set the S bit.

xml/hwbp_registers.xml
timsifive: Yeah. I regret that we don't have a standard location for them. They really belong with dmode at the top of each trigger, but I don't want to move them now. Same is true for the hit bit.

Also, 6 bits is way too much, so here I dropped it to 5. But even that seems like more than we'll ever need.

xml/hwbp_registers.xml
mwachs5: I think the "mode" in *cause are misleading, as noted by frequent questions about it on the mailing list. For example "Supervisor software interrupt" vs "Machine software interrupt" are sort of conventional naming that allow SW to route and delegate things a certain way, not something that the hardware really knows the difference between (it just knows they are different).

mwachs5: The only one I saw that was explicitly "from X-Mode" is "environment call". 

In general I think we should just keep the parallel structure to avoid having to track changes to the priv spec here as things are potentially added and removed.

mwachs5: And obviously some bits would be WARL if there is no code defined for that, which again parallels the ISA spec.

xml/hwbp_registers.xml
mwachs5: Oh. Yeah I read that wrong. How about "This trigger fires in M mode if and only if this bit is 1."

timsifive: Actually, let me just copy from \Rmcontrol:
When set, enable this trigger in S mode.

xml/hwbp_registers.xml
timsifive: There are also the 3 interrupts (software, timer, external), each of which explicitly have from M/S/U mode.

mwachs5: That's what I'm saying, that's really just a naming convention. There is no enforcement in the spec that those thigns really happen in M/S/U mode.

xml/hwbp_registers.xml
```diff
@@ -3,6 +3,36 @@
...
+    \caption{\Faction encoding}
+    \label{tab:action}
+    \begin{tabular}{|r|L|}
+    \hline
+    Value &amp; Description \\
+    \hline
+    0 &amp; Raise a breakpoint exception. (Used when software wants to use the
+    trigger module without an external debugger attached.) \\
+    \hline
+    1 &amp; Enter Debug Mode. (Only supported when \Fdmode is 1.) \\
```
mwachs5: does the caveat of \Fdmode make sense outside of the context of this table?

xml/hwbp_registers.xml
```diff
@@ -148,24 +180,7 @@
...
-
-            3: Stop tracing.
-
-            4: Emit trace data for this match. If it is a data access match,
-            emit appropriate Load/Store Address/Data. If it is an instruction
-            execution, emit its PC.
-
-            Other values are reserved for future use.
-        </field>
+        <field name="action" bits="17:12" access="R/W" reset="0" />
```
mwachs5: I think here you say something like "this is action, as defined by Table X"

xml/hwbp_registers.xml
```diff
@@ -3,6 +3,36 @@
     prevent untrusted user code from causing entry into Debug Mode without the
     OS's permission.
 
+    \Rtdataone always contains a \Faction field, but the position varies
```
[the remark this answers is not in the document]
mwachs5: This should probably be completely separate from this XML file.

xml/hwbp_registers.xml
```diff
@@ -270,23 +285,86 @@
             by 1.
         </field>
 
-        <field name="action" bits="5:0" access="R/W" reset="0">
-            Determines what happens when this trigger matches.
-
-            0: Raise a breakpoint exception. (Used when software wants to use the
-            trigger module without an external debugger attached.)
+        <field name="action" bits="5:0" access="R/W" reset="0" />
```
mwachs5: Again I think you still need to point to the Action table inside the description here.

xml/hwbp_registers.xml
```diff
@@ -270,23 +285,86 @@
...
+    </register>
 
-            1: Enter Debug Mode. (Only supported when \Fdmode is 1.)
+    <register name="Exception Trigger" short="etrigger" address="0x7a1">
+        This register is accessible as \Rtdataone when \Ftype is 4.
 
-            2: Start tracing.
+        If implemented, it allows triggering on a variety of interrupt and
+        exception sources. If such a trigger is taken, all the usual CSRs as
```
mwachs5: I think "the usual CSRs" is a little vague. Why not say CSRs are updated as usual as described by the RISC-V Priv Arch Manual Section XYZ (we should be pointing to a specific version of the spec anyway).

xml/hwbp_registers.xml
```diff
@@ -270,23 +285,86 @@
...
+        well as \Rpc are updated, and the requested action is taken just before
+        the first instruction of the interrupt/exception handler is executed.
 
-            3: Stop tracing.
+        The exceptions to trigger on are specified in \Rtdatatwo. Each of the
```
mwachs5: This  is still very brittle. The ISA manual allows users to define higher-bit interrupts all the way to XLEN. I think we should just define \Rtdatatwo to be the full XLEN width, and not try to share it like this between exceptions and interrupts.

xml/hwbp_registers.xml
```diff
@@ -270,23 +285,86 @@
...
-            execution, emit its PC.
+        \begin{table}[H]
+        \centering
+        \caption{\Retrigger exceptions in \Rtdatatwo}
+        \label{tab:etrigger}
+        \begin{tabular}{|r|l|}
+        \hline
+        Bit &amp; Description \\
+        \hline
+        0 &amp; Instruction address misaligned \\
```
mwachs5: I don't think we should reproduce this here. We should just point to the relevant section of the ISA manual, and be clear which version of the ISA manual we're talking about

mwachs5: And if we were keeping this, why isn't it in `Rtdatatwo` description? Why are we squeezing it into this register's description?

xml/hwbp_registers.xml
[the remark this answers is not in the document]
timsifive: I'm changing this. Not sure where better to move it to, though.

xml/hwbp_registers.xml
timsifive: I don't quite understand your question, but I've clarified the language a little.

xml/hwbp_registers.xml
timsifive: Done.

xml/hwbp_registers.xml
timsifive: Done.

xml/hwbp_registers.xml
timsifive: I've referred to the doc. I don't think the version is relevant until they standardize something.

xml/hwbp_registers.xml
```diff
@@ -3,6 +3,33 @@
...
+    \caption{\Faction encoding}
+    \label{tab:action}
+    \begin{tabular}{|r|L|}
+    \hline
+    Value &amp; Description \\
+    \hline
+    0 &amp; Raise a breakpoint exception. (Used when software wants to use the
+    trigger module without an external debugger attached.) \\
+    \hline
+    1 &amp; Enter Debug Mode. (Only supported when the trigger's \Fdmode is 1.)
```
mwachs5: What I meant was, does every trigger type have a \Fdmode defined?

xml/hwbp_registers.xml
```diff
@@ -271,22 +286,61 @@
...
 
-            2: Start tracing.
+        This trigger may also fire on up to XLEN of the Exception Codes defined
+        in \Rmcause (described in the Privileged Spec, with Interrupt=0). Those
+        causes are configured by writing the corresponding bit in \Rtdatathree.
+        (Eg. to trap on an illegal instruction, the debugger sets bit 2 in
+        \Rtdatathree.)
 
-            3: Stop tracing.
+        As with all other trigger functionality, hardware may only implement a
```
mwachs5: only implement -> implement only

xml/hwbp_registers.xml
```diff
@@ -271,22 +286,61 @@
...
+        subset of these options. A debugger must read back \Rtdatatwo and
+        \Rtdatathree after writing them to confirm the requested functionality
+        is actually supported.
 
-            4: Emit trace data for this match. If it is a data access match,
-            emit appropriate Load/Store Address/Data. If it is an instruction
-            execution, emit its PC.
+        The trigger only fires if the hart takes an exception because of the
```
mwachs5: exception -> trap? I need to double check the language used in the priv spec

xml/hwbp_registers.xml
```diff
@@ -271,22 +286,61 @@
...
+        cause. (Eg. it does not fire when a timer interrupt occurs but that
+        interrupt is not enabled in \Rmie.)
 
-            Other values are reserved for future use.
+        When the trigger fires, all CSRs are updated as defined by the
+        Privileged Spec, and the requested action is taken just before the
+        first instruction of the interrupt/exception handler is executed.
+
+        <field name="type" bits="XLEN-1:XLEN-4" access="R" reset="2" />
```
mwachs5: shouldn't type be 4?

mwachs5:
Do we need to clarify anything about how this interacts with delegation or lack thereof? I mean, do the M/S/U bits apply to the mode you were in when you got the interrupt, or the mode that would have handled the interrupt?

timsifive:
Yes. Thanks. I've attempted to clarify this.

timsifive:
Can I get a review on this?

mwachs5:
I'm fine with it, but should we point people to it before merging it to the spec? Perhaps in the email thread that Gadge sent out (since it does address some of requests)? Or discuss in the meeting?

timsifive:
Generally nobody ever comments on these PRs. I'd rather just merge this. I'm open to changing it later if people have better suggestions. I'm also all for advertising that this change is going/went in, but historically that doesn't actually get us any feedback.

timsifive:
If #279 goes in, I'll change this to be two trigger types which only use a single addition tdata register.

Issue #258, opened 2018-04-13
What happens if you write SBCS while SBCS.busy=1


mwachs5:
If you attempt to write SBCS while SBCS.busy is 1, it's not clear what happens. Should the write be ignored, cause an error, or should debuggers just not do that (as they are already told not to attempt to clear `SBCS.sberror` until `SBCS.busy` is 0?

mwachs5:
Closed by #270

PR #270, opened 2018-05-02
Debuggers must not write sbcs while sbbusy is set

timsifive:
Issue #258

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -583,7 +583,7 @@
             reason, and does not go low until the access is fully completed.
 
             Writes to \Rsbcs while \Fsbbusy is high result in undefined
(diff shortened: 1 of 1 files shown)


xml/dm_registers.xml
```diff
@@ -582,8 +582,9 @@
             bit goes high immediately when a read or write is requested for any
             reason, and does not go low until the access is fully completed.
 
-            To avoid race conditions, debuggers must not try to clear \Fsberror
-            until they read \Fsbbusy as 0.
+            Writes to \Rsbcs while \Fsbbusy is high result in undefined
+            behavior.  A debugger must not write to \Rsbcs until they read
```
mwachs5: They read -> it reads

timsifive: Thanks.

PR #271, opened 2018-05-02
Reading tselect doesn't guarantee a valid trigger.

timsifive:


xml/hwbp_registers.xml
```diff
@@ -10,8 +10,8 @@
 
         Writes of values greater than or equal to the number of supported
         triggers may result in a different value in this register than what was
-        written.  Debuggers should read back the value to confirm that what
-        they wrote was a valid index.
+        written.  Debuggers must read back the value when they are not sure
```
mwachs5: I don't know if "must" is right. It implies that the HW won't work right unless the debugger reads it back. I think "should" was fine here.

timsifive: But they only "must" "when they are not sure."

mwachs5: I mean, they could do it 10 minutes later if they *wanted* to. It feels weird to dictate what SW *must* do in this way when there is no HW requirement that they must do it...

timsifive: What I'm trying to convey is that when software wants to select trigger 4, but it doesn't know for sure that trigger 4 exists, it *must* read back tselect after writing it. Because tselect may not contain the value 4 even if that's what the software wrote, which indicates trigger 4 (and beyond) don't exist.

How can I better get that point across?

mwachs5: You could say "to verify that what they wrote is a valid index, debuggers can read back the value to confirm <the field> holds what they wrote"

timsifive: OK. I did that (pretty closely).

PR #279, opened 2018-05-21
Make trigger types writable.

timsifive:
This gives a hardware more options when implementing triggers. Instead
of requiring separate triggers for separate types, they could opt to
reuse the configuration storage.


xml/hwbp_registers.xml
```diff
@@ -61,6 +65,22 @@
         <field name="data" bits="MXLEN-1:0" access="R/W" reset="Preset" />
     </register>
 
+    <register name="Trigger Info" short="tinfo" address="0x7a4">
+        <field name="0" bits="XLEN-1:16" access="R" reset="0" />
+        <field name="info" bits="15:0" access="R" reset="Preset">
+            One bit for each possible \Ftype enumerated in \Rtdataone. Bit N
```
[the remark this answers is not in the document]
timsifive: We are out of bits in tdata1. (Note that 16 bits are required, which is a lot.) We could remove tdata3 if using an extra CSR is too painful.

xml/hwbp_registers.xml
```diff
@@ -61,6 +65,22 @@
...
+            corresponds to type N. If the bit is set, then that type is
+            supported by the currently selected trigger.
+
+            If the currently selected trigger doesn't exist, this field
+            contains 1.
+
+            For backwards compatibility, if this field is 0 then the trigger
+            only supports a single type, which the debugger can read from
+            \Rtdataone.
```
aswaterman: This is a new CSR, so the field won't be 0 on old implementations; it'll trap.

aswaterman: Actually, is tinfo really necessary? Can we just probe which trigger types a trigger supports by making tdata1.type WARL, so software can probe which values are supported by writing to tdata1.type and reading it back?

aswaterman: Need to address the backwards compatibility issue, regardless.  The probing algorithm needs to account for tinfo causing an exception.

timsifive: I've replaced reading 0 with a trap.

PR #283, opened 2018-06-01
Abstract memory

timsifive:
This is an alternative to #284.


debug_module.tex
```diff
@@ -217,6 +217,8 @@ \subsection{Abstract Command Listing}
         \hline
         1 & Quick Access & \pageref{quick access} \\
         \hline
+        2 & Access Memory Command & \pageref{access memory} \\
```
mwachs5: should we call this something like "Access Memory from Hart Perspective" (or a more catchy phrase?)

debugger_implementation.tex
```diff
@@ -298,6 +298,46 @@ \subsection{Using Program Buffer} \label{deb:mrprogbuf}
     \end{tabulary}
 \end{table}
 
+\subsection{Using Abstract Memory Access} \label{deb:mrabstract}
```
mwachs5: same note: should we clarify here that this is from the perspective of the selected hart?

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
         <field name="0" bits="23:0" />
     </register>
 
+    <register name="Access Memory">
```
mwachs5: same note, can it be clarified that this is from the selected hart's perspective?

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
...
+        \begin{steps}{This command gives the debugger access to memory
+            locations, including memory-mapped registers. It performs the
+            following sequence of operations:}
+        \item Copy data from the memory location specified in {\tt arg1} into the
+            {\tt arg0} portion of {\tt data}, if \Fwrite is clear and \Ftransfer is set.
```
mwachs5: do we need to care about `transfer` bit? or is this just for symmetry with the abstract registers?

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
...
+        \item Copy data from the {\tt arg0} portion of {\tt data} into the
+            memory location specified in {\tt arg1}, if \Fwrite is set and \Ftransfer is set.
+        \item Increment {\tt arg1}, if \Fpostincrement is set.
```
mwachs5: should we name these bits something less generic, like `amapostincrement` (abstract memory access post increement) .To avoid confusion with `sbaincrement` bits or the like.

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
...
+        \item Execute the Program Buffer, if \Fpostexec is set.
+        \end{steps}
+
+        If any of these operations fail, \Fcmderr is set and none of the
```
mwachs5: Are there additional cmderr types introduced for "abstract memory access" that we would need to define? Looking at SBA there are some other memory-access specific error types that aren't currently listed in `cmderr`.

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
...
+
+        \begin{commentary}
+            The encoding of \Fsize was chosen to match \Fsbaccess in \Rsbcs.
+        \end{commentary}
+
+        <field name="cmdtype" bits="31:24">
+            This is 2 to indicate Access Memory Command.
+        </field>
+        <field name="0" bits="23" />
+        <field name="size" bits="22:20">
```
mwachs5: again should we make these bitfield names less generic? I guess if we did in theory they should be renamed for access register as well... but what does `\Fsize` link to even in this document when you refer to it? You'd want it to link here for this command type but to the other table for the other command type.

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
...
+            0: Access the lowest 8 bits of the memory location.
```
mwachs5: what about misalignment? Is it an error? Undefined?

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
...
+
+            1: Access the lowest 16 bits of the memory location.
+
+            2: Access the lowest 32 bits of the memory location.
+
+            3: Access the lowest 64 bits of the memory location.
+
+            4: Access the lowest 128 bits of the memory location.
+
+            If \Fsize specifies a size larger than the memory location's actual size,
```
mwachs5: what does that mean? "The memory location's actual size"? Does that mean if you are running off the end, or if there is a misalignment error? Or just copy-paste confusion?

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
...
+            This field controls the Argument Width as referenced in
+            Table~\ref{tab:datareg}.
+        </field>
+        <field name="postincrement" bits="19">
+            After a memory access has completed, if this bit is 1, increment
+            {\tt arg1} (which contains the address used) by the number of bytes
+            encoded in \Fsize.
+        </field>
+        <field name="postexec" bits="18">
+            When 1, execute the program in the Program Buffer exactly once
```
mwachs5: Do we need this?

xml/abstract_commands.xml
```diff
@@ -94,4 +94,82 @@
...
+            after performing the transfer, if any.
+        </field>
+        <field name="transfer" bits = "17">
```
mwachs5: do we need this?

debug_module.tex
timsifive: I think this is fine here, but I do agree that in other places we need to be more explicit about what's going on.

xml/abstract_commands.xml
timsifive: I'm removing it.

xml/abstract_commands.xml
timsifive: Yes. Done.

xml/abstract_commands.xml
timsifive: I've added a "bus error" state. We could define to command to modify arg0 to reflect a more precise error (alignment, timeout, ...) when there is a bus error. Do you think that's a good idea?

xml/abstract_commands.xml
timsifive: Done, for size at least.

xml/abstract_commands.xml
timsifive: I'm adding a sentence that basically says that the behavior is whatever it is if the hart is running M mode code.

xml/abstract_commands.xml
timsifive: Getting rid of this in favor of above-mentioned sentence.

xml/abstract_commands.xml
timsifive: No. Removed.

xml/abstract_commands.xml
timsifive: Ditto.

debugger_implementation.tex
```diff
@@ -183,6 +185,10 @@ \subsection{Using System Bus Access} \label{deb:mrsysbus}
 
 \subsection{Using Program Buffer} \label{deb:mrprogbuf}
 
+Through the Program Buffer, the hart performs the memory accesses. Addresses
+are physical or virtual (depending on \Fmprven and other system
```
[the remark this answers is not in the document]
timsifive: I specifically had in mind the value of the mprv bit itself, but I assume there are other ways that address translation can be affected so I didn't want to just write down that one bit.

PR #284, opened 2018-06-01
Add busview bit to SBA.

timsifive:
This is an alternative to #283.


mwachs5:
I think #283 is the cleaner way to go for this functionality (yes I know this was my suggestion but I've been brought to the other side).

timsifive:
Abandoning this approach in favor of #283.

PR #286, opened 2018-06-06
datacount cannot be 0

timsifive:
It must be at least 1 so GPRs can be accessed.


xml/dm_registers.xml
```diff
@@ -353,7 +353,7 @@
         <field name="0" bits="7:4" access="R" reset="0"/>
         <field name="datacount" bits="3:0" access="R" reset="Preset">
             Number of {\tt data} registers that are implemented as part of the
-            abstract command interface. Valid sizes are 0 - 12.
+            abstract command interface. Valid sizes are 1 -- 12.
```
mwachs5: Maybe we should just change this to a `commentary` that only sizes make sense to be useful.

palmer-dabbelt: I like the commentary option.

timsifive:
I've updated the absolute minimums to reflect how wide of a hart can be supported like that. I'm not sure what you want to see in the commentary.

In the end the actual decision involves the abstract commands an implementation chooses to implement (to determine how many arguments are needed), and the max XLEN harts they wish to support. Perhaps I should just put that in there? (I'm not sold on formatting it as commentary, since then it ends up somewhere outside the table.)

palmer-dabbelt:
After considering this, I don't think this is a valid change to make this late in the process.  Even though the wording change is pretty small it's actually a big restriction on the allowable hardware implementations as it becomes mandatory to support at least one data register.

I vote we close the issue without merging anything.

timsifive:
Supporting accessing GPRs through abstract commands is already mandatory, which in turn requires at least one data register. An implementation cannot be compliant with a datacount of 0.

Section 3.5.1.1 (Access Register):
> Debug Modules must implement this command and must support read and write access to all GPRs
when the selected hart is halted.

This is part of the program buffer vs abstract command compromise we settled on.

mwachs5:
I think we can leave it as is, with commentary at most:

“In practice, in order to implement abstract commands, in particular the access GPR abstract command, data count cannot be 0” 

The commentary can go anywhere in the text so the argument that it looks weird formatted in the table is pretty moot. 

To me this is more clear than putting a lot of equations in the place where we say what the legal values of data count are.

timsifive:
I've moved the XLEN discussion into a commentary, but left datacount valid sizes to range 1-12 (which is the original change that started this PR).

timsifive:
We could remove "Valid sizes are 1 -- 12" but making it read "0-12" is just wrong. It can never be zero, just like it can't be 13.

mwachs5:
Good point. There is a mismatch between the way/reason the upper and lower bounds are set. I guess I’m fine with this then

Issue #290, opened 2018-06-09
Add abstract command to deal with cache coherence


timsifive:

timsifive:
On the June 13, 2018 debug call nobody thought this should be included right now.

PR #292, opened 2018-06-11
Add Memory Visibility abstract command.

timsifive:
This lets a debugger flush caches to ensure a consistent view of all
memory across all harts. (See #290.)

There may be a better way to think about this, but I want to get the ball rolling with a concrete proposal.


timsifive:
No consensus on whether we need this and what exactly it should look like.

PR #298, opened 2018-06-16
Define MXLEN.

timsifive:


timsifive:
#287

timsifive:
Almost. In the priv manual it's XLEN per the *current* value of the MXL field in mstatus. MXLEN here is the reset value of MXL. (Ignoring the possibility that MXL is allowed to be 0.)

mwachs5:
Hmmm, does it HAVE to be different? If we’re going to give it a different meaning then we should probably use a different term...

timsifive:
We need something that doesn't change depending on what code is running.

Trigger registers (which comprise almost all of the spec's use of MXLEN) must be at least the max possible XLEN width so you can set triggers for any mode. The document currently treats MXLEN as the max possible XLEN. (I don't think I realized MXLEN is writable when I switch the document over.)

We need the same thing for dpc, and the GPRs when in Debug Mode.

Is there a term for that already?

timsifive:
I'm thinking we should just define DXLEN: XLEN of GPRs in Debug Mode, which is in turn the value of MXLEN as reset.

What do you think?

mwachs5:
I like that suggestion. It is what we actually mean and works nicely with treating debug mode as anther execution mode.

mwachs5:
It’s still a little fuzzy for trigger registers though. Say something is writable as DXLEN, but is actually going to be compared against something SXLEN wise (because S bit is set). We need to be clear about what happens in that case. And what happens when M mode writes the trigger when MXLEN < DXLEN

timsifive:
This feels like an important clarification to get into the spec. Can I get a review?

timsifive:
Could I get a review of this? @mwachs5 @allenjbaum

allenjbaum:
I'm missing something here. 
 "define DXLEN: XLEN of GPRs in Debug Mode, which is in turn the value of MXLEN as reset."
First, I can't find where the 20Aug spec says what the XLEN of GPRs are in debug mode, for register access, program buffer instruction execution or anything else.
   Is it the case that it always works to use the largest supported XLEN? (which means the debugger has to interrogate DCSR.PRV to ensure that it ignores the upper half of register reads etc.)

The debug spec defines some of what happens on entry to debug mode  - (sec 4.1 bullet points 1..12) but not to the extent that the priv spec specifies what happens on trap handler entry (e.g. pc   ->dpc, , mode-> <x>PP, or which of the fields of mstatus should be in effect (besides MPRV)
I assumed it basically uses Mstatus, as modified by DCSR, but don't see any statement to that effect.

So: either that wording exists and I can't find it, or should be added - but that's a separate issue.

So: this change does have the effect of defining a term for the widest XLEN supported by the implementation (which is the prev spec definition of what is in  MXLEN at reset)  and ensuing that registers will (primarily triggers) will be that wide.

timsifive:
What I'd like is that in Debug Mode, the hart acts as if XLEN is the maximum supported XLEN. If mstatus says the effective XLEN is less than that, the debugger will have to decide how to present that to the user.

Does that clear things up?

allenjbaum:
Ok, that works for me. 

-Allen

allenjbaum:
In a separate email, Tim said he'd rather that debug mode defaulted to the widest (==reset) value of MXL, and that the debugger should deal with the case that MXL is RW and interrogate mstatus to see what the effective value is and use that to handle narrower value explicitly. Its a bit more work for the debugger, but it works, and I'll support it

xml/hwbp_registers.xml
```diff
@@ -79,19 +83,23 @@
...
+
+        If XLEN is less than DXLEN, writes to this register are sign-extended.
+        <field name="data" bits="XLEN-1:0" access="R/W" reset="Preset" />
     </register>
 
     <register name="Trigger Data 3" short="tdata3" address="0x7a3">
         Trigger-specific data.
-        <field name="data" bits="MXLEN-1:0" access="R/W" reset="Preset" />
+
+        If XLEN is less than DXLEN, writes to this register are sign-extended.
```
mwachs5: "writes to this register" so you mean if I am in M-mode, and MXLEN < DXLEN, then the M-mode write will be sign-extended? I'm not saying you need to add more text, just making sure I understand your intention.

timsifive: Yes. Maybe the paragraph at the beginning of the trigger section should mention that there are always DXLEN storage bits. This line is saying that when writing it when XLEN<DXLEN, all storage bits are in fact updated, and the high bits are determined by sign extending.

introduction.tex
```diff
@@ -26,6 +26,8 @@ \section{Terminology}
 A single RISC-V core contains one or more hardware threads, called
 \emph{harts}.
 
+\emph{DXLEN} of a hart is the effective XLEN in M-mode, at reset.
```
mwachs5: Is this assuming that XLEN in M-mode at reset is the largest possible XLEN (perhaps MXLEN can be increased or SXLEN can be greater than XLEN in M-mode at reset).

The fact that the priv spec makes no rules about these things is certainly complicating our spec, but we should make our assumptions clear.

timsifive: The privspec says: "The MXL field is always set to the widest supported ISA variant at reset."

But we might as well be explicit here. I'll change that.

mwachs5: ah, if it says that, excellent. I think this is good then.

PR #299, opened 2018-06-19
Document breakpoint exception + enter debug mode

timsifive:


timsifive:
Can I get a review on this?

timsifive:
This addresses #288.

allenjbaum:
To see if I understand this, when a hart gets triggers that cause both breakpoint exception actions and enter debug mode actions:
If breakpoint is handled first, it takes the exception (so saves PC in EPC, etc), then enters debug mode (saves the exception handler PC into DPC, etc).
If enter debug mode occurs first, it enters debug mode (saves the exception handler PC into DPC, halts, etc) - and the breakpoint exception is ignored? pended, to be taken after resume from debug mode?
This is a little different from an exception occurring as a result of an action taken by debug mode, but the same thing could happen if debug mode touched a memory location which had a trigger set on it.
Normally, an exception during debug mode results in a cmderr being set and the debug action terminated, without updating  any other architectural state (like a pend bit). In this case, does cmderr get set (and to what?), even though there is no debug action to be terminated? The "hit" bit is supposed to detect this case, but it is optional, so its possible that the breakpoint trigger is completely lost .

timsifive:
If enter debug mode occurs first, and ideally the breakpoint exception is still pending. However, the language allows for the breakpoint exception to be ignored.

What happens to triggers while in Debug Mode is a separate issue. I'll make a PR to discuss that.

Issue #300, opened 2018-06-26
The Debug Module Debug Bus Registers table is confusingly placed inside the dmstatus section


timsifive:
Reported by Bruce Ableidinger

PR #304, opened 2018-06-27
Force the register-address table in place.

timsifive:
Previously it was confusing that especially Table 3.8 would end up in
the dmstatus Section. #300


timsifive:
It results in more whitespace, but less reordering. Thanks.

PR #305, opened 2018-06-27
Completing progbuf exec is I/O for fence insts.

timsifive:
We need something like this so a debugger can ensure a memory change is
propagated to all harts in a system.


mwachs5:
@aswaterman does this language make sense to you?

PR #309, opened 2018-07-11
Solution to native triggers in M mode only systems

timsifive:
This came out of today's debug call.

### DIFF
--- a/trigger.tex
+++ b/trigger.tex
@@ -36,6 +36,34 @@ separately.
     \end{enumerate}
 \end{commentary}
 
+\section{Native M-Mode Triggers}
+\label{sec:mmtrigger}
+
+Triggers can be used for native debugging. On a fully featured system triggers
+will be set using \Fu or \Fs, and when firing they can cause a debug exception
+to trap to a more privileged mode. It is possible to set triggers natively to
+fire in M mode as well. In that case there is no higher privilege mode to trap
+to. When such a trigger causes a debug exception while already in a trap
+handler, this will leave the system unable to resume normal execution.
+
+On full-featured systems this is a remote corner case that can probably be
+ignored. On systems that only implement M mode, however, it is recommended to
+implement one of two solutions to this problem. This way triggers can be useful
+for native debugging even M mode code.
+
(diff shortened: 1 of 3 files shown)


trigger.tex
```diff
@@ -36,6 +36,34 @@ \chapter{Trigger Module}
     \end{enumerate}
 \end{commentary}
 
+\section{Native M-Mode Triggers}
+\label{sec:mmtrigger}
+
+Triggers can be used for native debugging. On a fully featured system triggers
+will be set using \Fu or \Fs, and when firing they can cause a debug exception
```
mwachs5: There is no such thing as “debug exception”. To avoid confusion we should use the same term as the priv spec, which I think is breakpoint exception.

trigger.tex
```diff
@@ -36,6 +36,34 @@ \chapter{Trigger Module}
...
+to trap to a more privileged mode. It is possible to set triggers natively to
+fire in M mode as well. In that case there is no higher privilege mode to trap
+to. When such a trigger causes a debug exception while already in a trap
+handler, this will leave the system unable to resume normal execution.
+
+On full-featured systems this is a remote corner case that can probably be
+ignored. On systems that only implement M mode, however, it is recommended to
+implement one of two solutions to this problem. This way triggers can be useful
+for native debugging even M mode code.
```
mwachs5: Even-> Of even

trigger.tex
```diff
@@ -36,6 +36,34 @@ \chapter{Trigger Module}
...
+
+The simple solution is to have the hardware prevent triggers with action=0 from
+firing while in M mode and while interrupts are disabled. Its limitation is
```
mwachs5: Can we be more specific what you mean “when interrupts are disabled”? MIE in mstatus=0?

trigger.tex
timsifive: Yes. Thanks for catching that.

trigger.tex
timsifive: Done.

trigger.tex
timsifive: Done.

timsifive:
Can I get a review on this?

PR #310, opened 2018-07-12
Add size to mcontrol.

timsifive:
This resolves problems when matching against instructions that aren't
MXLEN bits wide, and allows matches against data accesses of specific
sizes.

### DIFF
--- a/riscv-debug-spec.tex
+++ b/riscv-debug-spec.tex
@@ -53,6 +53,7 @@
 \input{sw_registers.tex.inc}
 
 \deffieldname{\Fhartsel}{\hyperref[hartsel]{hartsel}}
+\deffieldname{\Fsize}{\hyperref[sizelo]{size}}
 \deffieldname{\Faction}{action}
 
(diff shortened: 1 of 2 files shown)


timsifive:
Can I get a review for this?

mwachs5:
Didn’t triggers already have a mechanism for data accesses of certain siZes?

timsifive:
In Section 5.1 (Trigger Registers) is the following:
> Most trigger functionality is optional. All tdata registers follow write-any-read-legal semantics. If a
> debugger writes an unsupported configuration, the register will read back a value that is supported
> (which may simply be a disabled trigger). This means that a debugger must always read back
> values it writes to tdata registers, unless it already knows already what is supported. Writes to
> one tdata register may not modify the contents of other tdata registers, nor the configuration of
> any trigger besides the one that is currently selected.

I believe this covers this field (and every other field) being WARL. Let me know if you think it needs to be made more explicit.

PR #321, opened 2018-07-31
Clarify that resumereq is not level-sensitive.

timsifive:


timsifive:
What I'm trying to say is that writing 1 to resumereq causes harts that are halted at the moment of the write to start a resume. If a hart halts later, before dmcontrol is written again, that hart doesn't instantly resume just because the last write to resumereq was 1.

Does that make sense?

How can I state that more clearly in the spec?

allenjbaum:
That sounds like it’s edge sensitive. 
Both cases need a reset condition.
For level sensitive designs it’s when the level drops. If it doesn’t then you keep paying attention to it- and you specifically say you shouldn’t.

More interesting is the reset condition for edge sensitive. Usually, the edge is converted to a level  by going into the set input of an SR flop, and that is fed back to inhibit the set condition. 

The reset condition is likely when the hart goes to running or is in running- another edge vs level case.


So, two scenarios: halted core is resumed, halts again, but won’t resume again until resumereq is lowered then raised again.

The second is if a core is running and the resumereq goes high. Then it halts- if the reset condition is a level (is running) then the SR flop never got set (assume R overrides S). If the reset condition is edge (going to running) then it did get set, and it will resume right away.

So, I would say you want it to be edge triggered, and run state disables/ resets the edge condition. 

I do seem to recall that resumereq while running is ignored, so that takes care of part of it, but it doesn’t take into account the overlap case. Running asserted -> Resumereq assert -> Running deasserted

So: the hart resumes if Resumereq becomes asserted while the hart is in halted state is what you want.

-Allen

timsifive:
It's not level sensitive, but it's not exactly edge sensitive. Consider:
1. hart is halted
2. write resumereq=1
3. hart resumes, hits a breakpoint and is halted again
4. write resumereq=1
5. hart resumes again

If it was truly edge sensitive a write of resumereq=0 would be required in between, but that's not the case. It's the act of writing 1 that prompts an action to be taken.

allenjbaum:
And now as I think about it more,  I am a more confused. `
One thing not made explicit in the spec is the priorities of the different
commands - what if more than one req bit is set at a time?
Sometimes it doesn't matter, since haltreq is ignored when halted, and
resumereq is ignored when running. If both are set when the state is
neither halted nor running (inreset for example), then what

And, although resumereq acts as if writing it causes a pulse, the spec
specifically says that haltreq must be cleared before a hart is resumed.
That implies it is not pulsed, and also implies that it has priority over
resumereq.
hartreset is similar - it must be manually cleared, so it is level
sensitive, and I'm guessing it has priority over both haltreq and resumereq.

AckHaveReset will clear the havereset status - but what if stays high
between resets ? IS it pulse or level
  hartreset is sent
  reset occurs & havereset is asserted
  hart starts running and shartrest or some other reset cause happens -
ackhavereset is still high: does havereset stay unasserted?
Do I need to write ackhavereset to 1 again, even though it is currently 1?
It is logically pulse or a level?

What happens when a hart is resumed, then is halted, but resumereq is not
lowered? Does resumeack stay high when halted?

I'm going to try to draw up a complete state diagram of the 4 commands
(runreq, haltreq, hartreset, ackhavereset - the set/clearresethaltreq seems
to be orthogonal)
and the 6?7?  legal states: running, halted, running+resumeack, halted+
resumeack?, running+havereset, halted+havereset: there
By the way, what happens when both set and clrhaltresetreq are asserted at
the same time? I am guessing that whatever the value is, it only takes
effect immediately after (some) reset is deasserted

allenjbaum:
I've been re-reading this more closely. In order to clear the resumeack
status, you need to clear resumreq - that makes it a level, not an edge.
So, it acts like a level for some things, and an edge for others. This is
getting kind of nasty

timsifive:
Just to take a step back, the problem that we're solving is:
1. Debugger writes resumereq=1
2. Hart starts running
3. Hart hits breakpoint and halts

The hart must stay halted at that point. If resumereq is level sensitive then it would immediately resume again. We could define it as edge sensitive, but then you're forcing the debugger to (potentially) do an extra write with resumereq=0 even if it does just want to resume again right away. As you point out above, the Debug Module can trivially create a pulse when a write happens so it seems unnecessary to force the debugger to do it in that case.

It makes sense to have haltreq/resumereq behave the same way. The current wording is definitely confusing on that.

It doesn't bother me that dmcontrol is not idempotent.

Reset is completely orthogonal. It took a lot of back-and-forth to get that to where it is now. I'd like to keep it as-is, unless there's some problem with the mechanism.

I'll update the PR to make haltreq and resumereq more symmetrical.

Issue #315, opened 2018-07-24
Can hartsel be changed while an abstract command is executing?


jhauser-us:
Is changing `hartsel` forbidden while `abstractcs.busy` is set?  Or must a DM support the possibility of `hartsel` being changed (perhaps to examine `hartinfo` or one of the halt summary registers) while an abstract command is still executing for a hart?

timsifive:
You can't change `hartsel` while it's being used. Please let me know if #322 adequately explains that.

PR #322, opened 2018-07-31
Can't change harts during operations, and the current hart becoming unavailable may terminate the abstract command with error.

timsifive:
See #315.


jhauser-us:
I think `hasel` needs to be included too.

Concerning `hawindow`, you have
> `hawindow` may not be written while it is in use by eg. an abstract command being executed, or pending halt or resume requests.

Seems to me it's more clearly said to be the Hart Array Mask that's in use here, not `hawindow` exactly.

Also, since the Hart Array Mask is ignored for abstract commands, does an abstract command really count as a "use" of it?  I'm fine with prohibiting writes to `hawindow` during an abstract command if you want, but better just to say that instead of that `hawindow` is in use, which seems untrue.

I would suggest spelling out exactly all the actions during which modification of `hartsel`, `hasel`, and the Hart Array Mask is prohibited, so as not to leave the term "in use" open to interpretation.

timsifive:
I'm walking this back to the narrower request you made. reset/halt requests should be able to continue while a selection is changed, because reset/halt might take a long time if a hart is temporarily powered down, hung on a bus access, etc...

allenjbaum:
Does the attempt to change hartsel during an abstract command cause cmderr to be set?
IF so, that should be documented

timsifive:
What's nice about setting cmderr is it allows a debugger to eg. optimistically write the PC on every hart as fast as possible, without checking that the command has completed.

On the other hand, requiring cmderr to be set might be tricky to implement (now it's written by a different thing, possibly from a different clock domain), and might be too complex of a requirement to force on minimum implementations.

What do you think?

allenjbaum:
How is this different from the case of cmderr= (1 (busy): An abstract command was executing while command, abstractcs, abstractauto was written, or when one of the data or progbuf registers was read or written (and now add: or when dmcontrol.hartsel is changed.

timsifive:
Good point. It isn't really different.

jhauser-us:
It sounds like a debugger will be allowed to change `hartsel` while an abstract command is executing, only that the debug module must set `cmderr` to 1 (busy) if that happens.  Is that what you intend?  If so, what's expected of the debug module with regards to the incomplete abstract command?  Is the DM required to cleanly continue the abstract command for the original hart even when `hartsel` is changed in the middle?  Or does the state of either or both the original and newly selected harts become unknown?

Or, alternatively, did you mean for `hartsel` to be read-only while `abstractcs.busy` = 1?

mwachs5:
I meant to put this comment here:
This is the only one that is in DMCONTROL, so far. The fact that DMCONTROL handles debug-across-reset makes it a bit of a special case. At any rate, it is ALWAYS easier on the hardware to say “debuggers, don’t do this” vs requiring a specific error type to be handled and reported. The places where we do require that should be with some use cases in mind (discovery, efficient bursts, etc).

jhauser-us:
> At any rate, it is ALWAYS easier on the hardware to say “debuggers, don’t do this” vs requiring a specific error type to be handled and reported. The places where we do require that should be with some use cases in mind (discovery, efficient bursts, etc).

I agree.  (Although it's possible to make some exceptions when the cost is small.  Specifying that `haltreq` beats `resumereq` in `dmcontrol` isn't absolutely needed, for example, but it literally only costs a couple of gates.  So one can easily go either way on it.)

timsifive:
It sounds like we're headed towards "Debuggers must not change hartsel while an abstract command is executing." What happens if an abstract command never completes, eg. because something is wedged inside a hart. Do they have to reset? It would be nice to still be able to access other harts if that were the case.

jhauser-us:
If the debugger and hardware are both operating correctly, I'm not aware of a reason why abstract commands should ever get stuck.  If it does happen, I think the easiest way to deal with it is for the debugger to reset the hart.  To justify a more sophisticated alternative, you'd need at a minimum a normal use case that requires that the stuck hart not be reset.

The document ought to have a statement somewhere that if a hart is executing an asbstract command and the hart gets reset for any reason, then the abstract command is terminated and `abstractcs.busy` returns to 0.  Probably `cmderr` should be set to something too, such as 4, indicating "wrong state".  In case the reset wasn't triggered by the debugger itself, it can find an explanation in `dmstatus.anyhavereset`.

mwachs5:
Is this PR still live, or deprecated in favor of #324?

debug_module.tex
```diff
@@ -222,6 +210,21 @@ \section{Abstract Commands} \label{abstractcommands}
...
+    without error.  In the common case the debugger will be much slower than
+    the target and commands succeed, which allows for maximum throughput. If
+    there is a failure, the interface ensures that no commands execute after
+    the failing one.  To discover which command failed, the debugger has to
+    look at the state of the DM (eg. contents of \Rcommand) or hart (eg.
+    contents of a register modified by a Program Buffer program) to determine
+    which one failed.
+\end{commentary}
+
+If a hart is reset while executing an abstract command, then the abstract
```
mwachs5: "If a hart is reset..." do you mean via a write to \Fhartsel, or by any means including a non-debugger controlled system wide reset? 

This is a backwards incompatible change to add this requirement. Also, the error doesn't really make sense. We don't really have abstract commands fail when a hart is in reset, so far. Maybe THAT's the bug, but maybe you want to be able to read registers in a hart while it's in reset...

timsifive: The way I imagine this working is that the DM knows that an abstract command is executing, and if it sees havereset or something go high during that time, it knows a reset occurred during that time. In that case the DM can assume that the abstract command execution got lost,  so it sets error, and the debugger can try again, etc.

This all assumes that the hart is the thing that ends up executing the abstract command. If the DM is doing it all, then it still seems fairly straightforward to implement.

The only alternative I can think of is that somehow the abstract command completes even though the hart got reset. That seems potentially really complex to implement.

allenjbaum:
In the case that the command completed, and then the hart reset:
  - in a HW inplementation, if havereset is set before busy is cleared,
then no error, else error.
 - when the hart is performing the command, then it depends on how busy is
reported.
But, regardless: what are the consequence if the debugger always assumes
the command did not complete if it sees an unexpected havereset?
The only case that matters is when the command accessed something that was
not idempotent, which sounds unusual for a debugger to do.
There are other cases that are worse and probably more common, e.g. a reset
in code using locks - a reset could cause a lock sequence to deadlock.
This reset/abstract command case should b handled in a similar way ("if it
hurts when you do that, don't do that - or least, be careful")

debug_module.tex
mwachs5: I don't think we should *mandate* that the command has to end in error if the hart is reset. If an implementation supports a sideband read of a register, then that may and should work if the hart is in reset, not just give an error. This is useful functionality in general.

I think it's a separate case as to the debug module realizing that an abstract command will never complete because a hart is reset, but that is not ALWAYS the case.

mwachs5: Perhaps a better solution is that the debugger MAY report the "wrong state" error, or some new error meaning "we're not sure if that worked or not". The previously-named "halt/resume" error was a clear indicator that the debug module didn't do the command at all, because the hart wasn't in the right state. This is more of "something went wrong while we were doing that so we don't know what happened" sort of case.

timsifive: The main thing I want to prevent is abstract commands/the DM hanging due to a hart being reset at the wrong time. Currently we can abort a hanging abstract command by resetting the hart. If you allow abstract commands while reset is asserted, then I don't see how you can kick a hart that's hung in an abstract command. (Abstract commands might hang because of an infinite loop in the Program Buffer, or perhaps if a reset causes some protocol problems between the hart and DM.)

I'm not against allowing register access during reset, but I don't think it's essential, since a debugger can always access registers immediately out of reset and then access them.

I'm fine adding an error code that an implementation may set if an abstract command was interrupted by reset, but also giving implementations the option to have the command complete (possibly returning bogus results).

mwachs5: I agree with pretty much all of this, that we need some elegant way to cancel a command or have it cancel itself if something bad is detected. But I do think we should keep in mind the "abstract" part of the abstract command--- we shouldn't assume it has anything to do with a hart's execution since in theory halt is supposed to be able to do anything to a hart, like completely gate its clock, to allow registers to be examined.

>"If you allow abstract commands while reset is asserted, then I don't see how you can kick a hart that's hung in an abstract command. "

I think you have it a little backwards... what you are really asking for is a way to cancel the abstract command. If you allow hart reset while a abstractcs.busy=1, then you can reset the hart. But what we don't have right now is a way to hard-clear abstrtactcs.busy. Saying that hartreset is the way to do that seems a little weird. Why not just have a "kill abstractcs, and I don't care what the consequences are"? Either way, it's going to potentially screw up the debug module's view of the world, so you may have to resort to dmactive level reset.

allenjbaum: Has there been consensus on unexpected reset? 
The other case is a hanging abstract command (without unexpected havereset).
That could be the result of a HW issue, or it could be SW (e.g. the abstract command is running program buffer code spinning on a lock that has been lost). The former case is fatal; (it should never happen) and be the resultof a system wide hang, in which case even a directed hartreset isn't good enough. 
The latter is livelock, not deadlock - but there is no way to determine whether that is the case or not. Even if you could cancel the command (and I would suggest writing a command of allzero be the way to signal that - register_access w/no transfer and no postexec) you can't know if it were livelock or deadlock and must assume the worst. 

And, you can't "cancel" an external transaction;  it leaves microachitectural state hanging around that you have no control over which will cause hangs further down the road.
So: my conclusion is that a hanging abstract command is fatal, and a directed hartreset is the least you can do to recover.

debug_module.tex
```diff
@@ -227,13 +215,32 @@ \section{Abstract Commands} \label{abstractcommands}
...
 While an abstract command is executing (\Fbusy in \Rabstractcs is high), a
 debugger must not change \Fhartsel, and must not write 1 to \Fhaltreq,
 \Fresumereq, \Fackhavereset, \Fsetresethaltreq, or \Fclrresethaltreq.
 
+If a hart is reset while executing an abstract command then the abstract
+command is terminated, \Fbusy must go low, and \Fcmderr should be set to 6
+(interrupted).
+
+If an abstract command appears to be hung, a debugger can only regain control
+by resetting the hart.
```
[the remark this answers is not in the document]
timsifive: I assumed that the DM would notice that the hart reset (using havereset), and clean up whatever needs to be cleaned up on its side.

xml/dm_registers.xml
```diff
@@ -126,7 +126,10 @@
         width of \Fhartsel is called {\tt HARTSELLEN}. It must be at least 0
         and at most 20. A debugger should discover {\tt HARTSELLEN} by writing
         all ones to \Fhartsel (assuming the maximum size) and reading back the
-        value to see which bits were actually set.
+        value to see which bits were actually set. Debuggers must not change
+        \Fhartsel while an abstract command is executing. (If an abstract
+        command appears to be hung, the debugger can reset the selected hart by
```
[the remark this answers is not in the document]
timsifive: You're right, it is always true. The caveat was intended to clarify when a debugger might want to reset the hart while an abstract command is running.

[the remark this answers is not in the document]
timsifive: This is really just a line to answer the question the previous one might bring up, "how do I change hartsel if an abstract command is executing and is hung?" Removing the clause makes me wonder why the line is there in the first place. What do you think?

debug_module.tex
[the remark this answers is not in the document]
mwachs5: No, I'm saying there may be other problems in the system (e.g. bus lock up) that means that even resetting the hart might not cause the debugger to regain control.

As a concrete example, in rocket-chip if the bus locks up (e.g. there's broken peripheral that just doesn't respond to bus transactions), potentially a hart can't execute any Program Buffer/Debug ROM instructions. Even if that hart was reset, it would still not be able to access those instructions.

Keep in mind that "hart reset" was an optional concept only added to the spec after a lot of debate, and was really only added there to save a bit for it. i don't think we should rely too heavily on the concept of "resetting a hart" for key debugger functionality.

debug_module.tex
```diff
@@ -229,13 +217,32 @@ \section{Abstract Commands} \label{abstractcommands}
...
+\end{commentary}
+
 Before starting an abstract command, a debugger must ensure that \Fhaltreq,
 \Fresumereq, and \Fackhavereset are all 0.
 
 While an abstract command is executing (\Fbusy in \Rabstractcs is high), a
 debugger must not change \Fhartsel, and must not write 1 to \Fhaltreq,
 \Fresumereq, \Fackhavereset, \Fsetresethaltreq, or \Fclrresethaltreq.
 
+If a hart is reset while executing an abstract command then the abstract
```
mwachs5: with regards to the discussion in #354 , should this be "if a hart becomes unavailable while executing an abstract command"?

I'll say it again: we're mixing two things here: what happens to an abstract command if a hart becomes unavailable? And how can I cancel a running abstract command (For whatever reason)?

You can already cancel a running abstract command by resetting the debug module (via DMACTIVE). So let's not mix the desire to have a way to do that with the potential error case of what happens when a hart becomes unavailable/reset during the command which is executing.

debug_module.tex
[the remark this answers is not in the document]
timsifive: There are definitely problems that won't be solved by resetting a hart, but those kinds of problems shouldn't get fixed by resetting the DM. The DM should notice if one of its harts is reset, and if it's currently interacting with that hart, it needs to make sure to clean up state regarding to that interaction.

allenjbaum:
The kinda of problems caused by a hung transaction are extremely likely to be solved only by NDMReset.
 Resetting a hart would need to reset state just for the specific transaction that was hung, throughout whatever network connects the requestor and target. Resetting anything else will cause other transactions to hang as well.

There are probably some very simple systems where hartreset  would work on its own, but not many

-Allen

debug_module.tex
jhauser-us: Yes, it will have to be "if a hart becomes unavailable...."

jhauser-us:
Megan is right that there are several issues getting addressed together here, when they arguably should be separated.  I think I'm hearing agreement on two points:
1. A debugger may not change the selected hart during an abstract command.
2. If a hart changes to unavailable while executing an abstract command, then we want the DM not to crash but instead terminate the command with an error reported in `cmderr`.

Am I correct that these are agreed?  If so, I propose we proceed on them, and leave the possibility (or not) of terminating an active abstract command to a new issue, which one of us can open.  It's my belief the points above don't depend on the latter being resolved, so long as we accept that an active command must always be ended in some way before the debugger is free to change the selected hart.

debug_module.tex
```diff
@@ -229,13 +217,32 @@ \section{Abstract Commands} \label{abstractcommands}
...
 
 While an abstract command is executing (\Fbusy in \Rabstractcs is high), a
 debugger must not change \Fhartsel, and must not write 1 to \Fhaltreq,
 \Fresumereq, \Fackhavereset, \Fsetresethaltreq, or \Fclrresethaltreq.
 
+To abort an abstract command, the debugger first resets the hart (using
+\Fhartreset or \Fndmreset), and then the Debug Module (using \Fdmactive).
+
+If a hart becomes unavailable while executing an abstract command then the
+abstract command may be terminated, with \Fbusy going low, and \Fcmderr being
```
mwachs5: can we say "the debug module can terminate the abstract command, setting \Fbusy low,...". It's unclear who is doing the termination.

debug_module.tex
```diff
@@ -229,13 +217,32 @@ \section{Abstract Commands} \label{abstractcommands}
...
+
 Before starting an abstract command, a debugger must ensure that \Fhaltreq,
 \Fresumereq, and \Fackhavereset are all 0.
 
 While an abstract command is executing (\Fbusy in \Rabstractcs is high), a
 debugger must not change \Fhartsel, and must not write 1 to \Fhaltreq,
 \Fresumereq, \Fackhavereset, \Fsetresethaltreq, or \Fclrresethaltreq.
 
+To abort an abstract command, the debugger first resets the hart (using
+\Fhartreset or \Fndmreset), and then the Debug Module (using \Fdmactive).
```
mwachs5: We may also want to say this is not expected scenario and should only happen due to something wrong (not normal SW debugging scenarios). It may not actually resolve the situation, depending on what went wrong (e.g. locked up bus is not resolved by this procedure).

mwachs5: Maybe we should say "If an abstract command does not complete in the expected time and appears to be hung, the following procedure can be attempted to abort the command:..."

debug_module.tex
timsifive: Clarified.

debug_module.tex
```diff
@@ -229,13 +217,35 @@ \section{Abstract Commands} \label{abstractcommands}
...
 While an abstract command is executing (\Fbusy in \Rabstractcs is high), a
 debugger must not change \Fhartsel, and must not write 1 to \Fhaltreq,
 \Fresumereq, \Fackhavereset, \Fsetresethaltreq, or \Fclrresethaltreq.
 
+If an abstract command does not complete in the expected time and appears to be
+hung, the following procedure can be attempted to abort the command. First the
+debugger resets the hart (using \Fhartreset or \Fndmreset), and then it resets
+the Debug Module (using \Fdmactive).
+
+If a hart becomes unavailable while executing an abstract command then the the
```
mwachs5: "the the"

debug_module.tex
```diff
@@ -229,13 +217,35 @@ \section{Abstract Commands} \label{abstractcommands}
...
+Debug Module may terminate the abstract command, setting \Fbusy low, and
+\Fcmderr to 4 (halt/resume). Alternatively, the command could just appear to be
+hung.
```
mwachs5: hung (\Fbusy does not go low)

PR #326, opened 2018-08-02
DCSR is not writable while a hart is running.

timsifive:


timsifive:
I think of abstract commands as kind of orthogonal to the mode the hart is in. Eg. if a hart is in U mode, I would still expect abstract commands to be able to access M mode CSRs (assuming accessing CSRs while the hart is running is implemented).

timsifive:
I'd rather blacklist registers that we know might cause trouble, instead of whitelisting ones we think have value to allow access to.

mwachs5:
Sorry I’m missing something then, why are we special casing DCSR?

timsifive:
DCSR is special because it contains the step bit, and this is simpler than specifying what happens when the step bit is set while the hart is running. I'm just as happy to say debuggers mustn't do that.

I could make an argument that a debugger will want to set stopcount/stoptime while a hart is running, before it's ever halted, to have minimal impact on running software.

timsifive:
Right.

PR #329, opened 2018-08-02
Add dcsr.progbufjump.

### LINKED ISSUE #314: Are all program buffer implementations required to support jumps out of the program buffer?
Section 3.6, "Program Buffer", says
> The debugger can write whatever program it likes (including jumps out of the Program Buffer) ...

Later we have in a comment:
> The slightly inconsistent behavior with a Program Buffer of size 1 is to accommodate hardware designs that prefer to stuff instructions directly into the pipeline when halted, instead of having the Program Buffer exist in the address space somewhere.

If a DM implements the program buffer by injecting instructions directly into the pipeline of a halted hart, what is supposed to happen when the instruction being injected is a branch that jumps out of the program buffer?  The first quotation above suggests that such jumps must be supported, yet  it will be costly for a DM of this type to do so.

(Please be aware that a DM might prefer to implement program buffer execution by instruction injection even when the program buffer is larger than one word.)

timsifive:
This is needed to support instruction stuffing implementations, where
adding jumps to the Program Buffer doesn't make any sense.

Fixes #314.

### DIFF
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -15,8 +15,7 @@
           15: There is external debug support, but it does not conform to any
           available version of this spec.
         </field>
-        <field name="0" bits="27:18" access="R" reset="0" />
-        <field name="progbufjump" bits="17:16" access="R" reset="Preset">
+        <field name="progbufjump" bits="27:26" access="R" reset="Preset">
             0: All control transfer instructions are supported in the Program
             Buffer.
 
@@ -28,6 +27,7 @@
             3: No control transfer instructions are supported in the Program
             Buffer. Instead they act as illegal instructions.
         </field>
+        <field name="0" bits="25:16" access="R" reset="0" />
         <field name="ebreakm" bits="15" access="R/W" reset="0">
             When 1, {\tt ebreak} instructions in Machine Mode enter Debug Mode.
         </field>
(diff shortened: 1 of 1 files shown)


xml/core_registers.xml
```diff
@@ -15,7 +15,17 @@
           15: There is external debug support, but it does not conform to any
           available version of this spec.
         </field>
-        <field name="0" bits="27:16" access="R" reset="0" />
+        <field name="0" bits="27:18" access="R" reset="0" />
+        <field name="progbufjump" bits="17:16" access="R" reset="Preset">
+            0: All jumps/branches are supported in the Program Buffer.
```
mwachs5: Is there a RISC-V way of referring to this? To be really concrete about what a jump or branch instruction is?

timsifive:
You raise a good point about trying and seeing if it works. If we go that route we should mention explicitly which instructions are allowed to fail. There must be some instructions (eg. csrw) that the debugger can rely on.

I'd be just as happy with that, really.

xml/core_registers.xml
timsifive: Looks like the ISA spec calls them Control Transfer Instructions.

mwachs5:
I prefer the alternate proposal

timsifive:
Merged #331 instead.

PR #332, opened 2018-08-09
clrresethaltreq trumps setresethaltreq

### LINKED ISSUE #319: What happens if setresethaltreq and clrresethaltreq are set simultaneously?
If `dmcontrol` is written with bits `setresethaltreq` and `clrresethaltreq` both set (= 1), what happens?

timsifive:
Fixes #319.

Also explain why there are separate bits in the first place, and make
them both W1.


timsifive:
Thanks, Bryan. Somehow your review isn't enough to convince github that I can merge. :-(

PR #334, opened 2018-08-10
Side effects happen for abstract register accesses

timsifive:
Clarifies #318.

### DIFF
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -3,10 +3,14 @@
     <register name="Access Register">
         \begin{steps}{This command gives the debugger access to CPU registers and program buffer.
         It performs the following sequence of operations:}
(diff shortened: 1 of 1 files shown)


timsifive:
I don't think so. What conflict are you thinking of? Writing the trigger registers doesn't have any side effects.
Are we using different definitions of "side effect?" For me, a side effect to writing a register is some state changing besides the content of the register that's written.

allenjbaum:
A side effect is that it could trigger on some thing it wasn't triggering
on before, or stop triggering on something that it would have triggered on
before.
Basically, if a CSR bit goes to some bit of logic besides the read logic
for that CSR, it has a (podtential) side effect.
And now i don't remember, and can't find the original thread, but its moot,
probably.

PR #339, opened 2018-08-17
mprven may be tied high or low.

timsifive:


xml/core_registers.xml
```diff
@@ -87,11 +87,12 @@
...
-            When 1, \Fmprv in \Rmstatus takes effect during debug mode.
-            When 0, it is ignored during debug mode.
-            Implementing this bit is optional.
-            If not implemented it should be tied to 0.
+        <field name="mprven" bits="4" access="R/W" reset="Preset">
+            0: \Fmprv in \Rmstatus is ignored in Debug Mode.
+
+            1: \Fmprv in \Rmstatus takes effect in Debug Mode.
+
+            Implementing this bit is optional. It may be tied to either 0 or 1.
```
mwachs5: I'm fine with this, but is there a different  `access=R/W` that basically says this?

timsifive: Nothing that we're currently using. "R/W?" might mean readable and optionally writable. I just made that up. I'm not sure how intuitive it is.

allenjbaum:
Can’t you just define it as WARL with preset initial value?

-Allen

timsifive:
Yes. Thanks for pointing that out.

PR #341, opened 2018-08-24
On trigger chains, only the last action is taken.

timsifive:


timsifive:
There is no "no action" encoding.

PR #350, opened 2018-09-07
Allow global reset to reset the DM.

### LINKED ISSUE #346: System reset
The `dmactive` says: When `1` no other mechanism should exist that may result in resetting the Debug Module after power up. 

That may be obvious to others but do you suggest that also an external (hard) reset pin can't reset the system or are you only specifying that the `ndmreset` register can not reset the DM + DTM?

timsifive:
Fixes #346.

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -242,8 +242,9 @@
             1: The module functions normally.
 
             No other mechanism should exist that may result in resetting the
-            Debug Module after power up, including the platform's system reset
-            or Debug Transport reset signals.
+            Debug Module after power up, with the possible (but not
+            recommended) exception of a global reset signal that resets the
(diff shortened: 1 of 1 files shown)


mwachs5:
I don't really like this change. What's the point of saying it? Someone can always do it that way and say the implementation isn't quite compliant, as they are free to not implement any part of the spec.

timsifive:
Without this change, people have to implement some special debug reset logic which complicates their design. Resetting everything except for the debug logic is already present as ndmreset. I don't see a reason to require an additional mechanism.

timsifive:
Additionally, saying somebody has the option of implementing something and saying they're not compliant is not really an option at all. The whole point of a spec is that people can implement it and advertise that, so their platforms will work with debug tools.

Issue #353, opened 2018-09-07
What states are possible while a hart is in reset?


jhauser-us:
The DM spec should say explicitly what states may be reported for a hart while being held in reset or while "coming out of reset".  Currently, in section 3.2, Reset Control, we have
> The system may take an arbitrarily long time to come out of reset, as reported by `allunavail`, `anyunavail`, or other implementation specific indicators.

I read this to imply that, while a hart is in reset, it should be reported as *unavailable*, unless there is some other "implementation-specific indicator" of reset.  If that's the intention, the spec would benefit from saying so explicitly, i.e., harts being reset are reported as *unavailable*, unless reset is indicated by some other, implementation-specific means.

Furthermore, I believe the current spec is not clear about what other states, if any, may be indicated for a hart while in reset.  Can or should a DM report *halted* or *running* for any harts being reset?  For example, if a halted hart gets reset, should (or may) the *halted* state remain asserted during reset?

Please note the interaction of this issue with #351.

Lastly, is there any possibility of dropping the option of an "implementation-specific indicator" for reset?  Even if a DM has an extra, nonstandard indicator, I would like for every DM to consistently report the state of harts in reset, whether that's as *unavailable* or some other agreed state.

jhauser-us:
The discussion for this issue seems to be occurring elsewhere, which I think is unfortunate.  For the benefit of anyone who might look here, some desire has been expressed to allow harts in reset to report a state other than *unavailable*.  Under #355, Tim Newsome wrote:
> The best solution is to say that while held in reset, harts may report being unavailable (no abstract commands possible) or running (abstract commands are possible). Running then means something like "doing what the hart would do without the debugger interfering."

I haven't seen anyone arguing yet for allowing harts in reset to report *halted*.  Does anyone believe that one should be possible also?

(From issue #351, it appears to be agreed that *unavailable*, *running*, and *halted* are always mutually exclusive.)

mwachs5:
[This comment isn't really suggesting anything, just running through the thought experiment]:

> I haven't seen anyone arguing yet for allowing harts in reset to report halted. Does anyone believe that one should be possible also?

If a hart was halted, then (unknown to the debugger) got reset and didn't halt out of reset, then it would be confusing to report it as Halted because the debug module wouldn't actually have control of it. But, the debugger could observe the 'hasreset' and deduce what happened. Still, not optimal situation.

[...]

PR #354, opened 2018-09-07
Harts can be in exactly 1 of 4 states.

### LINKED ISSUE #351: Suggest stating that each hart index is associated with exactly one of four states
For each implemented hart index (in the range determined by HARTSELLEN), four states can be reported in `dmstatus`:  "nonexistent", "unavailable", "running", or "halted".  I believe the following is implied or can be inferred from the DM spec:

- "Nonexistent" is exclusive of the other three states.  (Seems obvious.)

- "Unavailable" is exclusive of "halted".  (The DM spec says, "Unavailable/nonexistent harts are not considered to be halted.")

- "Unavailable" is exclusive of "running".  (Implied, I think.)

- "Running" and "halted" are mutually exclusive.  (Seems obvious, and is implied in numerous ways by the DM spec.)

- If a hart exists but is not running or halted, it must be "unavailable".  (Implied, I think.)

In sum, unless I'm wrong, exactly one of these four states is required at any time.  I believe it would be helpful to state this more clearly, possibly like so:
> For each implemented hart index (as determined by HARTSELLEN), exactly one of the following four states is associated with the index at any given time:  *nonexistent*, *unavailable*, *halted*, or *running*.

The best place to say this might be in section 3.4, "Run Control".

This clarification is relevant to other issues I will be opening to advance the discussion about reset and abstract commands.

If my interpretation is wrong, then I assert that the DM spec needs an explanation for how and why I'm wrong.

timsifive:
Fixes #351.


debug_module.tex
```diff
@@ -126,13 +126,43 @@ \section{Selecting Harts} \label{selectingharts}
...
+if there are other harts with higher indexes than this one. E.g.\ in a
+multi-hart system some might temporarily be powered down, or a system might
+support hot-swapping harts. Systems with very large number of harts may
+permanently disable some during manufacturing, leaving holes in the otherwise
+continuous hart index space. In order to let the debugger discover all harts,
+they must show up as unavailable even if there is no chance of them ever
+becoming available.
+
+Harts are running when they're executing normally, as if there was no debugger
+attached. This includes being in a low power mode or waiting for an interupt,
```
mwachs5: interrupt

debug_module.tex
```diff
@@ -126,13 +126,43 @@ \section{Selecting Harts} \label{selectingharts}
...
+Harts are unavailable if they might exist/become available at a later time, or
+if there are other harts with higher indexes than this one. E.g.\ in a
+multi-hart system some might temporarily be powered down, or a system might
+support hot-swapping harts. Systems with very large number of harts may
+permanently disable some during manufacturing, leaving holes in the otherwise
+continuous hart index space. In order to let the debugger discover all harts,
+they must show up as unavailable even if there is no chance of them ever
+becoming available.
+
+Harts are running when they're executing normally, as if there was no debugger
```
mwachs5: they're -> they are

mwachs5: there was -> there were (?)

timsifive:
What state is it in when a hart is in reset is #353. It does seem natural to address it in this PR as well.

debug_module.tex
timsifive: Thanks.

debug_module.tex

timsifive: Reworded.

xml/dm_registers.xml
```diff
@@ -357,7 +357,7 @@
             (e.g.\ while executing the Program Buffer).
 
             4 (halt/resume): The abstract command couldn't execute because the
-            hart wasn't in the required state (running/halted).
+            hart wasn't in the required state (running/halted), or unavailable.
```
mwachs5: I think we should just make this its own error type, or do what you did before and rename this error type. 

What if a hart becomes unavailable during the course of this? Didn't we have another PR that adds a new error type for that?

jhauser-us: I don't see a need for a separate error code for this case.  Clearly the hart is in the wrong state when the abstract command is attempted.  I don't have a strong feeling about the name.

debug_module.tex
```diff
@@ -139,17 +139,17 @@ \section{Hart States}
...
 
 Harts are unavailable if they might exist/become available at a later time, or
-if there are other harts with higher indexes than this one. The most common
-example of this state is when a hart is held in reset. Other examples are a
-hart that is temporarily powered down, or one that might be hot-swapped in
-later.
+if there are other harts with higher indexes than this one. Harts may be
+unavailable for a variety of reasons including being reset, temporarily powered
+down, and not being plugged into the system.  No abstract commands can be
+executed while an unavailable hart is selected.
```
jhauser-us: While a statement is needed that an unavailable hart can't execute abstract commands, I suggest moving this to the section on abstract commands.  In my opinion, it's better not to say too much about abstract commands before they've been properly introduced.

timsifive: Good points. I've addressed them.

xml/dm_registers.xml
timsifive: @mwachs5 How important is changing the error code to you? I think that's the only thing that's preventing this PR from being approved.

debug_module.tex
```diff
@@ -172,6 +209,8 @@ \section{Abstract Commands} \label{abstractcommands}
 are optional. Depending on the implementation, the debugger may
 be able to perform
 some abstract commands even when the selected hart is not halted.
+No abstract commands can be
+executed while an unavailable hart is selected.
```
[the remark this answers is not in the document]
timsifive: This PR doesn't address that. Merged earlier today:
> If a hart becomes unavailable while executing an abstract command then the Debug Module may
> terminate the abstract command, setting busy low, and cmderr to 4 (halt/resume). Alternatively,
> the command could just appear to be hung (busy does not go low).

This text here just says that abstract commands won't be executed, but not what happens if it's attempted. I feel the behavior should be identical to what happens when a hart becomes unavailable. Should I clarify that here?

Note that there is always a race condition, where a hart might become unavailable just before the DM is starting to execute the abstract command.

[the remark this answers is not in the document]
jhauser-us: I believe this statement is intended to be about the case when the selected hart is reported as unavailable at the time the `command` register is written.  This isn't really different than the hart being in the wrong running or halted state when `command` is written.  One hardware check tests for both cases, with `cmderr` getting set to 4 if the state is wrong.  Perhaps the wording needs to be adjusted to be clearer.  (Or Tim can correct me if I've got it wrong.)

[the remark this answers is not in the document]
mwachs5: Right, I was just objecting to the passive voice, not the actual intended meaning. "no commands can be executed" means (I believe) "commands may result in ABSTRACTCS.error=4" and/or "debugger should not write commands...".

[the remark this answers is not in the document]
timsifive: I've attempted to make this more obvious.

PR #360, opened 2018-09-12
Abstract commands might work on a hung hart.

timsifive:
From a discussion on today's call.

### DIFF
--- a/implementations.tex
+++ b/implementations.tex
@@ -14,6 +14,10 @@ using the Access Register abstract command.
 Memory is accessed using the Abstract Access Memory command or through System
 Bus Access.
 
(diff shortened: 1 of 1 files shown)


implementations.tex
```diff
@@ -14,6 +14,10 @@ \section{Abstract Command Based}
 Memory is accessed using the Abstract Access Memory command or through System
 Bus Access.
 
+Since this implementation doesn't rely on the hart executing any instructions,
+it could allow a debugger to collect information from the hart even if it is
+otherwise completely hung.
```
mwachs5: I don't really know what "hung" means here. "Unable to execute instructions" makes more sense, though it's a bit redundant with the above.

PR #363, opened 2018-09-18
Add ASID and context compare for triggers

timsifive:
There was some e-mail discussion, subject:"ContextID trigger/trace"

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -95,11 +95,23 @@
         <field name="data" bits="XLEN-1:0" access="R/W" reset="Preset" />
     </register>
 
-    <register name="Trigger Data 3" short="tdata3" address="0x7a3">
-        Trigger-specific data.
+    <register name="Trigger Extra" short="textra" address="0x7a3">
+        \begin{commentary}
+            ASIDLEN is limited in size when XLEN is less than DXLEN to preserve
+            the maximum number of bits for future use.
+        \end{commentary}
+        <field name="0" bits="XLEN-1:2+ASIDLEN" access="R" reset="0" />
+        <field name="tasid" bits="1+ASIDLEN:1" access="R/W" reset="0">
+            If XLEN is 32 then ASIDLEN is 9. Otherwise ASIDLEN is 16.
+        </field>
+        <field name="tasiden" bits="0" access="R/W" reset="0">
+            0: \Ftasid is ignored.
 
-        If XLEN is less than DXLEN, writes to this register are sign-extended.
-        <field name="data" bits="XLEN-1:0" access="R/W" reset="Preset" />
+            1: This trigger will only match if \Fasid in \Rsatp equals the
+            lower bits of \Ftasid. The number of bits to compare depends on the
+            current value of SXLEN. If it is 32, then 9 bits are compared,
+            otherwise 16 bits are compared.
+        </field>
     </register>
 
     <register name="Trigger Info" short="tinfo" address="0x7a4">
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -145,6 +145,34 @@
         <field name="0" bits="2:0" access="R" reset="0" />
     </register>
 
+    <register name="Machine Context" short="mcontext" address="0x7a8">
+        This register is only writable in M mode and Debug Mode.
+
+        <field name="mcontext" bits="XLEN-1:0" access="R/W" reset="0">
+            Machine mode software can write a context number to this register,
+            which can be used to set triggers that only fire in that specific
+            context.
+
+            An implementation may tie any number of high bits in this field to
+            0. It's recommended to implement no more than 6 bits on RV32, and
+            13 on RV64.
+        </field>
+    </register>
+
+    <register name="Supervisor Context" short="scontext" address="0x7aa">
+        This register is only writable in S mode, M mode and Debug Mode.
+
+        <field name="data" bits="XLEN-1:0" access="R/W" reset="0">
+            Supervisor mode software can write a context number to this
+            register, which can be used to set triggers that only fire in that
+            specific context.
+
+            An implementation may tie any number of high bits in this field to
+            0. It's recommended to implement no more than 16 bits on RV32, and
+            34 on RV64.
+        </field>
+    </register>
+
     <register name="Match Control" short="mcontrol" address="0x7a1">
         This register is accessible as \Rtdataone when \Ftype is 2.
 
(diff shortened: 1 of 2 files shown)


xml/hwbp_registers.xml
```diff
@@ -95,11 +95,23 @@
...
-    <register name="Trigger Data 3" short="tdata3" address="0x7a3">
-        Trigger-specific data.
+    <register name="Trigger Extra" short="textra" address="0x7a3">
+        \begin{commentary}
+            ASIDLEN is limited in size when XLEN is less than DXLEN to preserve
+            the maximum number of bits for future use.
+        \end{commentary}
+        <field name="0" bits="XLEN-1:2+ASIDLEN" access="R" reset="0" />
+        <field name="tasid" bits="1+ASIDLEN:1" access="R/W" reset="0">
+            If XLEN is 32 then ASIDLEN is 9. Otherwise ASIDLEN is 16.
```
mwachs5: Shouldn't it be as the priv spec defines it:

```
The number of ASID bits is also implementation-defined and may be zero.  The
number of implemented ASID bits, termed {\mbox {\em ASIDLEN}}, may be
determined by writing one to every bit position in the ASID field, then
reading back the value in {\tt satp} to see which bit positions in the ASID
field hold a one.  The least-significant bits of ASID are implemented first:
that is, if ASIDLEN~$>$~0, ASID[ASIDLEN-1:0] is writable.  The maximal value
of ASIDLEN, termed ASIDMAX, is 9 for Sv32 or 16 for Sv39 and Sv48.
```

mwachs5: Then we don't need to redefine it at all here.

xml/hwbp_registers.xml
```diff
@@ -95,11 +95,23 @@
...
+        </field>
+        <field name="tasiden" bits="0" access="R/W" reset="0">
+            0: \Ftasid is ignored.
 
-        If XLEN is less than DXLEN, writes to this register are sign-extended.
-        <field name="data" bits="XLEN-1:0" access="R/W" reset="Preset" />
+            1: This trigger will only match if \Fasid in \Rsatp equals the
+            lower bits of \Ftasid. The number of bits to compare depends on the
+            current value of SXLEN. If it is 32, then 9 bits are compared,
```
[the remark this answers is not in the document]
timsifive: Because ASID is in the satp register, which the privspec says is SXLEN bits wide. It then makes ASID dependent on SXLEN exactly like I did here.

xml/hwbp_registers.xml
```diff
@@ -95,11 +95,23 @@
         <field name="data" bits="XLEN-1:0" access="R/W" reset="Preset" />
     </register>
 
-    <register name="Trigger Data 3" short="tdata3" address="0x7a3">
-        Trigger-specific data.
+    <register name="Trigger Extra" short="textra" address="0x7a3">
```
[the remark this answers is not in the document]
timsifive: It's intended to go with any kind of trigger. Specifically with match triggers as Bruce requested, and count triggers as Chuanhua requested. I'll try to clarify that.

xml/hwbp_registers.xml
timsifive: I missed that language. Let me take that into account.

xml/hwbp_registers.xml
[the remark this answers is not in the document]
mwachs5: Yeah, I didn't realize that `\Fasid` was already a thing, I thought we had to have it added to the spec.

Then I guess the question (as discussed in the email thread) is whether `asid` is acceptable standin for a "process" or whether we need a specific CSR for this. My gut is we should go with the `asid` already defined in the spec, but I'm not an OS programmer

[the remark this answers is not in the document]
timsifive: I imagine operating systems would use `asid` to prevent processes from walking on each other's memory, so in UNIX systems every process would have its own `asid`. Threads within processes might share `asid` though, and I'm sure there are other systems where a context ID would be better. I'd rather not define a context ID in the debug spec, though. If we do, it should be in a separate PR after this one merges.

xml/hwbp_registers.xml
```diff
@@ -95,11 +95,25 @@
...
+        not implemented. The checks specified in this register apply regardless
+        of \Ftype in \Rtdataone.
+
+        \begin{commentary}
+            TASIDLEN is limited in size when XLEN is less than DXLEN to
+            preserve the maximum number of bits for future use.
+        \end{commentary}
+        <field name="0" bits="XLEN-1:1+TASIDLEN" access="R" reset="0" />
+        <field name="tasid" bits="TASIDLEN:1" access="R/W" reset="0">
+            If XLEN is 32 then TASIDLEN is 9. Otherwise TASIDLEN is 16.
```
mwachs5: How does this statement relate to LOG(ASIDMAX) and the above statement that "any of the bits can be hardwired to 0". I *think* you mean that even if TASIDLEN is 16, this entire field may not be writable, if LOG(ASIDMAX) < TASIDLEN. Confirming that's what you mean, and thinking if this needs re-wording if that is the case. You could just say `TASIDLEN` is CEIL(LOG2(ASIDMAX)) if that's what you really mean.

mwachs5: I guess let's say you had 7 possible ASID in a system (ASIDMAX = 3). Could I have only 1 writable bit in this field?

Again I think the answer is "yes" just sort of confusing how the different values are defined. The way the priv spec defines it is very clear.

mwachs5: I was confusing myself, ASIDMAX is already the width. Will update my comments above.

mwachs5:
How does this relate to the hypervisor version of ASID (maybe this question applies to all triggers):
https://github.com/riscv/riscv-isa-manual/blob/a94d4b0534d1f75f4ab4325b480d65128c554b8c/src/hypervisor.tex#L626

mwachs5:
I've been chatting with some of the SW folks and it seems like ASID isn't really the right thing to use here. There is Bruce's suggestion for the mailing list which feels less ad hoc than this:

> 
[BA] I think it is much better to have separate enable/disable control bits per what I would call global comparison fields.  ContextID is one; a separate comparison on ASID is another.  And future cores will certainly implement virtualization and each virtual machine will have an ID.  ARM labels them VMID, MIPS labels them GuestID.  They will need to be included in trigger patterns and in trace sync messages whenever they change.

There is also suggestion that why don't we just do a trap on any context switch/write to SATP (controlled by dcsr whether it goes to debugger or not) then debugger can do whatever it wants in terms of setting up triggers, without adding extra HW

xml/hwbp_registers.xml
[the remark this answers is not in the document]
timsifive: Good point. I'll fix that.

xml/hwbp_registers.xml
timsifive: What I'm trying to resolve is a 64-bit core, with the current XLEN being 32. Such a core could have ASIDLEN=16, but even if that is the case, when the current XLEN is 32 I still want TASIDLEN to be 9. This leaves us the most number of bits available for future use in this register, and also makes it easier to use. The running code doesn't have to know the actual max XLEN. Just based on the current XLEN it knows TASIDLEN.

Clearly I completely failed to communicate that. Does this comment make more sense?

Just above I mention explicitly that any bit in the register may be tied to 0 by an implementation, which should cover your case of implementing fewer bits when ASIDLEN<ASIDMAX.

timsifive:
OK. What do you propose? I'm not really invested in a particular solution. I'm just making sure that progress is happening.

I don't like trapping on any write to satp. This would be a huge performance hit if context switches are at all frequent. (In my Linux VM there were 770 context switches per second, just randomly looking.)

xml/hwbp_registers.xml
```diff
@@ -145,6 +145,34 @@
         <field name="0" bits="2:0" access="R" reset="0" />
     </register>
 
+    <register name="Machine Context" short="mcontext" address="0x7a8">
```
mwachs5: is this address appropriate for a register accessible in M and Debug mode?

xml/hwbp_registers.xml
```diff
@@ -145,6 +145,34 @@
...
+        This register is only writable in M mode and Debug Mode.
+
+        <field name="mcontext" bits="XLEN-1:0" access="R/W" reset="0">
+            Machine mode software can write a context number to this register,
+            which can be used to set triggers that only fire in that specific
+            context.
+
+            An implementation may tie any number of high bits in this field to
```
mwachs5: high bits? Upper bits?

xml/hwbp_registers.xml
```diff
@@ -145,6 +145,34 @@
...
+            0. It's recommended to implement no more than 6 bits on RV32, and
```
mwachs5: where does this recommendation come from?

xml/hwbp_registers.xml
```diff
@@ -145,6 +145,34 @@
...
+            13 on RV64.
+        </field>
+    </register>
+
+    <register name="Supervisor Context" short="scontext" address="0x7aa">
+        This register is only writable in S mode, M mode and Debug Mode.
+
+        <field name="data" bits="XLEN-1:0" access="R/W" reset="0">
+            Supervisor mode software can write a context number to this
+            register, which can be used to set triggers that only fire in that
```
mwachs5: maybe "match" instead of "set"?

xml/hwbp_registers.xml
```diff
@@ -145,6 +145,34 @@
...
+            specific context.
+
+            An implementation may tie any number of high bits in this field to
+            0. It's recommended to implement no more than 16 bits on RV32, and
```
mwachs5: same question here, why these recommendations? Why do we need to recommend anything at all?

xml/hwbp_registers.xml
```diff
@@ -493,4 +521,58 @@
...
+        <!-- This is space left open for H mode support. -->
+
+        <field name="svalue" bits="17:2" access="R/W" reset="0">
+            An implementation may tie any number of high bits in this field to
+            0.
+        </field>
+        <field name="sselect" bits="1:0" access="WARL" reset="0">
+            0: Ignore \Fmvalue.
+
+            1 (optional): This trigger will only match if the low bits of
```
mwachs5: rather than having all these things "optional" can we just say one "optional" thing at the top for this entire register?

xml/hwbp_registers.xml
timsifive: Yes. 0x7a0 -- 0x7af are reserved for M-mode "standard read/write debug CSRs".

xml/hwbp_registers.xml
timsifive: Upper bits is better. Thanks.

xml/hwbp_registers.xml
timsifive: The width of the *value fields in textra.

xml/hwbp_registers.xml
timsifive: match is not what I mean at all. Configure is. Is that better?

xml/hwbp_registers.xml
timsifive: We could remove the recommendation, and assume that implementers will notice that they're only useful in the context of textra, and find the limits themselves. I was trying to be helpful.

xml/hwbp_registers.xml
timsifive: I've made that change, but I think the other way is more clear.

timsifive:
> In general, this seems OK, but seems like a pretty big change to shove in at the last minute without any prototyping, SW support, etc.

I agree, but I lean towards the spec being better with it than without it.

xml/hwbp_registers.xml
```diff
@@ -493,4 +521,60 @@
...
+            0: Ignore \Fmvalue.
+
+            1: This trigger will only match if the low bits of
+            \Rmcontext equal \Fmvalue.
+        </field>
+
+        <field name="0" bits="24:18" access="R" reset="0" />
+        <!-- This is space left open for H mode support. -->
+
+        <field name="svalue" bits="17:2" access="R/W" reset="0">
```
mwachs5: why so many more bits for svalue vs mvalue?

timsifive: My thinking was that an OS would run in S-mode, and write a PID to svalue (context id). OSs can have many processes.

Similarly I think of the stuff running in M-mode being more akin to a hypervisor, which has VMs or something like them. You don't usually have that many VMs on a system.

I may be totally off-base here.

timsifive:
Agreed. This is not ideal, but it's better to have it than not to have it.

PR #364, opened 2018-09-18
Distinguish draft and release builds.

timsifive:
Rename the output PDFs so it's less likely to cause confusion.
The release build is not labelled as DRAFT, does not include the preface
warning, and does not include the change log.


timsifive:
Yes. The only way to build the release is to `make release`. And then you'll end up with a file called riscv-debug-release.pdf.

mwachs5:
I thought we deliberately deleted serial.tex

timsifive:
serial.tex is only referred to by future.tex. The file itself is gone, but it's generated from xml/serial.xml. We could get rid of future.tex, but that seems like work for another PR. I don't think it hurts to keep it around.

PR #366, opened 2018-09-20
Little language and formatting cleanups.

timsifive:


debug_module.tex
```diff
@@ -442,7 +437,7 @@ \section{System Bus Access} \label{systembusaccess}
 access to devices that a hart does not have access to.
 \end{commentary}
 
-\section{Quick Access}
+\section{Minimally Intrusive Debugging}
```
mwachs5: Why change this title? Is there nowehre else we refer to Quick Access?

xml/dm_registers.xml
```diff
@@ -435,8 +434,7 @@
       hold identifier information which is not
       further specified in this document.
 
-      The Device Tree itself is described in the RISC-V Privileged
-      Specification.
+      The Device Tree itself is described in the Privileged Spec.
```
mwachs5: Is it? It may have been removed.

timsifive: It's mentioned, but not defined anywhere. Instead it talks about a more handwavy "configuration string" and specifically mentions that platforms can provide a "configuration string pointer." I'll rename this register to reflect that.

debug_module.tex
mwachs5: makes sense

xml/dm_registers.xml
mwachs5: I thought we changed this to Device tree FROM config string because they were going to get rid of Config string... weird.

timsifive: From the preface to Version 1.9.1:
> Change configuration string proposal to be use a search process that supports various formats including Device Tree String and flattened Device Tree.

I don't know if I went and looked at the language when I changed the debug spec to refer to Device Tree.

mwachs5:
If we are renaming Device Tree -> Configuration String again, weren't there some registers that we renamed to "Device Tree Valid" or something like that?

mwachs5:
I am referring to: https://github.com/riscv/riscv-debug-spec/blob/472587910ce4796558fbb03b4c3a7ea29505010f/xml/dm_registers.xml#L89

mwachs5:
Can we back out the device tree -> cfg string from this PR ,and make it a seperate one? I think the rest of this is benign and that's the only thing blocking my approval on it

mwachs5:
Oh, I see that you already changed those registers name

PR #368, opened 2018-09-25
Stepping over wfi does not enter wait state.

timsifive:


mwachs5:
Does the spec clearly state somewhere that debug halt requests force wfi to compete?

timsifive:
I agree it's a bit convoluted. I'm accounting for the possibility that wfi causes an exception. From the privspec:
> This instruction may raise an illegal instruction exception when TW=1 in mstatus, as described in Section 3.1.10.

In that case the exception must happen, and it should not be a nop.

The spec doesn't say that wfi is interrupted by halt. I'll add that.

mwachs5:
Your reading of the spec is closer than mine, sounds good!

core_debug.tex
```diff
@@ -66,6 +66,12 @@ \section{Load-Reserved/Store-Conditional Instructions}
     automate this.
 \end{commentary}
 
+\section{Wait for Interrupt Instruction}
+
+If halt is requested while {\tt wfi} is executing, then the hart must leave the
+stalled state, completing this instruction's execution, and then enter Debug
```
mwachs5: is there a reason to use "stalled' here and "wait" in your other change? Aren't they the same state? Can we use the same term?

timsifive: I'll make it consistently use stalled (which is the term used in the priv spec).

PR #369, opened 2018-09-26
Reset dmi.op to 0 instead of 2.

timsifive:
Now it's consistent with dtmcs.dmistat.

This reverts 7f0f09a0. It is no longer true that a debugger can assume
dmi.data contains a valid value unless a read was explicitly executed,
so there's no sense in accommodating that case.


mwachs5:
Was 7f0f09a part of a PR? I was looking to see if there was any discussion around why that was added in the first place, before we go reverting it

timsifive:
I couldn't find any evidence of a PR, and nothing in my e-mail either.

PR #370, opened 2018-09-26
Add dcsr.cause for being halted out of reset.

timsifive:


xml/core_registers.xml
```diff
@@ -83,9 +83,11 @@
 
             2: The Trigger Module caused a breakpoint exception. (priority 4, highest)
 
-            3: The debugger requested entry to Debug Mode. (priority 2)
+            3: The debugger requested entry to Debug Mode. (priority 1)
 
-            4: The hart single stepped because \Fstep was set. (priority 1, lowest)
+            4: The hart single stepped because \Fstep was set. (priority 0, lowest)
+
+            5: The hart halted directly out of reset. (priority 2)
```
mwachs5: is it OK to report `3` in this case? otherwise this is not a backwards compatible change, and I can imagine this is hard to implement in all cases.

timsifive: That's a good point. 3 should be acceptable.

xml/core_registers.xml
```diff
@@ -83,9 +83,12 @@
 
             2: The Trigger Module caused a breakpoint exception. (priority 4, highest)
 
-            3: The debugger requested entry to Debug Mode. (priority 2)
+            3: The debugger requested entry to Debug Mode. (priority 1)
 
-            4: The hart single stepped because \Fstep was set. (priority 1, lowest)
+            4: The hart single stepped because \Fstep was set. (priority 0, lowest)
+
+            5: The hart halted directly out of reset. It is also acceptable to
```
mwachs5: is this just "out of reset" or "due to reset halt request" (vs due to haltreq was set when hart was reset)?

timsifive: I intended "out of reset." I don't know why somebody would want to differentiate between haltreq and resethaltreq.

mwachs5: As I said on the email thread (sorry for carrying conversation in two places), because the SW use cases are different. `haltreq` implies the hart was already under control of the debugger, while `resethaltreq` catches a hart which wasn't under the debugger's control and brings it there, similar to a breakpoint. 

Since this cause is already redundant with cause = `haltreq` and `hasreset` being set, I don't think it really matters, but we may want to allow either behavior. Maybe that's already implied by "it is also acceptable..." clause.

PR #371, opened 2018-09-26
Clean up language, formatting, consistency.

timsifive:


core_debug.tex
```diff
@@ -97,7 +98,7 @@ \section{Single Step}
 \section{Reset}
 
 If the halt signal (driven by the hart's halt request bit in the Debug Module)
-is asserted when a hart comes out of reset, the hart must
+or \Fresethaltreq are asserted when a hart comes out of reset, the hart must
```
mwachs5: I don't think its \Fresethaltreq. It's the conceptual {resethaltreq} bit.

core_debug.tex
```diff
@@ -133,7 +134,4 @@ \section{Core Debug Registers} \label{debreg}
 
 \section{Virtual Debug Registers} \label{virtreg}
 
-Virtual debug registers are a requirement on the debugger SW/interface,
```
mwachs5: weren't people confused about this? I think we added this sentence to avoid confusion.

core_debug.tex
timsifive: I've made \Fresethaltreq point to a description that says that.

core_debug.tex
timsifive: Yes. I hope that this is covered by the beginning of sw_registers.xml, which I also rewrote in this change:
    A virtual register is one that doesn't exist directly in the hardware, but
    that the debugger exposes as if it does. Debug software should implement
    them, but hardware can skip this section. Virtual registers exist to give
    users access to functionality that's not part of standard debuggers without
    requiring them to carefully modify debug registers while the debugger is
    also accessing those same registers.

It's bit tricky to see that by reading the source, but this ends up exactly in the space of the line that I deleted there.

core_debug.tex
mwachs5: but that's only readable if someone is looking at a PDF, clickable version of the spec? That seems tricky to rely on.

core_debug.tex
mwachs5: Ah, I see it was added back below.

core_debug.tex
mwachs5: Ah, I see, it's ONLY that it's not actually a bit in any register. Ok.

PR #372, opened 2018-09-26
Clarify W1.

timsifive:
The definition used to be vague. I think this is the one that makes the
most sense, and shouldn't cause any problems. (Possibly we don't want
this for ackhavereset. I sent e-mail about that.)

The bits that are defined as W1 are:
DM: ackhavreset, setresethaltreq, clrresethaltreq
JTAG: dmihardreset, dmireset

### DIFF
--- a/introduction.tex
+++ b/introduction.tex
@@ -84,7 +84,8 @@ indexed on page \pageref{index}.
         \hline
         W & Write-only. When read this field returns 0. \\
         \hline
-        W1 & Write-only. Only writing 1 has an effect. \\
+        W1 & Write-only. Every time that 1 is written an action is taken. When
+        read this field returns 0. \\
         \hline
         WARL & Write any, read legal. A debugger may write any value. If a
         value is unsupported, the implementation converts the value to one that
--- a/introduction.tex
+++ b/introduction.tex
@@ -84,8 +84,7 @@ indexed on page \pageref{index}.
         \hline
         W & Write-only. When read this field returns 0. \\
         \hline
-        W1 & Write-only. Every time that 1 is written an action is taken. When
-        read this field returns 0. \\
+        W1 & Write-only. Only writing 1 has an effect. \\
         \hline
         WARL & Write any, read legal. A debugger may write any value. If a
         value is unsupported, the implementation converts the value to one that
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -198,7 +198,7 @@
             select. This hart is always part of the currently selected harts.
         </field>
         <field name="0" bits="5:4" access="R" reset="0" />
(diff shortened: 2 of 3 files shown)


timsifive:
It may make sense to change dmihardreset and dmireset to use the same semantics as e.g. hartreset: a debugger must write 1, then 0, in order to effect a reset.

That allows for more implementations, and doesn't hurt anybody who has already implemented reset-on-write-of-1.

introduction.tex
```diff
@@ -84,7 +84,8 @@ \subsection{Register Definition Format}
         \hline
         W & Write-only. When read this field returns 0. \\
         \hline
-        W1 & Write-only. Only writing 1 has an effect. \\
+        W1 & Write-only. Every time that 1 is written an action is taken. When
```
mwachs5: I will need to go back through and see if this is correct. Is it really "every time"? I don't think that's true.

xml/dm_registers.xml
```diff
@@ -198,7 +198,7 @@
             select. This hart is always part of the currently selected harts.
         </field>
         <field name="0" bits="5:4" access="R" reset="0" />
-        <field name="setresethaltreq" bits="3" access="W1" reset="0">
+        <field name="setresethaltreq" bits="3" access="W1" reset="-">
```
mwachs5: for example, if it's already set here, you don't actually do anything when you W1 again.

introduction.tex
mwachs5: Do you mean "An action may be taken only when a 1 is written. Writing 0 has no effect."

timsifive: That doesn't feel strong enough.

I want every write of 1, regardless of the previously written value, to take the action. Eg. every time haltreq is written, we request that the hart halts. (This is ignored if the hart is already halted.) Likewise, every time setresethaltreq is written high, it sets the conceptual resethaltreq bit high. If it's already high, that's a no-op, but the attempt is made.

I believe this was a possible interpretation of the previous language, and I want to be clear that it's the only interpretation.

Does that make sense?

timsifive: haltreq is actually not a W1 bit. Pretend I wrote an example with dmireset.

xml/dm_registers.xml
mwachs5: It's confusing that "set reset halt req" is a pretty clear "action" in your terminology, yet actually writing "set reset halt req" doesn't actually do anything...?

I'm not saying what youre writing is wrong or doesn't make sense, I just don't see that it's actually more clear than what's currently there.

introduction.tex
mwachs5: I guess I dont' see that as more clear than the current "has an effect".

timsifive: OK. I've reverted this definition.

PR #379, opened 2018-10-01
Explicitly state what's required for compliance.

timsifive:


debug_module.tex
```diff
@@ -18,12 +18,10 @@ \chapter{Debug Module (DM)} \label{dm}
...
-In order to implement memory access, a target must implement at least one of
-Program Buffer, System Bus Access, or Abstract Access Memory command
-mechanisms.
-
-In order to implement access to all registers, targets should implement either
-the Program Buffer or abstract access to all registers.
+In order to be compliant with this specification an implementation must
+implement all the required features listed above, at least one of Program Buffer,
+System Bus Access, or Abstract Access Memory command mechanisms, and either the
+Program Buffer or abstract access to all registers.
```
mwachs5: "all registers" means "all CSRs and GPRs"? What about FPRs? What about vector registers? I think this is a bit vague. And can't an implementation be "fully compliant" and generally useful for debugging even if e.g. you can't access the FPRs?

timsifive: Let's tackle the second question first:
> can't an implementation be "fully compliant" and generally useful for debugging even if e.g. you can't access the FPRs?

It's better than not having any debugging, but the same could be said for single step, halt out of reset, and a host of other features which are not hard to implement, and provide a lot of value.

A debug solution that can't access all the registers is incomplete at best. If you have a vector unit, your debug implementation must provide access to its registers. Use some of the non-standard extension space in abstract commands if you have to.

In what kind of system would you choose not to implement debug access to all registers?

mwachs5: I'm just trying to make it clear, I think "registers" is a little fuzzy of a term.

timsifive: I've pushed a change that hopefully is more clear.

asb:
I'm sympathetic to requiring that all registers are accessible, but I can also see an argument for keeping requirements as minimal as possible: either having a program buffer or exposing GPRs + dpc via the abstract commands. I'd then add something like "It is strongly recommended that implementations provides access to all registers that are visible to software running on the hart".

It seems plausible that you might provide very minimal support for the RISC-V debug spec to help support very early system bring-up, and rely on gdbstub or similar beyond that.

Otherwise this looks great to me.

timsifive:
Thanks, @asb. Part of my reasoning to require all register access is that the more registers you have, the less overhead adding a Program Buffer is. If you feel strongly, I could be convinced to not require all registers, but such implementations must be advertised as conforming to the "Minimal RISC-V Debug Specification".

debug_module.tex
```diff
@@ -18,12 +18,15 @@ \chapter{Debug Module (DM)} \label{dm}
...
-In order to implement access to all registers, targets should implement either
-the Program Buffer or abstract access to all registers.
+\begin{steps}{In order to be compliant with this specification an
+    implementation must:}
+\item Implement all the required features listed above.
+\item Implement at least one of Program Buffer, System Bus Access, or Abstract
+    Access Memory command mechanisms.
+\item Implement either the Program Buffer, or abstract access to all registers
+    that are visible to software running on the hart including all the
+    registers that are present on the hart and listed in Table~\ref{tab:regno}.
```
mwachs5: I'm approving this, though I also would approve @asb's suggestion of "Minimal compliance". Though that's a whole nother level of something to define, and I think adding the program buffer is always the right choice, so...

timsifive:
I merged this, but I'll create a separate PR for "minimal compliance," in case that proves to be popular.

PR #381, opened 2018-10-02
Minimal implementations can't access all registers

timsifive:


debug_module.tex
```diff
@@ -23,9 +23,16 @@ \chapter{Debug Module (DM)} \label{dm}
...
-    registers that are present on the hart and listed in Table~\ref{tab:regno}.
+\item
+    \begin{steps}{Do at least one of:}
+        \item Implement the Program Buffer.
+        \item Implement abstract access to all registers that are visible to
+            software running on the hart including all the registers that are
+            present on the hart and listed in Table~\ref{tab:regno}.
+        \item Advertise the implementation as conforming to the ``Minimal
+            RISC-V Debug Specification \versionnum'', instead of the ``RISC-V
+            Debug Specification \versionnum''.
```
asb: I suggest adding "and implement abstract access to all GPRs and dcsr", so there is at least a defined minimal baseline.

asb:
It looks good from my perspective. I forgot about `dcsr`, so thanks for adding that.

PR #383, opened 2018-10-03
Make haltreq and resumereq proper write-only.

timsifive:
They already mostly were, but then I clarified some stuff and broke
keeping haltreq high during reset, which is the only mechanism to catch
a hart out of reset if optional reselthaltreq isn't implemented.

Also:
* resume ack is cleared by the DM when resumereq is written 1.
* haltreq can be high together with hartreset.

This is the minimal version of #378. It has all the same functionality,
but lacks some of #378's clarifications, which may be making this harder
to review.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -176,7 +176,7 @@ deasserted. Finally they end up either running or halted, depending on
 
 \section{Run Control} \label{runcontrol}
 
-For every hart, the Debug Module contains 4 conceptual bits of state: halt
+For every hart, the Debug Module tracks 4 conceptual bits of state: halt
 request, resume ack, halt-on-reset request,  and hart reset.
 (The hart reset and halt-on-reset request bits are optional.)
 These bits all reset to 0. A debugger can write them for the currently selected
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -177,24 +177,24 @@ deasserted. Finally they end up either running or halted, depending on
 \section{Run Control} \label{runcontrol}
 
 For every hart, the Debug Module contains 4 conceptual bits of state: halt
-request, resume request, halt-on-reset request,  and hart reset.
+request, resume ack, halt-on-reset request,  and hart reset.
 (The hart reset and halt-on-reset request bits are optional.)
 These bits all reset to 0. A debugger can write them for the currently selected
-harts through \Fhaltreq, \Fresumereq, \Fsetresethaltreq/\Fclrresethaltreq and
-\Fhartreset in \Rdmcontrol, only one of which may be high in any single write.
-In addition the DM receives halted, running, and resume ack signals from each
-hart.
+harts through \Fhaltreq, \Fsetresethaltreq/\Fclrresethaltreq and
+\Fhartreset in \Rdmcontrol.
+In addition the DM receives halted and running signals from each hart.
 
-When a running hart receives a halt request, it responds by halting,
+When a running hart sees its halt request bit high, it responds by halting,
 deasserting its running signal, and asserting its halted signal. The halted
 signals of all selected harts are reflected in the \Fallhalted and \Fanyhalted
-bits. \Fhaltreq is ignored by halted harts.
+bits. Halted harts ignore their halt request bit.
 
 When a halted hart receives a resume request, it responds by resuming, clearing
-its halted signal, and asserting its running signal and resume ack signals. The
-resume ack signal is lowered when the resume request is deasserted.  These
+its halted signal, and asserting its running signal. The DM
+reacts to a hart's running signal going high by setting resume ack for that
+hart. These
 status signals of all selected harts are reflected in \Fallresumeack,
-\Fanyresumeack, \Fallrunning, and \Fanyrunning. \Fresumereq is ignored by
+\Fanyresumeack, \Fallrunning, and \Fanyrunning. Resume requests are ignored by
 running harts.
 
 When halt or resume is requested, a hart must respond in
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -189,10 +189,10 @@ deasserting its running signal, and asserting its halted signal. The halted
 signals of all selected harts are reflected in the \Fallhalted and \Fanyhalted
 bits. Halted harts ignore their halt request bit.
 
-When a halted hart receives a resume request, it responds by resuming, clearing
-its halted signal, and asserting its running signal. The DM
-reacts to a hart's running signal going high by setting resume ack for that
-hart. These
+When a debugger writes 1 to \Fresumereq, each selected hart's resume ack bit is
+cleared and each selected, halted hart is sent a resume request. Harts respond
+by resuming, clearing their halted signal, and asserting their running signal.
+At the end of this process the resume ack bit is cleared.  These
 status signals of all selected harts are reflected in \Fallresumeack,
 \Fanyresumeack, \Fallrunning, and \Fanyrunning. Resume requests are ignored by
 running harts.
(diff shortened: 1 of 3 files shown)


debug_module.tex
```diff
@@ -177,24 +177,24 @@ \section{Hart States}
...
-request, resume request, halt-on-reset request,  and hart reset.
+request, resume ack, halt-on-reset request,  and hart reset.
 (The hart reset and halt-on-reset request bits are optional.)
 These bits all reset to 0. A debugger can write them for the currently selected
-harts through \Fhaltreq, \Fresumereq, \Fsetresethaltreq/\Fclrresethaltreq and
-\Fhartreset in \Rdmcontrol, only one of which may be high in any single write.
-In addition the DM receives halted, running, and resume ack signals from each
-hart.
+harts through \Fhaltreq, \Fsetresethaltreq/\Fclrresethaltreq and
+\Fhartreset in \Rdmcontrol.
```
mwachs5: We aren't explaining here how resume ack gets set, so it seems like a weird omission from this list. It is explained later, but shouldn't we say something like "A debugger can change their value for the currently selected hart through... \Fresumereq ... " 

I think it's weird to say resume ack is the bit of state. It's not like a hart sees the conceptual resume ack go low and then does something. Harts see the resume request go low, and then resume, and Debug module records that they resumed in resume ack

debug_module.tex
```diff
@@ -177,24 +177,24 @@ \section{Hart States}
...
+In addition the DM receives halted and running signals from each hart.
 
-When a running hart receives a halt request, it responds by halting,
+When a running hart sees its halt request bit high, it responds by halting,
```
mwachs5: I think this change is relevant to keep.

debug_module.tex
```diff
@@ -177,24 +177,24 @@ \section{Hart States}
...
 deasserting its running signal, and asserting its halted signal. The halted
 signals of all selected harts are reflected in the \Fallhalted and \Fanyhalted
-bits. \Fhaltreq is ignored by halted harts.
+bits. Halted harts ignore their halt request bit.
```
mwachs5: I think this is relevant to keep

debug_module.tex
```diff
@@ -177,24 +177,24 @@ \section{Hart States}
...
 
 When a halted hart receives a resume request, it responds by resuming, clearing
-its halted signal, and asserting its running signal and resume ack signals. The
-resume ack signal is lowered when the resume request is deasserted.  These
+its halted signal, and asserting its running signal. The DM
+reacts to a hart's running signal going high by setting resume ack for that
```
mwachs5: I don't know why we need to be so prescriptive about what the debug module does and what the hart does. I would rewrite this to say:

When the DM receives a resume request for a hart, that hart's resumeack bit is cleared and it's resume req bit is set. The hart responds by resuming, clearing its halted signal, and asserting its running signal. In this process the resumeack bit is cleared, either by the hart or the DM.

debug_module.tex
```diff
@@ -177,24 +177,24 @@ \section{Hart States}
...
+hart. These
 status signals of all selected harts are reflected in \Fallresumeack,
-\Fanyresumeack, \Fallrunning, and \Fanyrunning. \Fresumereq is ignored by
+\Fanyresumeack, \Fallrunning, and \Fanyrunning. Resume requests are ignored by
```
mwachs5: sounds good

debugger_implementation.tex
```diff
@@ -46,8 +46,9 @@ \section{Checking for Halted Harts}
 \section{Halting} \label{deb:halt}
 
 To halt one or more harts, the debugger selects them, sets \Fhaltreq, and then
-waits for \Fallhalted to indicate the harts are halted before clearing
-\Fhaltreq to 0.
+waits for \Fallhalted to indicate the harts are halted. Then it can clear
```
mwachs5: sounds good

debugger_implementation.tex
```diff
@@ -46,8 +46,9 @@ \section{Checking for Halted Harts}
...
+\Fhaltreq to 0, or leave it high to catch a hart that unexpectedly resets while
```
mwachs5: though I don't know if we need "unexpectedly". This is how you do it when you do `ndmreset`, and are expecting the hart to reset.

xml/dm_registers.xml
```diff
@@ -128,30 +128,29 @@
         \end{commentary}
 
         On any given write, a debugger may only write 1 to at most one of the
-        following bits: \Fhaltreq, \Fresumereq, \Fhartreset, \Fackhavereset,
+        following bits: \Fresumereq, \Fhartreset, \Fackhavereset,
```
mwachs5: sounds good

xml/dm_registers.xml
```diff
@@ -128,30 +128,29 @@
...
 
             Writes apply to the new value of \Fhartsel and \Fhasel.
         </field>
-        <field name="resumereq" bits="30" access="W" reset="-">
-            0: May cancel a resume request for any of the currently selected
-            harts, if those harts haven't resumed yet.
-
-            1: Causes the currently selected harts to resume once, if they are
-            currently halted.
+        <field name="resumereq" bits="30" access="W1" reset="-">
```
mwachs5: looks good

mwachs5:
SO I just re-read #378 for why you don't like calling resumereq a state bit. We can just say that once a hart resumes, the resumereq bit is self-cleared. It's a resume-once request.

debug_module.tex
timsifive: The DM clearing resume ack is already in the register definition, but we could add it here as well.

There is no resume request bit. There might be, but it's not state that has any reason to persist, beyond for implementation details.

How about:
When a debugger writes 1 to \Fresumereq, each selected hart's resume ack bit is cleared, and each selected, halted hart is sent a resume request. Harts respond by resuming, clearing their halted signal, and asserting their running signal. At the end of this process the resume ack bit is cleared.

mwachs5: Yes, that makes more sense to me.

debugger_implementation.tex
timsifive: Removed "unexpectedly."

debug_module.tex
mwachs5: Ok, fine if you want to track resumeack instead of resumereq. But this paragraph is weird from a readability standpoint.  This paragraph starts with a list of bits and it seems like it's about how to modify those bits. But you don't in this paragraph actually say anything about resume ack bit. Yes, the debugger doesn't directly "set" resume ack, but it does modify it:  it can be cleared by a write to \Fresumereq as you describe below. Does it need a forward reference?

debug_module.tex
timsifive: Changed.

debug_module.tex
timsifive: I'm tempted to make this paragraph simply:
For every hart, the Debug Module contains 4 conceptual bits of state: halt
request, resume ack, halt-on-reset request,  and hart reset.
(The hart reset and halt-on-reset request bits are optional.)
These bits all reset to 0. ~~A debugger can write them for the currently selected
harts through \Fhaltreq, \Fsetresethaltreq/\Fclrresethaltreq and
\Fhartreset in \Rdmcontrol.~~
In addition the DM receives halted and running signals from each hart.

Then update the following paragraphs with a sentence on which bit the debugger writes to get a conceptual bit/signal to do something.

Thoughts?

mwachs5: Yeah

mwachs5: Also maybe “contains” should be “tracks” or “maintains”?

timsifive: Changed to tracks.

PR #391, opened 2018-10-15
Bump version to 0.13.1.

timsifive:


timsifive:
Can I get a review on this?

mwachs5:
Should we make an `0.13.1` branch for this vs pushing to the current `master`? Is the idea to keep bumping the minor version number every time we add a change, or...?

timsifive:
`master` should always point to the latest work in progress, which currently is 0.13.1-DRAFT. We'll keep using 0.13.1-DRAFT until that gets ratified as 0.13.1, or until we need to make incompatible changes. (At that point we may end up with a compatible and incompatible branch.)

mwachs5:
Why don't we create a tag for 0.13, and then explain the above rationale in the README for this repository, also pointing people to the errata document (or maybe that goes in the other PR)

timsifive:
I was going to create a tag for 0.13 once it's ratified, to avoid the case of us tagging something that turns out not to be true. Once that is created, I agree it's a good idea to update the README. Currently there is a tag for what we sent off to be ratified.

timsifive:
Can you continue this discussion/approve the change?

PR #392, opened 2018-10-15
Fix sbdata0 read order of operations.

timsifive:
Also created an errata document, since this is a real bug in the spec.
I've implemented the new semantics in spike and confirmed OpenOCD still
"just works." Rocket had implemented the right thing from the start.


timsifive:
Can I get a review on this?

mwachs5:
Presumably this is blocked by #391 ?

rmac-sifive:
Hey @timsifive, LGTM, but I think the spec would be more clear if there were an explicit note that the increment doesn't happen until the system bus memory transaction completes

README.md
```diff
@@ -1,7 +1,14 @@
 RISC-V Debug Specification
 ==========================
 
-A pre-built version of the spec is included [here](https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-draft.pdf).
+This repository contains the latest draft version of the Debug Specification,
+as well as an errata document that accompanies any ratified versions of the
+Debug Specification.
+
+The following pre-built PDFs may be of interest:
+* [0.13-DRAFT, which will likely be ratified](https://github.com/riscv/riscv-debug-spec/blob/task_group_vote/riscv-debug-draft.pdf)
```
mwachs5: "likely be ratified" -> "has been submitted to the TC for ratification"?

timsifive: I think the former is more clear to people not familiar with the ratification process. Do you prefer to hedge a bit?

timsifive:
@mwachs5 Yes, #391 must merge first (and once this merges, I'll work on #393).

@rmac-sifive I've clarified the language in the draft spec, but not in the errata. I feel the errata is only for fixing real bugs, in as simple a way as possible to avoid introducing new bugs.

README.md
mwachs5: I think these README edits belong with #391.

README.md
```diff
@@ -1,7 +1,14 @@
...
+* [Errata Document](https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-errata.pdf)
+* [Latest draft](https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-draft.pdf)
```
mwachs5: Should we say something in the README about the rationale of 0.13.1?

I was assuming it was as described on: https://semver.org/
```
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

```

README.md
timsifive: Yes, but I wanted to mention the errata doc in the README, which isn't added until this change. I could have made changes in both and then resolve conflicts later, but this seemed much simpler.

README.md
timsifive: The spec itself should mention all previous ratified versions, and their compatibility, but not until there actually is a ratified spec.

timsifive:
This change now only fixes the actual spec bug.

PR #404, opened 2018-11-06
Add halt groups and external triggers.

timsifive:
This allows a group of harts to simultaneously halt when any of them
hits a breakpoint, or even when some other event occurs in the system.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -221,7 +221,9 @@ while the hart is selected, or by DM reset.
 
 An optional feature allows a debugger to create halt groups. When any hart in a
 halt group halts all the other harts in that group will quickly halt, even if
-they are currently in the process of resuming.
+they are currently in the process of resuming. Adding a hart to a halt group
+does not automatically halt that hart, even if other harts in the group are
+already halted.
 
 It is also possible to add external triggers to a halt group. External triggers
 are abstract concepts that can signal the DM and/or receive signals from the
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -217,6 +217,31 @@ The hart's halt-on-reset request bit remains set
 until cleared by the debugger writing 1 to \Fclrresethaltreq
 while the hart is selected, or by DM reset.
 
+\section{Halt Groups and External Triggers}
+
+An optional feature allows a debugger to create halt groups. When any hart in a
+halt group halts all the other harts in that group will quickly halt, even if
+they are currently in the process of resuming.
+
+It is also possible to add external triggers to a halt group. External triggers
+are abstract concepts that can signal the DM and/or receive signals from the
+DM.  When an external trigger fires, all harts in its halt group are quickly
+halted. When a hart in a halt group halts, the external triggers in the same
+halt group are notified.
+
+\begin{commentary}
+    External triggers could be used to implement near simultaneous halting of
+    all cores in a system, when not all cores are RISC-V cores.
+\end{commentary}
+
+Halt group 0 is special, and has none of this behavior. Harts in halt group 0
+halt as if halt groups aren't implemented at all.
+
+To near simultaneously resume all harts in a group, the debugger can select all
+the harts, and write 1 to \Fresumereq as described above. An implementation
+must guarantee that all harts in the group halt again as soon as one of them
+halts, even if that hart halts before all harts have resumed.
+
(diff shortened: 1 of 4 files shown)


xml/dm_registers.xml
```diff
@@ -540,6 +540,46 @@
         <field name="data" bits="31:0" access="R/W" reset="0" />
     </register>
 
+    <register name="Advanced" short="advanced" address="0x32">
+        <field name="0" bits="31:11" access="R" reset="0" />
+        <field name="exttrigger" bits="10:7" access="WARL" reset="0">
+            This field contains the currently selected external trigger.
+
+            If a non-existant trigger value is written here, the hardware will
```
ernie-sifive: non-existant should be non-existent.

xml/dm_registers.xml
```diff
@@ -540,6 +540,46 @@
         <field name="data" bits="31:0" access="R/W" reset="0" />
     </register>
 
+    <register name="Advanced" short="advanced" address="0x32">
```
mwachs5: Can we give this a name something more descriptive of what is in it than "advanced". At the very least, can it be "advancedcs" for "advanced control and status" which was more following our convention? There is no description for this register. What is envisioned for current and/or future contents of the "Advanced" register?

timsifive: Yes, but I'm not sure what a better name is. I think of this as a second DM control and status register. Except there is no first one, because it's separated into dmcontrol and dmstatus.

Should I split this into dmcontrol2 and dmstatus2? Call it dmcs2?

mwachs5: dmcs2 makes sense to me

timsifive: I've renamed it.

mwachs5: Do we usually write descriptions for the registers themselves (not just the fields)? Should there be one for this new register?

debug_module.tex
```diff
@@ -217,6 +217,31 @@ \section{Run Control} \label{runcontrol}
 until cleared by the debugger writing 1 to \Fclrresethaltreq
 while the hart is selected, or by DM reset.
 
+\section{Halt Groups and External Triggers}
+
+An optional feature allows a debugger to create halt groups. When any hart in a
```
mwachs5: elsewhere in this document we ahve things like "lists of features supported by this specification". If this list still exists, should this be added to that list?

xml/dm_registers.xml
timsifive: I had added:
This register contains DM control and status bits that didn't easily
        fit in \Rdmcontrol and \Rdmstatus, and are only likely to be
        implemented in larger systems.

introduction.tex
```diff
@@ -170,6 +170,9 @@ \section{Supported Features}
        involving any hart. (Optional)
    \item A RISC-V hart can be halted when a trigger matches the PC,
        read/write address/data, or an instruction opcode. (Optional)
+    \item Harts can be grouped, and harts in the same group will all halt when
+        any of them halts. These groups can also react to or notify external
```
mwachs5: "notify" external triggers? Or "raise"/"cause" external triggers? I don't know what it means to "notify" a trigger.

xml/dm_registers.xml
```diff
@@ -540,6 +540,50 @@
         <field name="data" bits="31:0" access="R/W" reset="0" />
     </register>
 
+    <register name="Debug Module Control and Status 2" short="dmcs2" address="0x32">
+        This register contains DM control and status bits that didn't easily
+        fit in \Rdmcontrol and \Rdmstatus, and are only likely to be
+        implemented in larger systems.
```
mwachs5: what is a "larger" system... you mean one with many harts?

introduction.tex
timsifive: How about "react to or assert external signals"? (Together with rewriting other uses of trigger in this context to signal.) Trigger inherently implies an input.

xml/dm_registers.xml
timsifive: Systems that are so large they don't mind an extra hundred gates for debug functionality. Is there a better way to say that?

mwachs5: Why not just say "are optional".

timsifive: Done.

Issue #409, opened 2018-11-22
the reset value (0) of tcontrol.mte is not backward compatible with the debugger behavior of previous spec


chuanhua:
The reset value (0) of tcontrol.mte will cause the debugger of the previous debug spec (no concept of mte) to fail to turn on debug-mode trigger on M-mode code. The old debugger will only set action to 1 to turn on debug-mode trigger on M-mode code. So this is not backward-compatible to the old debugger.

Two possible solutions:
1. change the mte reset value to 1.
2. add a constraint that mte only has effect when action=0.

Thanks.

Chuanhua

chuanhua:
Thanks for resolving this issue.

Chuanhua
Andes Technology

PR #411, opened 2018-12-13
mte only applies when action=0

timsifive:
Addresses #409.


xml/hwbp_registers.xml
```diff
@@ -129,9 +129,9 @@
         <field name="mte" bits="3" access="R/W" reset="0">
             M-mode trigger enable field.
 
-            0: Triggers do not match/fire while the hart is in M-mode.
+            0: Triggers with action=0 do not match/fire while the hart is in M-mode.
 
-            1: Triggers do match/fire while the hart is in M-mode.
+            1: Triggers with action=0 do match/fire while the hart is in M-mode.
```
ernie-sifive: Saying triggers with action=0 fire almost implies that triggers with other actions don't fire.  What do you think about a clarifying statement saying that mte has no effect on triggers with action!=0?

timsifive: How about reverting just the 1 line:
> 1: Triggers do match/fire while the hart is in M-mode.

Issue #416, opened 2019-01-08
Should decide/document whether/if haltreq causes halt out of reset


timsifive:
See mailing list subject:"Halt on reset"

PR #419, opened 2019-01-09
Clarify that harts halt out of reset if haltreq=1

### LINKED ISSUE #413: there is no space for custom trigger types
We should reserve trigger types 12--14 for custom use.

timsifive:
Fixes #413.
This is a pretty important clarification. Is it an erratum?

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -102,6 +102,9 @@ can be cleared by writing 1 to \Fackhavereset in \Rdmcontrol.
 The {\tt havereset} bits may or may not be cleared
 when \Fdmactive is low.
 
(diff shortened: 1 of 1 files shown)


mwachs5:
I don't think this is an erratum. Elsewhere in the spec we clearly state that the intention is to halt a hart immediately out of reset before any instructions have executed.

timsifive:
@mwachs5 
> I don't think this is an erratum. Elsewhere in the spec we clearly state that the intention is to halt a hart immediately out of reset before any instructions have executed.

Can you point me where we state that intention? I couldn't find it when I went looking for it, which is why I wrote this clarification.

timsifive:
Fixes #416.

PR #423, opened 2019-02-07
Create bugfix 0.13.1.

timsifive:
This replaces the 0.13.1 draft (which will be renumbered to 0.14) and
the errata document. The bugfixed document will live in this branch
(release), and be the one that is distributed (e.g. on riscv.org).

I went through the existing errata document
(https://github.com/riscv/riscv-debug-spec/blob/800450fb0ac1c3a62909b0ccb587fe4adcfe5323/riscv-debug-errata.pdf)
and applied all those changes to this document. Then I wrote a small
section on versioning to explain what is going on. The new document also
contains a list of bug fixes that have been applied since 0.13.


timsifive:
Calling it good, then.

PR #424, opened 2019-02-08
Errata go in 0.13.x, this is 0.14.

timsifive:
Part of changes to roll the errata into the main document, so people
don't have to chase down a separate document to check if the ratified
one is sane.


timsifive:
That's exactly right. And 0.13.X will be distributed as "the spec" until 0.14 is ratified. (There's no schedule for the latter, and I wouldn't expect it to happen soon.)

If you have any questions, let me know because I don't want to start distributing 0.13.1 and then discover that's not what we wanted after all.

ernie-sifive:
I agree this approach will end up being cleaner and easier for consumers of the spec.  So the errata discovered and fixed between 0.13 ratification and today will be released in 0.13.1.  Will each new fixed erratum result in another bugfix revision, i.e. 0.13.2, 0.13.3, etc, or would we group several errata together in each new bugfix release (say, monthly or quarterly)?

PR #425, opened 2019-02-12
Clarify that harts halt out of reset if haltreq=1 (#419)

timsifive:
This change already went into master. This PR is to add the same text into release as a bugfix.

I don't believe the release version mentions anywhere that haltreq applies coming out of reset. It used to be in there, and is still the intention. There was some e-mail discussion about this: subject:"Halt on reset"


timsifive:
Brought this up in today's meeting, and people agree that this is an important bugfix.

PR #430, opened 2019-02-15
Try to get travis to build the release branch.

timsifive:


timsifive:
I'm pretty vague on what travis looks at, but I think it might look at the master of this file to determine which branches to build, so that means it won't do anything until this change actually merges.

Issue #422, opened 2019-01-31
"tselect" and "tdata1" always present when trigger module is not implemented?


chuanhua:
From the Trigger Module description (chap. 5), it seems to imply that the "tselect" and ("tdata1" or "tinfo") has to be always present in the CSR space, tselect as R/W and tdata1/tinfo as RO, even when Trigger Module is not implemented.

Is this true? If this is true, then some more explicit description of this fact should be added to clarify it.

Thanks.

Chuanhua Chang
Andes Technology

timsifive:
That is true. In systems that don't have any triggers, tselect, tdata1, and tinfo can always return 0 when read, which should be very cheap to implement.

chuanhua:
Thank you for the clarification. But, the spec says that tinfo needs to return 1, right?

chuanhua:
In the trigger module chapter, it has the following paragraph:
-----
A hart can be compliant with this speciﬁcation without implementing any trigger functionality at
all, but if it is implemented then it must conform to this section.
-----
Then the following paragraphs describes the procedure to find out how many triggers are supported. The quoted paragraph let me think that the procedure is for a design that implements the trigger module. And for a design that does not implement the trigger module, it does not need to conform to the procedure for finding out how many triggers are implemented. And it seems that "tselect" and "tdata1" CSRs do not need to always be present. 

So I think a more explicit clarification is needed.

timsifive:
Thinking about this more, if there are no triggers I don't really see a problem with allowing accesses to tselect/tdata* to result in an illegal instruction exception. I agree that the old spec was ambiguous here, but good software should already handle such errors gracefully.

Is this what you had in mind in the first place?

mwachs5:
I don't think we can mandate the existence of any of the trigger registers as the priv spec is fairly clear about what registers return 0 vs cause exception if not impelmented. 

I agree there isn't a problem with allowing accesses to trigger registers to result in illegal instruction exception if they are not present.

chuanhua:
Either way is fine with me. But if an illegal instruction exception is allowed, could you add the following sentence after the previously quoted paragraph to make this more explicit?

"If the trigger module is not implemented, accessing to any of the trigger module CSRs may/will generate an illegal instruction exception."

PR #431, opened 2019-02-20
All trigger registers are optional

timsifive:
See #422. This might break software that's not ready for the
possibility, but I believe the privilege spec is already clear that the
Debug CSRs are optional.

### DIFF
--- a/trigger.tex
+++ b/trigger.tex
@@ -10,14 +10,16 @@ Trigger Module is broken out as a separate piece that can be implemented
 separately.
 
 A hart can be compliant with this specification without implementing any
-trigger functionality at all, but if it is implemented then it must conform
-to this section.
+trigger functionality at all, but if it is implemented then it must conform to
+this section. If triggers aren't implemented, the CSRs may not exist at all and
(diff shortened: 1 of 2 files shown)


ernie-sifive:
I wonder if it would be useful to remind readers in the trigger intro section that if the optional trigger registers don't exist, then accessing them will cause an illegal instruction exception.  The current sequence described in the intro would fail after step 1 (write 0 to tselect) with an exception.

A related question would be how a debugger is supposed to determine if triggers exist.  Is the only way to attempt to write tselect and see if it generates an exception?

timsifive:
> I wonder if it would be useful to remind readers in the trigger intro section that if the optional trigger registers don't exist, then accessing them will cause an illegal instruction exception. The current sequence described in the intro would fail after step 1 (write 0 to tselect) with an exception.

Good suggestion. I've clarified that.

> A related question would be how a debugger is supposed to determine if triggers exist. Is the only way to attempt to write tselect and see if it generates an exception?

That's right. If a debugger only cares about the number of registers, then a binary search should find that relatively quickly.

PR #433, opened 2019-03-15
Explain how to detect the version.

timsifive:
Also make some more required guarantees about future versions.

Unfortunately there are a number of side effects to reading the version,
but none of them are a big deal.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -535,6 +535,7 @@ while writes should be ignored, with the following mandatory exceptions:
     \item Write \Rdmcontrol, preserving \Fhartreset, \Fhasel, \Fhartsello, and
         \Fhartselhi from the value that was read, setting \Fdmactive, and
         clearing all the other bits.
+    \item Read \Rdmstatus, which contains \Fversion.
 \end{steps}
 
 \begin{steps}{This has the following unavoidable side effects:}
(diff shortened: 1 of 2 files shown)


markuslauterbach:
I think the version detection procedure should include the step that actually reads version, i.e. "read dmstatus" as last step.

Also, the procedure describes preserving dmcontrol.hartreset/hasel/hartsello/hartselhi, and guarantees that this will work with all versions. This implies that for all future spec versions, dmcontrol will always be existent at its current address, and the hartreset/hasel/hartsello/hartselhi bits will also always be existent at their positions for all future versions.

If this is really intended, then we should not only indirectly imply it, but say it explicitly instead. To my knowledge, currently we only explicitly say this about dmstatus.version: " Its address will not change in the future, because it contains version." Strictly speaking, even here the spec is only explicit about the address, and only implies that dmstatus will always contain 'version' at its current position.

If it is not intended that hartreset/hasel/hartsello/hartselhi will always exist and be where they are, then the procedure is too version-dependent.

timsifive:
> I think the version detection procedure should include the step that actually reads version, i.e. "read dmstatus" as last step.

Good point. I'll change that.

> Also, the procedure describes preserving dmcontrol.hartreset/hasel/hartsello/hartselhi, and guarantees that this will work with all versions. This implies that for all future spec versions, dmcontrol will always be existent at its current address, and the hartreset/hasel/hartsello/hartselhi bits will also always be existent at their positions for all future versions.

I disagree. It implies that the described procedure is safe for whatever bits exist at those locations, but it doesn't say these bits have the same function, or any function at all. I'll make that more explicit.

markuslauterbach:
> 
> 
> I disagree. It implies that the described procedure is safe for whatever bits exist at those locations, but it doesn't say these bits have the same function, or any function at all. I'll make that more explicit.

Okay thanks, looks good to me know.

Issue #441, opened 2019-04-22
Assumptions on type field of tdata1?


neelgala:
Have a couple of questions on the type field of tdata1.

1. Why do we need type=0 to indicate that the trigger doesn't exist? We can get that info from tselect and tinfo itself? Isn't this redundant?

2. From what I am seeing implemented in riscv-openocd ([here](https://github.com/riscv/riscv-openocd/blob/15f2d35bc6413d89453c64eff39a80dfc52243f5/src/target/riscv/riscv.c#L470-L484)) it looks like the type is supposed to be a read-only field and cannot change at runtime. Is that true? If not, I am happy to fix the logic in riscv-openocd which assumes this.

3. Is it mandatory to imlpement all the encodings of the type field (0, 2, 3, 4, and 5)? If not, there should be discovery routine as well to identify which types are supported on an available trigger.

4. Since we are already at 0.14 version does it still make sense to have the SiFive legacy encoding of 1.

timsifive:
On Mon, Apr 22, 2019 at 3:38 AM Neel Gala <notifications@github.com> wrote:

> Why do we need type=0 to indicate that the trigger doesn't exist? We can
> get that info from tselect and tinfo itself? Isn't this redundant?
>
It's allowable to make tselect e.g. 3 bits wide, even if there are fewer
than 8 triggers. The idea was that this might be a simpler implementation
option than putting extra logic on tselect accesses to make sure e.g. 7
cannot be written there. In that case trigger 7 can be selected, and when
read back its type will show to be 0.

From what I am seeing implemented in riscv-openocd (here
> <https://github.com/riscv/riscv-openocd/blob/15f2d35bc6413d89453c64eff39a80dfc52243f5/src/target/riscv/riscv.c#L470-L484>)
> it looks like the type is supposed to be a read-only field and cannot
> change at runtime. Is that true? If not, I am happy to fix the logic in
> riscv-openocd which assumes this.
>

Type may not be read-only, but OpenOCD doesn't support writing the type.

Is it mandatory to imlpement all the encodings of the type field (0, 2, 3,
> 4, and 5)? If not, there should be discovery routine as well to identify
> which types are supported on an available trigger.
>

No. Everything about triggers is optional. It's assumed the debugger will
attempt to set whatever trigger it wants, and then discover if it happened
or not. If so it can try something else or report an error to the user.

For types specifically, the tinfo register can be used to discover which
types are supported by a trigger.

[...]

PR #444, opened 2019-04-23
All tdata functionality is optional...

timsifive:
... so don't mention on a few bits that they are optional, because
that's confusing (see #441).


timsifive:
Waiting on @neelgala to confirm (probably in #441) that this change is clear enough.

PR #450, opened 2019-05-10
Create feature discovery.

timsifive:


timsifive:
> Table 3.2 needs to be updated (in debug_module.tex) to include the new command.

Done. Thanks.

allenjbaum:
Might there be a want (or a need) to have some kind of padding to make sure that structures can be 4byte aligned in systems that don't have unaligned access? Or do we have enough flexibility? (e.g.we can pad sections with EndOfData section header bytes until aligned, and make sure the section body size is always a multiple of 4 or 8. Also varnums can choose a larger size than required to maintain alignment).

Q: does the section size include only bytes after the 2 varnum args?
The section table (that lists sectionIDs 0,10,29,30,31) is a bit deceptive, in that the type and name are not adjacent, because there is the section body size varnum in between. Would it make sense to reverse the order of section_type, section_body_size? In that case we might not need EndOfData section types, since a size of zero would be used to indicate no more sections exist. 

Was there some reason that 29 was chosen as Magic rather than 0?

Krste brought up more other considerations than I thought possible.

One was that sections need not be contiguous. I think this structure handles that by making the section length really be an offset to the next section, and using section specific encodings to indicate that the section is ended (e.g.ranges with hi<lo, masks of zero, etc). So the parser would end a section either when it's parsed the number of byes indicated, or when a section specific config is encountered. Another possibility is a dummy section type that has no data except a pointer to a nonsequential address or offset.

My preference is that the format be binary rather than human readable. Parsers will need to exist to use the data anyway, and I suspect that it can be used to easily convert to a human readable form, and that both the runtime parser and the data will be much smaller, which will be important on the microcontroller implementations.

I think Bi-endian will only be useful for implementations that are bi-endian... in which case theyll be able to do bi-endian reads anyway., so keep it little-endian.

I think your format keeps the parser simple, and the parser and data size small, two other concerns of Krste's.

Finally: There will be uses of this that will want to be runtime visible, I believe (or, at least, we shouldn't rule it out).
For both use cases, we should have an architectural method of finding this data.
From the debugger side, that could mean piggybacking on the existing config_string pointer.
For the runtime, we may want a new CSR that either points to the MMIO locations, or allows writes to the CSR to set offsets, and redas (with autoincrement) to to  read the data.

timsifive:
> Might there be a want (or a need) to have some kind of padding to make sure that structures can be 4byte aligned in systems that don't have unaligned access? Or do we have enough flexibility? (e.g.we can pad sections with EndOfData section header bytes until aligned, and make sure the section body size is always a multiple of 4 or 8. Also varnums can choose a larger size than required to maintain alignment).

I imagined that parsers would always read byte-by-byte. That's still easy to implement, and for the amount of data we're talking not significantly slower than reading a word at a time.

> Q: does the section size include only bytes after the 2 varnum args?
> The section table (that lists sectionIDs 0,10,29,30,31) is a bit deceptive, in that the type and name are not adjacent, because there is the section body size varnum in between. Would it make sense to reverse the order of section_type, section_body_size? In that case we might not need EndOfData section types, since a size of zero would be used to indicate no more sections exist.

> Was there some reason that 29 was chosen as Magic rather than 0?

Yes. It's not 0 (0 is not a good magic since it's often used as filler), and when encoded as a varnum it's 0x52, which is ASCII for 'R'. Maybe not a good reason, but that's the reason.

> Krste brought up more other considerations than I thought possible.

That's why it's so important to get input from other people/groups. Which doesn't seem to be happening. :-(

> One was that sections need not be contiguous. I think this structure handles that by making the section length really be an offset to the next section, and using section specific encodings to indicate that the section is ended (e.g.ranges with hi<lo, masks of zero, etc). So the parser would end a section either when it's parsed the number of byes indicated, or when a section specific config is encountered. Another possibility is a dummy section type that has no data except a pointer to a nonsequential address or offset.

I'm thinking of adding a Jump section, which can do any of:
1. continue parsing at another offset (relative or absolute)
2. include sections located elsewhere (relative or absolute) and then resume parsing from here

> My preference is that the format be binary rather than human readable. Parsers will need to exist to use the data anyway, and I suspect that it can be used to easily convert to a human readable form, and that both the runtime parser and the data will be much smaller, which will be important on the microcontroller implementations.

I strongly agree. There are a number of companies that would decide not to implement this if the data is too large, which human-readable will probably be.

> I think Bi-endian will only be useful for implementations that are bi-endian... in which case theyll be able to do bi-endian reads anyway., so keep it little-endian.

I think of parsing as happening byte by byte, so endianness is irrelevant.

> Finally: There will be uses of this that will want to be runtime visible, I believe (or, at least, we shouldn't rule it out).
> For both use cases, we should have an architectural method of finding this data.
> From the debugger side, that could mean piggybacking on the existing config_string pointer.
> For the runtime, we may want a new CSR that either points to the MMIO locations, or allows writes to the CSR to set offsets, and redas (with autoincrement) to to read the data.

I agree that this structure will be useful to e.g. an OS that should be able to find it at run-time. How that happens is beyond the scope of the debug group. I'm trying to define something that suits our needs, and is general enough to also work for other groups.

markuslauterbach:
** Hart Select: "Note that hart id may not be the same as hart index, which is used in much of
the Debug Specifi cation." Why? Wouldn't it be easier if this number would always correlate to the hart index that can be selected in dmcontrol?

** Trigger Support: "data and mask are XLEN-bit values". Shouldn't they be varnums (of size XLEN) as well? I think the parser does not know XLEN (see also next point).

Would it further make sense to have a section for "hart information"? Information could include: 
- base ISA / XLEN
- information that equals misa, mvendorid, marchid, mimpid (these registers are all optional, reading them is intrusive, reading them in one go requires knowledge of XLEN).

Ernie recently asked me how to implement system/hart discovery+identification for RISC-V in CoreSight that would resemble to Arm's core ID registers. This may be one way to have a standardized approach for this.

rtwfroody:
> ** Hart Select: "Note that hart id may not be the same as hart index, which is used in much of
the Debug Specifi cation." Why? Wouldn't it be easier if this number would always correlate to the hart index that can be selected in dmcontrol?

This is to make the structure useful outside the context of an external debugger.

Currently XLEN can already be determined by attempting register reads. (That's what OpenOCD does, at any rate.) It's not great because it requires halting the target.

I intentionally limited the scope of this proposal to the bare essentials. In general I fully agree that this structure could completely describe the system, and that would be useful in all kinds of contexts. (The most obvious being an OS booting up.) To that end I've tried to get some interest in it from other people, but so far without success.

I'll see if I can spec some more sections later this week, although I'm happy to look at your contributions as well.

allenjbaum:
The issue here is that hart index is (intended to be) fixed (e.g physical
XY location), while hartID could change at manufacturing, or even at boot,
and therefore there is some kind of mapping involved. E.g. in a 3 core
system, a manufacturer would sell it as 2 cores because one is defective,
but hartID 0 must always exist. If that's the core that is defective, then
you need remapping.
You (well, I) would prefer that the debug path to a hart be fixed; but SW
debugging will want to use hartID, so the DM will need to be aware of that
mapping to actually talk to any specific hart.

And, this brings up another thing that debug discovery needs: hartID to
hart index mapping.
Can we add this to the list?

timsifive:
> ** Hart Select: "Note that hart id may not be the same as hart index, which is used in much of
the Debug Specifi cation." Why? Wouldn't it be easier if this number would always correlate to the hart index that can be selected in dmcontrol?

This is to make the structure useful outside the context of an external debugger.

Currently XLEN can already be determined by attempting register reads. (That's what OpenOCD does, at any rate.) It's not great because it requires halting the target.

I intentionally limited the scope of this proposal to the bare essentials. In general I fully agree that this structure could completely describe the system, and that would be useful in all kinds of contexts. (The most obvious being an OS booting up.) To that end I've tried to get some interest in it from other people, but so far without success.

I'll see if I can spec some more sections later this week, although I'm happy to look at your contributions as well.

markuslauterbach:
> Currently XLEN can already be determined by attempting register reads. (That's what OpenOCD does, at any rate.) It's not great because it requires halting the target.

Correct, and that is also what our debugger currently does. However: 
I think that a successful feature discovery should not depend on the success of any other preceding advanced discoveries or other advanced pre-requirements (especially if these other discoveries are intrusive and require the harts to be not only halted but also available at all). Also, we could see feature discovery as (part of) a solution to solve exactly these "intrusive discovery" issues.

> I intentionally limited the scope of this proposal to the bare essentials.

Good point, we should first have a look at the bigger picture and decide whether we want to go down that path in the first place.

allenjbaum:
Markus: Do you have a specific example of what you don't want to see?
e.g. the spec documents a method to discover hartID:hartIndex mappings, by
selecting each hartIndex and reading its HartID CSR - but that may require
stopping harts, which you want to avoid. A dedicated hart mapping section
in the discovery structure would avoid that.

The rest of this is a taxonomy (possibly incomplete) of the options to
implementing discovery that includes core self-discovery.
---------------------------------
This discovery feature defines a data structure format, and a method (using
abstract commands) that the debugger can use to query it.
In the larger picture, cores will want discovery as well. The data
structure easily supports this.
But abstract commands don't because that's a debugger only feature (and
optional at that...)
IF we assume that cores will be able to do self discovery, then there are
options which may or may not be acceptable:

The data structure is at most 16MB (given the abstract_command register
format- we may want to re-think the command number to leave room for
expansion).
This 16MB space could be
 - accessible only to cores via normal loads and store (i.e. the 16MB
structure is an MMIO address)
 - accessible only to cores via CSRs (e.g. hart CSRs that replicate
something like the abstract_command and Data0 Debug regs, )
 - is only a dedicated ROM that is part of the DM, or
 - accessible to both: directly to the DM, and to cores either via CSRs or
via Ld/St
If the first,  then
   --the DM needs to know where in the larger (32b+) address this 16MB
section starts; it almost certain can't be at base=0, and either
     --- System Bus Access is implemented or
     --- cores must be halted and code downloaded to the program buffer
must be executed
If the second
   -- cores must be halted and code downloaded to the program buffer must
be executed
If the third, then we may need a separate implementation for other types of
discovery that cores want to do.
If both, and core access is via loads and stores, then cores still need a
way to discover the structure base address.

markuslauterbach:
> Markus: Do you have a specific example of what you don't want to see?
e.g. the spec documents a method to discover hartID:hartIndex mappings, by
selecting each hartIndex and reading its HartID CSR - but that may require
stopping harts, which you want to avoid. A dedicated hart mapping section
in the discovery structure would avoid that.

What I wouldn't want to see is that this hartID:hartIndex mapping is only available for certain chip- or hart states, and that a successful mapping requires harts to be powered and available (it's not only about halted harts, but about available harts in general - e.g. during early boot stages).

If a "fixed" feature is only chip-dependent and does not depend on any state of the chip or the state any of its harts, then the debugger should be able to directly discover this feature non-intrusive, and even when none of the harts are currently available.

Going with the "dynamic" hart ID instead of the "static" hart index makes sense for a hart "self discovery" because then the hart is up and running anyway, but for certain external debugger scenarios this may add (unnecessary) extra hurdles because the debugger is only interested in the hart index anyway.

allenjbaum:
I agree that this mapping should not be dynamic - but fixed.
Given that: a hart may be dynamically enabled or powered, and that can be
queried using the fixed mapping to see if the hart is available and/or
existent.

markuslauterbach:
> I agree that this mapping should not be dynamic - but fixed.
Given that: a hart may be dynamically enabled or powered, and that can be
queried using the fixed mapping to see if the hart is available and/or
existent.

Earlier you said that hart ID can change, e.g. during boot, from which I concluded that the mapping can not (always) be fixed. If the ID of a hart can really change its value (during run-time), then the index:id mapping and all references of that hart ID in the discovery ROM would need to be changed accordingly.
Admittedly I have no further knowledge about if and under which circumstances mhartid can change its value - the spec at least defines it as 'read-only'.

allenjbaum:
It there is fusing to enable/disable cores, then it depends on precisely
when/how those fuses are distributed and used to control the mapping.
My stake in the ground is that it all happens during the
reset/initialization sequence, and before the debugger takes control, so as
far the the debugger is concerned, it is fixed.

markuslauterbach:
> It there is fusing to enable/disable cores, then it depends on precisely
when/how those fuses are distributed and used to control the mapping.
My stake in the ground is that it all happens during the
reset/initialization sequence, and before the debugger takes control, so as
far the the debugger is concerned, it is fixed.

Is that reset/initialization sequence part of a boot sequence of e.g. the master core? If so, then this is where the debugger already needs control. Debugging usually starts right from the reset vector of the master core (in order to be able to debug any boot sequences and/or reset/initialization sequences).

timsifive:
I agree with Markus' goal.

I think what Allen is saying is that in his system, hart id to index mapping happens at power-up, *before* the debugger has any chance of doing anything. That is fine, as long as this mapping is reflected in the hart index to hartid map in the data structure when the debugger has a chance to read it.

Allen, does this map stay the same across resets? It would be nice if the debugger can assume the hardware won't change from under it when it resets, for all kinds of reasons.

allenjbaum:
If the mapping is in fuses, then I would expect that it is pretty well
fixed and nailed down - no changes after reset.
I can't come up with a good scenario for making the map change between
resets, even for a chip that could enable different numbers of cores for
different power scenarios or something.
In those cases, you could use the same mapping but make some unavailable or
even nonexistent.

But my interpretation of Markus's  concern was with a pre-boot
(initialization) scenario, when there is a management
engine/root-of-trust/etc, that has control before the risc-v cores even
come out of reset - he wants to ensure that those can be traced/debugged as
well, and he is entirely correct that this must be possible.

My feeling about that scenario is that it is a separate debug domain,
possibly with a separate DM and not even necessarily risc-v compatible.
You should always be able to do discovery of what you are going to debug,
and that discovery and mapping should be fixed at the time that the
debugger is capable of performing commands, in whichever domain you are
debugging - but not necessarily before.

Which begs a question: how does the debugger even know when anything will
respond to its commands?
In the above scenario, the DM for risc-v cores may be in reset during the
time that the ME/ROT/etc. is performing initialization.
What happens if someone tries to issue a command at that time - or is the
interface such that the external debugger knows when the DM is capable of
responding?

timsifive:
It sounds like we actually all agree that hart id to index mappings should be static and not change.

Allen, you raise some interesting issues regarding how a debugger can know when the target is ready for it. Can you start that discussion on the mailing list? It doesn't seem relevant to this pull request.

timsifive:
@markuslauterbach Does this PR give you all the information you would want? Does the format seem reasonable? What changes would you like to see?

markuslauterbach:
> Does this PR give you all the information you would want? Does the format seem reasonable? What changes would you like to see?

In general this discovery could be a very useful tool, and resolve several uncertainties and avoid intrusive behaviors. From inspection the format seems reasonable, but as usual most issues only come apparent when trying to use/parse it in practice.

Maybe we should clarify what we just discussed here: 
the discovery should always be possible and return all "static" discovery data for all harts ("always" means when the DM is accessible), even when the hart(s) that a certain discovery section refers to is currently not available.

Example of an interpretation that I would not like to see:
- let there be a section S that refers to hart X
- section S only contains "static" information
- if the discovery is started while hart X is available, then section S exists, otherwise S is not part of the discovery

With other words: It would be important that the debugger only needs to run the discovery once in order to get all static information, and does not need to poll the discovery information periodically to check for updated content or even new sections.

This would also lead to the question: which sections are "static", i.e. which sections can the debugger expect to not change after they were read once? Is there even any content that is ever allowed to change? Or is the whole thing a frozen ROM?

** Trigger Support: "data and mask are DXLEN-bit values".  Can we assume that the parser always knows DXLEN at this point? Or should these be varnums (of size DXLEN) as well?

** Hart Index to Hartid Map:
I think it does not become clear how important this (optional) section is. An external debugger wants "hart index -> (...) -> feature", and as all sections of the discovery are based on the hartID, the whole discovery is worthless for the debugger if this map section is not implemented. This may not be obvious at first, as the debugger does not really "care" about the hartID itself, but it just needs it as an interim-step for "hart index -> (...) -> feature".
We should make clear that this section is required **if** the discovery is to be used by an external debugger, because otherwise some system designers may think this section is only needed if the debugger actually cares about the hartID itself, and may skip that section.
Maybe put this section and its index further up in the list, so people can't miss it?

Also, even though that should be clear, maybe we should mention that unlisted harts can only be at the end of the list.

> When executed, this command will place the 4 bytes at address in the data structure into sbdata0,
in little-endian order.

Why not into data0? Is sbdata0 easier for the hardware?

> Because the CSR space is sparsely populated, it is allowed and recommended to list Access Register
CSR commands as supported even though not every CSR exists.

Do you mean: Because the CSR space is sparsely populated, it is allowed and recommended to list Access Register CSR commands as supported **for the whole CSR address space (0x0-0xfff)** even though not every CSR exists.

> This optional section describes which abstract commands (full register contents) are supported

This optional section describes which abstract commands (full register contents **of command**) are supported

allenjbaum:
I am unclear what an "unlisted" hart is, nor why those need to be at any
specific position in the list. Did you mean non-existent?
There is no need to have any discovery for non-existent harts.

There are some mentions in the spec like "at reset, XLEN is set to the
maximum supported by the implementation".
So I would expect (or mandate) that discovery reports info for all existing
harts, regardless of whether they are available, for their maximum
configurations.

There is a corner case there: Nonexistent harts are expected to be
nonexistent; they will never suddenly start existing.
Unavailable harts, on the other hand, can come and go - in theory, although
noncontiguous hart indexes can be permanently unavailable.
(the reason for that behavior is that the only way to discover harts prior
to or without this discovery process was to interrogate them one-by-on,
stopping at the first nonexistent hart.
This behavior assumed hart indexes were contiguous, but that is not always
the case,
The way around that is to make a hart index be existent but (permanently)
unavailable.
We should discuss whether discovery should be required to list permanently
unavailable harts - or even if discovery should make them appear to be
nonexistent even if reading status lists them as existent but unavailable.

Commands issued to a hart that is unavailable (even if they were discovered
to be available at some point) should either return an error (not sure that
is defined in the spec), or can be determined by looking at "all available"
status.

I can guess why sbdata is used instead of data0; this implies that the
discovery ROM is available to the system and accessible to harts just as
they are accessible to the debugger.
I am a bit unclear if this also implies that SystemBusAccess is required -
that is a concern we should discuss.

Access register: that should be for all of CSR space regardless of whether
an individual CSR exists. Attempts to access non-existent CSRs should
return error status from the hart.
(that behavior must exist now, independent of the discovery process)

timsifive:
I agree with almost everything you said, and have incorporated your
suggestions. Below are a few answers/justifications.

> ** Trigger Support: "data and mask are DXLEN-bit values". Can we assume
> that the parser always knows DXLEN at this point? Or should these be
> varnums (of size DXLEN) as well?
>
I'm adding some language that strongly encourages misa to be in the overall
structure. The debugger can get DXLEN and even more info from that. It only
adds 7 bytes, while using varnums in the trigger section adds 2 bytes per
trigger. (We could invent something different that makes the overhead just
1 byte per trigger.)

> Also, even though that should be clear, maybe we should mention that
> unlisted harts can only be at the end of the list.
>
Slightly clarified.

timsifive:
>
> I am unclear what an "unlisted" hart is, nor why those need to be at any
> specific position in the list.


Markus was referring to "unlisted" in the hart-index-to-hartid-map section,
which explicitly says:
"If this section exists but does not include every hart in the system, then
the unlisted harts must have a hartid that's equal to their hart index.
Simple systems where hartids and hart indices are equal can describe this
by including this section with a size of 0."

I would expect (or mandate) that discovery reports info for all existing
> harts, regardless of whether they are available, for their maximum
> configurations.
>

Right.


> We should discuss whether discovery should be required to list permanently
> unavailable harts - or even if discovery should make them appear to be
> nonexistent even if reading status lists them as existent but unavailable.
>

There is no mechanism to make harts appear non-existent in the data
structure. I don't know if this really matters for a debugger, since it's
immediately obvious that the hart is unavailable as soon as the debugger
selects it.
For an OS, I would imagine it would look at the structure and assume that
harts which don't e.g. have misa exposed to not be available to it.

I can guess why sbdata is used instead of data0; this implies that the
>

That was a typo. data0 is much more consistent with the other abstract
commands, and is also required to exist for all implementations. sbdata is
optional, so not a good choice here.

>

markuslauterbach:
> I'm adding some language that strongly encourages misa to be in the overall
structure. The debugger can get DXLEN and even more info from that. It only
adds 7 bytes, while using varnums in the trigger section adds 2 bytes per
trigger. (We could invent something different that makes the overhead just
1 byte per trigger.)

Should a parser be able to parse all sections in only one parser "run"? 
"Parsers don’t have to support all sections. They should skip sections of a type they do not support." -> from this I would expect that there are no dependencies between successful parsing of sections ("dependency" means I have to parse A before I can parse B). But having to parse the misa section before parsing the trigger section would be such a dependency.
If the misa section is at the very end, and other sections with DXLEN-sized values are located before the misa section, then these sections with DXLEN-sized values can only be parsed if the parsing is started multiple times.

> This optional section describes which abstract commands (full register contents
or \RdmCommand) are supported for the harts currently in the selected list.

or -> of

> This data structure is static. It may change when power the system is power cycled, but not due to a simple reset.

First "power" is a typo. 
Is there any (later) case that may make this "whole structure is static" too rectrictive? May we need/want "non-static" sections later on? Just want to put this out for discussion. On the other hand if we ever come up with a "dynamic" section type we could explicitly categorize each section type as either static or dynamic (which should be backwards compatible, or is it not?).

timsifive:
>
> Should a parser be able to parse all sections in only one parser "run"?
>
That is at a minimum nice-to-have. I took another look at the trigger
section, and it's actually easy to encode the length without using any
bytes so I'm making that change.

> Is there any (later) case that may make this "whole structure is static"
> too rectrictive? May we need/want "non-static" sections later on? Just want
> to put this out for discussion. On the other hand if we ever come up with a
> "dynamic" section type we could explicitly categorize each section type as
> either static or dynamic (which should be backwards compatible, or is it
> not?).
>
> That sounds possible, and compatible.
I'm still a bit concerned about declaring it static like this. I imagine an
IC where e.g. endianness is configurable as an input, and a dip switch on a
board that lets it be set. It seems potentially hard/counterintuitive to
force that dip switch to be ignored on reset but not during a power cycle.

>

markuslauterbach:
The "Hart Select" and "Trigger Support" sections seem both to be based on the same principle: the parser needs to have a "full list" of harts/triggers respectively, and then needs to select/add/remove harts/triggers while parsing. But how does the parser get these "full lists" of harts and triggers to begin with?

Full hart list:
- Parsers of an external debugger can use 'hartsel' and 'anynonexistent' of dmstatus and dmcontrol to get that full list. However this requires a separate preceding discovery process before being able to run the feature discovery, and also this is only possible for an external debugger and not for feature discovery by SW.

Full trigger list:
- Here I only see the algorithm described in the debug spec that uses 'tselect', 'tinfo' and 'tdata1'. However this algorithm is intrusive as it accesses CSRs and requires all harts to be available.

Is there something I do not see?

allenjbaum:
I just noticed that there is no mention of autoincrementing the address when reading sdata1 for feature .
That means you can't read a stream of discovery data; you need to continually increment the abscmd address by 4, write it into abscmd and read sdata1, instead of simply having a stream of reads.

I can't see any reason not to autoincrement (besides the cost of an incrementer, since you can always default to explicitly writing the explicitly incremented address).

Question: If a jump or include over/underflows the 16bit address (or if autoincrementing): there is a statement that it doesn't wrap, but no comment about what does happen. In the context of always explicitly writing the address for every access, this behavior is irrelevant: there is a 16bit field in abscmd which is the address, and the parser could do anything it wants.

So.what is the definition of what does happen? Just an error response which you can query at the end? ( with a forced return of 0 data?) Is the expectation that this is entirely enforced by the parser (since autoincrementing isn't defined)?

changab:
> The data structure is at most 16MB (given the abstract_command register
> format- we may want to re-think the command number to leave room for
> expansion).
> This 16MB space could be
>  - accessible only to cores via normal loads and store (i.e. the 16MB
> structure is an MMIO address)
>  - accessible only to cores via CSRs (e.g. hart CSRs that replicate
> something like the abstract_command and Data0 Debug regs, )
>  - is only a dedicated ROM that is part of the DM, or
>  - accessible to both: directly to the DM, and to cores either via CSRs or
> via Ld/St
> If the first,  then
>    --the DM needs to know where in the larger (32b+) address this 16MB
> section starts; it almost certain can't be at base=0, and either
>      --- System Bus Access is implemented or
>      --- cores must be halted and code downloaded to the program buffer
> must be executed
> If the second
>    -- cores must be halted and code downloaded to the program buffer must
> be executed
> If the third, then we may need a separate implementation for other types of
> discovery that cores want to do.
The third one is not the perfect solution because the feature discovery is not only limited to DM as  @allenjbaum mentioned. Other non-DM RISC-V core features are better to be able to discover in the same way using FDS. 

Regards to the first one, topmost 16MB on system bus bring the conflicts of multicores and multiprocessors. Those cores may have different FDS but all overlapped at the topmost 16MB region. Is my understanding correct?

CSR solution looks to be the best solution although we still have to discuss the detail implementation. If my understanding is correct on DM program buffer, DM could run a piece of code on the desired core to retrieve FDS from core CSR _(but where is the program buffer located?)_. We can also introduce a new CSR (zero when reset or power on) which indicates the base address to FDS on memory. The firmware could gather FDSs from each core and put it on a protected memory region. Firmware connects all FDSs using Jump section type and reports the base address to the starting FDS in CSR. Afterward, either debugger or other firmware drivers/applications can get the full FDSs on the system through this CSR.
Firmware should build up the entire FDSs as earlier as possible when system is power on.
One drawback of reporting a block of FDSs instead of using the mechanism like x86 CPUID is any application wants to know the specific feature availability has to parse the whole FDSs.

allenjbaum:
An issue with the MMIO approach is that I don't think we can fix an MMIO
address, so in any case we need a CSR.
That CSR either directly accesses the structure or points to it in MMIO
space.
If it's just a pointer, it could be hardwired copy in the DM; otherwise it
must be shared, and may require that a core be alive before it can be
accessed.


An issue with the CSR approach is that sharing becomes interesting.
Generally, we expect that CSRs are tightly connected to harts (each hart
has its own set), but in most cases here, all the harts will share the same
discovery data (in a heterogeneous implementation, there are likely only 1
set per type of core).
There is some precedent for CSRs to share resources (I'm thinking mtime) so
this CSR may be just another example of that.

That points to another general issue: we have to support multiple discovery
structures.
If each implementation has its own DM, then each might have its own
hardwired pointer to its structure.
I don't see why we have to parse an entire structure before we get to the
one that applies to a specific hart, instead of just scanning headers to
find the right one, with a linked list of headers (very much like what the
debug interface does to point to the next DM)

timsifive:
There will probably be a task group for this.

Issue #451, opened 2019-05-10
Linking to version field is broken


pdonahue-ventana:
The DM chapter discusses the "version" field a few times:
- "**version** in dmstatus is readable."
- "Read dmstatus, which contains **version**."
- "Its address will not change in the future, because it contains **version**."
- "For forward compatibility, **version** will always be readable when bit 1 (ndmreset) is 0 and bit 0
(dmactive) is 1."
- "Specifically, **version** may not return correct data."

All of those references link to dtmcs.version in the DTM chapter when you click on the word version.  Clearly most of them and I assume all of them should instead link to dmstatus.

pdonahue-ventana:
I think that the problem is that both dm_registers.tex.inc and jtag_registers.tex.inc have definitions for \Fversion and the last definition apparently wins.  As things are done now, field names are not namespaced by register so all field names must be globally unique (per document, not globally across all RISC-V documents).

I see a similar problem on the sentence "To change the privilege level, the debugger can write prv in
dcsr" which explicitly says that it's prv in dcsr but the prv link takes me to prv in priv, not dcsr.

PR #457, opened 2019-05-21
Fully qualify register/field macro names.

timsifive:
This solves the problem we have with the same field existing in multiple
registers, as reported in #451.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -32,8 +32,8 @@
         \item Implement abstract access to all registers that are visible to
             software running on the hart including all the registers that are
             present on the hart and listed in Table~\ref{tab:regno}.
-        \item Implement abstract access to at least all GPRs, \Rdcsr, and
-            \Rdpc, and advertise the implementation as conforming to the
+        \item Implement abstract access to at least all GPRs, \RcsrDcsr, and
+            \RcsrDpc, and advertise the implementation as conforming to the
             ``Minimal RISC-V Debug Specification \versionnum'', instead of the
             ``RISC-V Debug Specification \versionnum''.
     \end{steps}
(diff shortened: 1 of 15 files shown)


debug_module.tex
```diff
@@ -54,55 +54,55 @@ \section{Debug Module Interface (DMI)} \label{dmi}
...
 These bits must be set regardless of the cause of the reset.
 The {\tt havereset} bits for the selected harts
-can be cleared by writing 1 to \Fackhavereset in \Rdmcontrol.
+can be cleared by writing 1 to \FdmDmcontrolAckhavereset in \RdmDmcontrol.
 The {\tt havereset} bits may or may not be cleared
-when \Fdmactive is low.
+when \FdmDmcontrolDmactive is low.
 
-When a hart comes out of reset and \Fhaltreq or \Fresethaltreq are set, the
+When a hart comes out of reset and \FdmDmcontrolHaltreq or \Fresethaltreq are set, the
```
ernie-sifive: How did you decide which registers and fields were changed to the fully qualified name and which weren't?  `\Fhaltreq` was changed but `\Fresethaltreq` was not.  And `\Fhartsel` vs. `\FdmDmcontrolHasel`

jscheid-ventana:
Is the intent to namespace the field names, or just fix the document linking?  If the former, I suggest using the <registername>.<fieldname> convention found in the priv spec.

debug_module.tex
timsifive: I changed every macro that was autogenerated.

resethaltreq isn't a bit in a register. These bits exists somewhere in the DM, and are accessed by dmcontrol.setresethaltreq and dmcontrol.clrresethaltreq.

hartsel is similar. It's the combination of dmcontrol.hartsello and dmcontrol.hartselhi.

timsifive:
> Is the intent to namespace the field names, or just fix the document linking? If the former, I suggest using the . convention found in the priv spec.

For now just to fix the linking. If we want to refer to each field as register.field we can make that change easily in registers.py.

PR #460, opened 2019-05-29
Clarify mcontrol.size.

### LINKED ISSUE #455: mcontrol.size
In 0.14.0-DRAFT, the sizehi field of mcontrol is defined as a R/W field and described as "This field only exists if XLEN is greater than 32.  In that case it extends size. If it does not exist then hardware operates as if the field contains 0."

I have a few questions:

1. If sizehi does not exist (because XLEN=32), does that mean that this field is no longer R/W?  If so, could it just be WARL instead of not existing?  If not, does an RV32 implementation need to implement two state bits for this useless field simply to support R/W?
2. To further complicate matters: Section 5.2 says "In this section XLEN means MXLEN when in M-mode..."  If software on an RV64 implementation changes the MXL field in misa from 2 to 1 then XLEN goes from 64 to 32.  By the current definition, that means that sizehi previously existed but no longer exists.  What are the implications of this?
3. If an implementation has XLEN=32 but supports execution of instructions that are 48-bits or more, shouldn't sizehi be required or at least allowed?

I think that the sizehi description should only say:
"This field contains the 2 high bits of size.  See sizelo for more details."

Then either sizelo/sizehi could be WARL or the sizelo description could add: "Implementations with XLEN<=64 and that do not support instructions of length > 96 may hardwire size[3]=0.  Implementations with XLEN=32 and that do not support instructions of length > 32 may hardwire size[2]=0."

timsifive:
My attempt to fix #455.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -220,10 +220,10 @@
             size.
         </field>
         <field name="0" bits="XLEN-12:23" access="R" reset="0" />
-        <field name="sizehi" bits="22:21" access="R/W" reset="0">
-            This field only exists if XLEN is greater than 32. In that case it
-            extends \FcsrMcontrolSizelo. If it does not exist then hardware operates as if
-            the field contains 0.
+        <field name="sizehi" bits="22:21" access="WARL" reset="0">
+            This field contains the 2 low bits of the access size. The low bits
+            come from \FcsrMcontrolSizelo. See \FcsrMcontrolSizelo for how this
+            is used.
         </field>
         <field name="hit" bits="20" access="R/W" reset="0">
             If this optional bit is implemented, the hardware sets it when this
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -220,10 +220,10 @@
             size.
         </field>
         <field name="0" bits="XLEN-12:23" access="R" reset="0" />
-        <field name="sizehi" bits="22:21" access="R/W" reset="0">
-            This field only exists if XLEN is greater than 32. In that case it
-            extends \FcsrMcontrolSizelo. If it does not exist then hardware operates as if
-            the field contains 0.
+        <field name="sizehi" bits="22:21" access="WARL" reset="0">
+            This field contains the 2 low bits of the access size. The low bits
```
ernie-sifive: Shouldn't this say "This field contains the 2 high bits of the access size..."?

timsifive: Yes. Fixed. Thanks for catching that.

PR #463, opened 2019-06-04
Numerous tweaks, responding to Marc Gauthier

timsifive:
See https://lists.riscv.org/g/tech-debug/message/755


xml/abstract_commands.xml
```diff
@@ -26,14 +26,25 @@
...
         Debug Modules may optionally support accessing other registers,
         or accessing registers when the hart is running.
-        Each individual register (aside from GPRs) may be supported differently
-        across read, write, and halt status.
+        It is recommended that if one register in a group is accessible, then
+        all registers in that group are accessible, but each individual
+        register (aside from GPRs) may be supported differently across read,
+        write, and halt status.
+
+        Registers may not be accessible if they wouldn't be accessible by M
```
pdonahue-ventana:
```suggestion
        Registers might not be accessible if they wouldn't be accessible by M
```
"may not" is ambiguous (sometimes meaning "shall not" and sometimes meaning "might not").

debug_module.tex
```diff
@@ -325,9 +328,9 @@ \section{Abstract Commands} \label{abstractcommands}
 \FdmDmcontrolResumereq, \FdmDmcontrolAckhavereset, \FdmDmcontrolSetresethaltreq, or \FdmDmcontrolClrresethaltreq.
 
 If an abstract command does not complete in the expected time and appears to be
-hung, the following procedure can be attempted to abort the command: First the
-debugger resets the hart (using \FdmDmcontrolHartreset or \FdmDmcontrolNdmreset), and then it resets
-the Debug Module (using \FdmDmcontrolDmactive).
+hung, the debugger can try to reset the hart (using \FdmDmcontrolHartreset or
+\FdmDmcontrolNdmreset). If that doesn't clear \FdmAbstractcsBusy, then it
+can try resetting the Debug Module (using \FdmDmcontrolDmactive).
```
ernie-sifive: Wouldn't resetting the debug module always recover from this hang condition?  Not sure that "can try resetting" is the best terminology ... maybe "...then it must reset the Debug Module using..."

introduction.tex
```diff
@@ -84,9 +84,11 @@ \subsection{Register Definition Format}
         \hline
         R/W1C & Read/Write. For each bit in the field, writing 1 clears that bit. Writing 0 has no effect. \\
         \hline
-        W & Write-only. When read this field returns 0. \\
+        WARZ & Write any, read zero. A debugger may write any value. When read
+        this field returns 0. \\
         \hline
-        W1 & Write-only. Only writing 1 has an effect. \\
+        W1 & Write-only. Only writing 1 has an effect. When read the returned
+        value should be 0. \\
```
ernie-sifive: Use the same text as the WARZ case?  "When read this field returns 0"

xml/dm_registers.xml
```diff
@@ -396,7 +398,8 @@
         executed.
 
         Writing this register while an abstract command is executing causes
-        \FdmAbstractcsCmderr to be set to 1 (busy) if it is 0.
+        \FdmAbstractcsCmderr to become to 1 (busy) if it is 0, when busy
+        becomes 0.
```
ernie-sifive: This wording seems a little awkward.  I think you mean that cmderr is set to 1 because of this write, but only after the currently-executing command finishes?  Why the delay?  I guess if the current command exits with cmderr=2, then it is changed to cmderr=1 (i.e. 1 is a higher priority in case of multiple errors)?

debug_module.tex
timsifive: I could imagine some bug/condition in the hart itself causing an abstract command to hang, and resetting the DM does not reset the hart so wouldn't necessarily resolve that.

introduction.tex
timsifive: That would not be backwards compatible. Existing implementations might already be returning 1 in that case. I don't think it really matters what they return, but it would be neater if they always return 0.

xml/dm_registers.xml
timsifive: The idea is to allow implementations to have cmderr contain arbitrary values while the command is executing. So we don't define what its value should be until busy becomes 0. Sounds like this isn't clear. What would be a better way to phrase this concisely?

ernie-sifive: How about something like "Writing to this register while an abstract command is executing with cmderr=0 causes cmderr to be set to 1 (busy) at or before the conclusion of that command"?

marceg: Here's wording that had been suggested earlier:
"Writing this register while an abstract command is executing causes cmderr to become 1 (busy) once the command completes (busy becomes 0), if cmderr was 0 when the command started."

This phrasing avoids making any assumptions about the value of cmderr while the command is executing, as Tim requested.  Is it clear?  "executing with cmderr=0" makes such an assumption.

timsifive: "if cmderr was 0 when the command started" is not necessary. If cmderr wasn't 0, then the command would not have started. So how about

"Writing this register while an abstract command is executing causes cmderr to become 1 (busy) once the command completes (busy becomes 0)."

ernie-sifive: That sounds good to me.

marceg: Right.  Yes, sounds good.
Couple of observations.

1. This design means there is hidden state (whether the register was written while a command is executing, is not visible until the command completes).  Seems okay, if done knowingly.
2.  The same sentence exists for abstractcs, command, and abstractauto.  Will it change for the other two.

timsifive: I'm fine with hidden state. We already don't know what is being executed, or where things are in the process.

Thanks for pointing out the other sentences. I've updated them also.

PR #467, opened 2019-06-12
A few clarifications for triggers and abstract commands

marceg:
From commit logs:

    In Section 3.7 Abstract Commands:
    Clarify cmderr=2 (not supported) if not supported at all,  vs.
    cmderr=4 (halt/resume) if supported but not in the current hart state.

    Then in subsections for each abstract command, clarify that their
    sequence of operation is conditional on their being supported as
    described above.  This makes it clear for example that cmderr=2 is
    checked before other error conditions.

    "may not" can mean "must not", reword to avoid ambiguity.

    Fix section references in cmdtype table.
_____
    Triggers can both halt and cause breakpoint exceptions.
_____
    Add missing file to generate the draft spec.

### DIFF
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -3,7 +3,9 @@
     <register name="Access Register">
         \begin{steps}{This command gives the debugger access to CPU registers
         and allows it to execute the Program Buffer.
-        It performs the following sequence of operations:}
+        If the command and all requested options are supported in the current state
+	(see Section~\ref{abstractcommands}),
+        it performs the following sequence of operations:}
         \item If \FacAccessregisterWrite is clear and \FacAccessregisterTransfer is set, then copy data from
             the register specified by \FacAccessregisterRegno into the {\tt arg0} region of
             {\tt data}, and perform any side effects that occur when this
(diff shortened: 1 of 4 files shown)


xml/abstract_commands.xml
```diff
@@ -55,6 +57,9 @@
         This command modifies {\tt arg0} only when a register is read. The
         other {\tt data} registers are not changed.
 
+	For the Access Register Command, the \RdmCommand register (see Section~\ref{command})
```
timsifive: This seems unnecessary. We don't have a statement like that for any other register/command definition. Why do we need one here?

marceg: This provides a reference to the command register whose content is being further specified.  While this may be obvious to those familiar with the spec, it is not for new readers.  In my first readings of the spec, I missed seeing appropriate links and it takes extra reading to be confident to have correctly connected all the dots.  You're right that a similar statement is needed for other commands, to be consistent.

timsifive: If we want this, let's do it right and do it everywhere. Can you remove this line from this PR? I've opened #471 to track the broader issue. (Please feel free to remind me about that one if I don't follow up.)

timsifive:
Feel free to reopen if you still think this is important.

PR #473, opened 2019-07-11
Define what we mean by virtual address.

timsifive:


pdonahue-ventana:
The term "virtual address" is used extensively in the privileged architecture.  The debug extension builds on and presumes familiarity with the unprivileged and privileged architectures, including any terms used by them.  If the term "virtual address" is unclear in the base architecture, I think that an issue and/or pull request should be created for the base architecture rather than defining such a globally-relevant term just for debug.

I do agree that there should be a sentence that the system bus uses physical addresses (which is similarly a term used in the base architecture).

timsifive:
The reason to add this language was to make it at least specified that DPC contains a physical address when address translation is turned off. There are probably other registers where that applies, so I put it here. Is there a better way to achieve that?

marceg:
Perhaps if the privileged specs defines PC and load/store addresses as 
always being virtual addresses, with virtual addresses mapping one to 
one to physical addresses when translation is not enabled, all other 
documentation is simplified, able to simply refer to these as being 
virtual addresses.

This approach also maps more cleanly in RV32I, where without 
translation, a 32-bit virtual address maps to a 34-bit physical 
address.  GPRs are not able to hold a full physical address, and it is 
not quite accurate to simply say that a GPR holds a physical address.

-M


Tim Newsome wrote on 2019-07-12 12:16:

pdonahue-ventana:
I agree that the privileged architecture should clean this up if it's
unclear.  This kind of thing applies just as much to the non-debug epc as
to debug's dpc.

I don't think it's unclear, though.  The privileged architecture says:

> When MODE=Bare, supervisor virtual addresses are equal to supervisor
> physical addresses

and:

> When a trap is taken into M-mode, mepc is written with the virtual address
> of the instruction that was interrupted or that encountered the exception.

and the same thing for sepc as mepc.

Perhaps the debug spec should just say the exact same thing about depc as
mepc/sepc and not define virtual address, implicitly keeping the same
definition as the base architecture.


Thanks,

-Paul

allenjbaum:
I had forgotten that sentence in the priv spec. So a very careful reading of the priv mode spec would imply that. On the other hand, I searched for every use of the word "virtual" in the priv sepc, and still missed that comment it, so I would not characterize this as clear. Translucent, perhaps.
 In any case, there is nothing to lose, and a lot to be gained, in saying it again, especially if it is only mentioned only obliquely in an entirely separate spec. Specs should be as self contained as possible. In this case, it would difficult to even point to that "definition" of virtual address in the priv mode spec, since its buried in the description of other things.

PR #474, opened 2019-07-15
Mention the scontext reg number isn't conventional

### LINKED ISSUE #410: Clarifications are needed for the fields of textra CSR
Two clarifications are needed for the fields of **textra** CSR.

1. textra.**svalue** and textra.**sselect** can be written in **Debug/M/S** modes; textra.**mvalue** and textra.**mselect** can be written in **Debug/M** modes.
2. textra.mvalue and mcontex comparison, when turned on, matches in **M/S/U** modes. textra.svalue and scontext comparison, when turned on, matches in **S/U** modes.

If my proposed clarifications are wrong, please let me know. Thanks.

Chuanhua

timsifive:
It really should be in the S-mode section of CSRs, but that would be an
incompatible change. Fixes #410.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -165,6 +165,10 @@
     <register name="Supervisor Context" short="scontext" address="0x7aa">
         This optional register is only writable in S mode, M mode and Debug Mode.
 
+        Note that the register number does not follow the read and write
+        accessibility of the CSRs according to privilege level as defined in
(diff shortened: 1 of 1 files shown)


timsifive:
Agreed, but this change doesn't actually change anything. It simply points out the problem that was introduced (and ratified) in 0.13. It's worth discussing if/how to properly fix it on the list.

PR #475, opened 2019-07-15
Explicitly document confstrptr[1-3].

timsifive:


allenjbaum:
This removes my objections by explicitly describing all registers making up the config string.

PR #485, opened 2019-09-03
Git ignore intermediate and output files

imphil:
Add a gitignore file to keep the output of `git status` a bit cleaner. There's still the checked-in PDF that differs all the time, but that needs to be solved by CI or another way which doesn't include checking in this file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,27 @@
+# LaTeX intermediate files
+*.aux
+*.idx
+*.ilg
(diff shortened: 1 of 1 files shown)


timsifive:
The way I'm handling the PDF is that I never submit it, and once a month (before the meeting) I manually rebuild it and push it. So it's never more than a month out of date, and we don't clutter the history too much.

PR #493, opened 2019-09-16
DM reset must also reset all the DM's harts.

timsifive:
Alternative to #488.


timsifive:
@darius-bluespec, does this work?

darius-bluespec:
Looks good to me, thanks.

PR #494, opened 2019-09-19
Rewrite/clarify DM Reset Control

### LINKED ISSUE #480: havereset
> Individual harts (or several at once) can be reset by selecting them, setting and then clearing
hartreset.

> When harts have been reset, they must set a sticky havereset state bit.

Does havereset get set on the positive edge of hartreset or on the negative edge?

timsifive:
Hopefully fixes #480.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -56,25 +56,46 @@ debug devices, other cores, additional DMs, etc. If there are additional DMs
 on this DMI, the base address of the next DM in the DMI address space is given
 in \RdmNextdm.
 
-
 The Debug Module is controlled via register accesses to its DMI address space.
 
 \section{Reset Control} \label{reset}
 
-The Debug Module controls a global reset signal, \FdmDmcontrolNdmreset
-(non-debug module reset),
-which can reset, or hold in reset, every component in the platform,
-except for the Debug Module and Debug Transport Modules.
-Exactly what is affected by this reset is implementation dependent, as long as
-it is possible to debug programs from the first instruction executed.
+There are two methods that allow a debugger to reset harts.
+\FdmDmcontrolNdmreset resets all the harts in the system, as well as all other
+parts of the system except for the Debug Module and Debug Transport Modules.
+Exactly what is affected by this reset is implementation dependent, but it
+must be possible to debug programs from the first instruction executed.
+\FdmDmcontrolHartreset resets all the currently selected harts. In this case an
+implementation may reset more harts than just the ones that are selected. The
+debugger can discover which other harts are reset (if any) by selecting them
+and checking \FdmDmstatusAnyhavereset and \FdmDmstatusAllhavereset.
+
+To perform either of these resets, the debugger first asserts the bit, and then
+clears it. The actual reset may start as soon as the bit is asserted, but is allowed
+to start a long time after the bit is deasserted. While the reset is
+on-going, harts are either in the running state, indicating it's possible to
+perform some abstract commands during this time, or in the unavailable state,
+indicating it's not possible to perform any abstract commands during this time.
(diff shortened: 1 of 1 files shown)


timsifive:
@darius-bluespec, this isn't perfect but can you take a look anyway?

timsifive:
Oops, meant to tag @pdonahue-ventana instead.

debug_module.tex
```diff
@@ -56,25 +56,46 @@ \section{Debug Module Interface (DMI)} \label{dmi}
...
+Exactly what is affected by this reset is implementation dependent, but it
+must be possible to debug programs from the first instruction executed.
+\FdmDmcontrolHartreset resets all the currently selected harts. In this case an
+implementation may reset more harts than just the ones that are selected. The
+debugger can discover which other harts are reset (if any) by selecting them
+and checking \FdmDmstatusAnyhavereset and \FdmDmstatusAllhavereset.
+
+To perform either of these resets, the debugger first asserts the bit, and then
+clears it. The actual reset may start as soon as the bit is asserted, but is allowed
+to start a long time after the bit is deasserted. While the reset is
```
pdonahue-ventana: How about "but may start an arbitrarily long time after the bit is deasserted"?  It sounds a little more formal to me.

debug_module.tex
```diff
@@ -56,25 +56,46 @@ \section{Debug Module Interface (DMI)} \label{dmi}
...
+on-going, harts are either in the running state, indicating it's possible to
+perform some abstract commands during this time, or in the unavailable state,
+indicating it's not possible to perform any abstract commands during this time.
+Once a hart's reset is complete, {\tt havereset} becomes set.  When a hart comes out
+of reset and \FdmDmcontrolHaltreq or \Fresethaltreq are set, the hart will
+immediately enter Debug Mode. Otherwise it will execute normally.
```
pdonahue-ventana:
```suggestion
immediately enter Debug Mode and be in the halted state. Otherwise it will execute normally and be in the running state.
```

debug_module.tex
pdonahue-ventana: That language also happens to mirror language that was removed below that says that the reset itself can be arbitrarily long.  I think it would be useful to say something to that effect here so it's clear that entering and exiting reset can each take a while, not just the former.

debug_module.tex
```diff
@@ -56,25 +56,46 @@ \section{Debug Module Interface (DMI)} \label{dmi}
...
 
-The Debug Module controls a global reset signal, \FdmDmcontrolNdmreset
-(non-debug module reset),
-which can reset, or hold in reset, every component in the platform,
-except for the Debug Module and Debug Transport Modules.
-Exactly what is affected by this reset is implementation dependent, as long as
-it is possible to debug programs from the first instruction executed.
+There are two methods that allow a debugger to reset harts.
+\FdmDmcontrolNdmreset resets all the harts in the system, as well as all other
+parts of the system except for the Debug Module and Debug Transport Modules.
```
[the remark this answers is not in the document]
timsifive: Good points. I've incorporated that.

debug_module.tex
timsifive: Added. Thanks.

debug_module.tex
timsifive: This implies that being in Debug Mode doesn't automatically mean being halted, but being in Debug Mode is synonymous with being halted. If that's not clear, we could use something like:
immediately enter Debug Mode (be in the halted state). Otherwise it will execute normally (be in the running state).

What do you think?

pdonahue-ventana: That sounds good.  It's clear to me that they're synonymous but I just thought that because the rest of this section talks about the various states, they should be mentioned in this sentence, too.

PR #496, opened 2019-10-03
Document forward progress guarantees in Debug Mode.

timsifive:
Alternative to #481.


timsifive:
The real subtle thing I wanted to ensure is that there's forward progress when the code in the Program Buffer jumps to some other area of memory, which elsewhere we at least try to keep open as an option.

Good points on the other two. I'm updating the PR.

PR #500, opened 2019-11-20
Time may pass before dmactive becomes high.

timsifive:
This addresses a concern of how the debugger can know when hardware is
ready to be debugged, discussed on the mailing list. subject:"Email
discussion: how can a debugger know when a target is ready for it?"

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -242,9 +242,9 @@
             correct data.
 
             1: The module functions normally. After writing 1, the debugger should
-            read \RdmDmcontrol to confirm that this bit is now high. Hardware may
-            take a long time to set this bit if it's not currently ready for
-            debugging to take place.
+            poll \RdmDmcontrol until \FdmDmcontrolDmactive is high. Hardware may
+            take an arbitrarily long time to initialize and will indicate completion
+            by setting dmactive to 1.
 
             No other mechanism should exist that may result in resetting the
             Debug Module after power up.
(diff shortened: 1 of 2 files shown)


xml/dm_registers.xml
```diff
@@ -241,7 +241,10 @@
             to the module may fail. Specifically, \FdmDmstatusVersion may not return
             correct data.
 
-            1: The module functions normally.
+            1: The module functions normally. After writing 1, the debugger should
+            read \RdmDmcontrol to confirm that this bit is now high. Hardware may
+            take a long time to set this bit if it's not currently ready for
+            debugging to take place.
```
pdonahue-ventana: This isn't really crucial but I find this sentence to be a little less formal than specs normally are (like "long time" and what does it mean for "debugging to take place").  How about something like:
"Hardware may postpone setting this bit until it is in a state that allows further debugger accesses."
or combining with a rewrite of the previous sentence:
"After this bit is written with 1, the bit will continue to be seen by reads as being 0 until the hardware is in the defined post-reset state.  Therefore, after writing 1 the debugger should poll \RdmDmcontrol until it observes that it is set."
(because the debugger shouldn't just read it once, it should poll until set)

ernie-sifive: Or "After writing 1, the debugger should poll Dmcontrol until dmactive is high.  Hardware may take an arbitrarily long time to initialize and will indicate completion by setting dmactive to 1."

pdonahue-ventana: Perfect.

PR #501, opened 2019-11-20
Clarify action=1 (enter Debug Mode) with dmode=0

timsifive:
Otherwise you might end up in a situation where regular software can
control when to enter debug mode.

### DIFF
--- a/trigger.tex
+++ b/trigger.tex
@@ -63,8 +63,9 @@ Value & Description \\
     be executed to preserve the program flow.
 
     This action is only legal when the trigger's \FcsrTdataOneDmode is 1.
-    Hardware should ignore triggers which have this action configured without
-    having \FcsrTdataOneDmode set. \\
+    Since the {\tt tdata} registers are WARL, hardware should clear the action
+    field whenever the action field is 1, \FcsrTdataOneDmode is cleared, and the
+    new value of the action field would also be 1. \\
 \hline
 2 -- 5 & Reserved for use by the trace specification. \\
 \hline
--- a/trigger.tex
+++ b/trigger.tex
@@ -58,10 +58,13 @@ Value & Description \\
     \Rmepc must contain the virtual address of the next instruction that must
     be executed to preserve the program flow. \\
 \hline
-1 & Enter Debug Mode. (Only legal when the trigger's \FcsrTdataOneDmode
-    is 1.)
+1 & Enter Debug Mode.
     \RcsrDpc must contain the virtual address of the next instruction that must
-    be executed to preserve the program flow. \\
+    be executed to preserve the program flow.
+
+    This action is only legal when the trigger's \FcsrTdataOneDmode is 1.
+    Hardware should ignore triggers which have this action configured without
+    having \FcsrTdataOneDmode set. \\
 \hline
 2 -- 5 & Reserved for use by the trace specification. \\
 \hline
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -61,6 +61,8 @@
             selected \RcsrTselect.  Writes from other modes are ignored.
 
             This bit is only writable from Debug Mode.
+            When clearing this bit, the debugger should also clear the action field
+            (whose location depends on \FcsrTdataOneType).
         </field>
         <field name="data" bits="XLEN-6:0" access="WARL" reset="Preset">
             If \FcsrTdataOneType is 0, then this field is hard-wired to 0.


trigger.tex
```diff
@@ -58,10 +58,13 @@ \section{Actions}
...
-1 & Enter Debug Mode. (Only legal when the trigger's \FcsrTdataOneDmode
-    is 1.)
+1 & Enter Debug Mode.
     \RcsrDpc must contain the virtual address of the next instruction that must
-    be executed to preserve the program flow. \\
+    be executed to preserve the program flow.
+
+    This action is only legal when the trigger's \FcsrTdataOneDmode is 1.
+    Hardware should ignore triggers which have this action configured without
+    having \FcsrTdataOneDmode set. \\
```
ernie-sifive: Would it be better to say that hardware must not accept a write to mcontrol with dmode=0 and action=1.  rocket-chip forces action to 0 if dmode is being written with 0.

allenjbaum:
There are 4 possible scenarios (in what I think is the most complex to the
simplest HW implementation order):
 - ignore the write and return an error code
 - clear the action bit
 - allow the write, but inhibit the trigger
 - clear the output of the action bit, but not the action bit state

Note that because of the semantics of CSRRC and CSRRS, we may not be
actually writing the action bit if we just clear dmode, and not writing to
dmode bit if we are just setting action, so it is not the act of writing
that is important, but the final value of the bit

I'd prefer one of the last two, but could probably live with the rocket
implementation.

allenjbaum:
Based on the above, the spec change is #3, but rocket implements #2
#1 doesn't work if there is no external debugger; you can still ignore the
write, but can't return an error code.
I'm thinking that most people are implementing the rocket implementation.

Putting on my compliance hat: That makes the combination of (dmode,action)
be a WARL field with the lone illegal setting of 01 being forced to 00. I
*think* the proposed compliance framework WARL semantics can deal with
that; there is a conditional flag that can be put on a field to select
between two formats (e.g. RW or ROZero). Otherwise, they are non-contiguous
bits and we can't treat them as a field, and it's nasty
#3 or #4 is  easiest for compliance purposes.

timsifive:
A debugger (or software) should never write an illegal value. Hardware may choose to protect itself in case it does. I have no strong feelings if that happens by changing the value as it's written to something legal, or by ignoring the illegal trigger. Should I just mention both, or just say something vague like "hardware should ensure not to enter Debug Mode through a trigger where dmode=0?"

pdonahue-ventana:
> There are 4 possible scenarios (in what I think is the most complex to the simplest HW implementation order):
> - ignore the write and return an error code

How does an error code get returned on a CSR write?

> - clear the action bit

Note that the action field is 4 or 6 bits (depending on the type of trigger).  Would this mandate a specific WARL behavior that action=1 gets mapped to action=0 in this case (rather than allowing hardware to set action to some other value)?  It makes sense to me to map it to 0 rather than something related to trace or a currently reserved encoding but I think that the fact that action is more than one bit must be considered.

> - allow the write, but inhibit the trigger

Does this mean that the state is allowed to contain action=1 and dmode=0 but that while this situation exists the type field is treated by the logic (for all purposes other than CSR read) as being 0 regardless of the actual value?

> - clear the output of the action bit, but not the action bit state

Is this a mixture of the two above cases where the state is allowed to contain the illegal state but that while this situation exists the action that the logic uses (for all purposes other than CSR reads) is 0?


After reading this, I think that another alternative is to say that `type!=0 && action==1 && dmode==0` is illegal and let implementations do whatever WARL behavior they want (clear the type field, clear the action field, set action to some arbitrary legal value other than 1, keep dmode set, etc.).  I'm not a particularly big fan of this but it does seem more RISC-V-like.

ernie-sifive:
I think the hardware should never be allowed to have an illegal value in a register.  It should alter the register value in some way to make the value legal.  In this case, attempting to change dmode from 1 to 0 could either leave dmode at 1 or change action to 0.

pdonahue-ventana:
> I think the hardware should never be allowed to have an illegal value in a register. It should alter the register value in some way to make the value legal. In this case, attempting to change dmode from 1 to 0 could either leave dmode at 1 or change action to 0.

Or I think it could set type==0 to disable the trigger so dmode and action are irrelevant.

I agree that having illegal values in registers is potentially confusing and goes against the spirit (and probably also the letter) of WARL.  I'm OK with any of the 3 proposals (fixing up dmode or action or type).

allenjbaum:
This is all my fault for bringing it up to begin with. We found a bug in
this area, but looking at the spec, we had no idea what the right way to
fix it was.

The effect you want to disallow is halting on a trigger match when no
external debugger is connected (which is just a hang condition).
There is no wording preventing that setting exactly (it's an ambiguity),
but making the result implementation dependent, but specifying
 - what actions are disallowed and
 - listing which (combinations) of the above options are allowed
should achieve the desired result.

It might be easier to say that any series of actions that attempt to set
dmode=0, action=1 will cause implementation dependent behavior, including?
limited to? (clearing action, forcing break instead of halt, inhibiting the
write, inhibiting the trigger), whose result not halt based on this trigger.

With my well-worn compliance hat: it's yet another option that may need to
be discovered

allenjbaum:
- I thought I followed up with a subsequent email that the error code only
makes sense for an external debugger
 - yea, I know action a 4bit (or 6bit!)  field, but only 1 bit is RW in our
implementation because that's all the spec defines
    so in general, map the halt action to break action, which is just
clearing the LSB
- correct, #3 maintains state but  disables the trigger; no match can occur
- correct, #4 maintains state, but converts a trigger match halt into a
break

pdonahue-ventana:
> - I thought I followed up with a subsequent email that the error code only makes sense for an external debugger

If the external debugger does the CSR write via the program buffer then I think that there's no way to return an error code (unless there's some additional error bit introduced into a CSR and then the debugger would have to sample that after every write but that's ugly and slow).

> - yea, I know action a 4bit (or 6bit!) field, but only 1 bit is RW in our implementation because that's all the spec defines so in general, map the halt action to break action, which is just clearing the LSB

I'm just saying that if the spec imposes a specific behavior now then we should consider the future and not make a choice that somehow prevents flexibility in defining new actions in the future.  We shouldn't paint ourselves into a corner (not that I think that the proposal to clear action does that but I am inviting others to make sure that I'm not wrong).

timsifive:
Paul makes an excellent point that these registers are all WARL, so as Ernie says, the registers should never contain an illegal value. I'll update the PR.

PR #504, opened 2019-11-26
Halt state may not be preserved across reset.

### LINKED ISSUE #357: Do halted harts remain halted after reset?
In section 3.2, Reset Control, we have:
> The halt state of harts should be maintained across system reset provided that dmactive is 1, ....

Does "halt state" here refer to the fact that a hart is halted, or does it mean the halt-on-reset option?  In the first interpretation, that would mean that all harts that were halted before system reset will be halted again after reset, even if their halt-on-reset options weren't enabled.  (Presumably, any harts that have the hart-on-reset option enabled will also be halted after system reset, but that's a separate matter.)

If the correct interpretation is the second one (harts are halted after system reset only if their halt-on-reset option is set), I believe the current wording is misleading.

On the other hand, if the correct interpretation is the first one (halted harts remain halted after system reset, in addition to any with the halt-on-reset option set), then does this property of "remains halted" apply also to resets of selected harts by `dmcontrol.hartreset`, or to resets that occur by outside means?  If so, this needs to be stated in the document.  If not, why not?

(FWIW, I favor keeping halted harts halted after reset.  Otherwise, I believe a debugger will regularly need to set and clear the halt-on-reset option to ensure this is the case.  Making the behavior automatic will help reduce the traffic required between a debugger and the DM.)

timsifive:
It's not necessary to have this requirement, since there are other
mechanisms to ensure harts halt out of reset if that's what the debugger
wants.

Fixes #357.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -81,8 +81,8 @@ indicating it's not possible to perform any abstract commands during this time.
 Once a hart's reset is complete, {\tt havereset} becomes set.  When a hart comes out
 of reset and \FdmDmcontrolHaltreq or \Fresethaltreq are set, the hart will
 immediately enter Debug Mode (halted state).
-If the hart was initially running it will execute normally (running state).
-If the hart was initially halted it should now be running but may be halted.
+Otherwise, if the hart was initially running it will execute normally (running state)
+and if the hart was initially halted it should now be running but may be halted.
 
 \begin{commentary}
     There is no general, reliable way for the debugger to know when reset has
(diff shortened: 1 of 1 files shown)


debug_module.tex
```diff
@@ -80,18 +80,15 @@ \section{Reset Control} \label{reset}
 indicating it's not possible to perform any abstract commands during this time.
 Once a hart's reset is complete, {\tt havereset} becomes set.  When a hart comes out
 of reset and \FdmDmcontrolHaltreq or \Fresethaltreq are set, the hart will
-immediately enter Debug Mode (halted state). Otherwise it will execute normally
-(running state).
+immediately enter Debug Mode (halted state).
+If the hart was initially running it will execute normally (running state).
```
ernie-sifive: Do we need an "otherwise" or "If haltreq and resethaltreq are both clear, then..." before the second part of this?

timsifive: That's probably better.

PR #505, opened 2019-12-02
Make haltsum0 optional if there is only one hart.

### LINKED ISSUE #389: haltsum0 should not be required for single hart systems
Arguably it shouldn't be required at all, since its address is 0x40 which requires an extra DMI address bit.

timsifive:
This is not forwards compatible. Debuggers that rely on haltsum0 would
break with this change. But it makes the debug interface on tiny 1-hart
systems smaller, especially because it allows the number of address bits
in the DMI to be 1 less. (Haltsum0 at 0x40 due to other efforts to
maintain backwards compatibility.)

Fixes #389.

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -615,6 +615,8 @@
         is halted or not. Unavailable/nonexistent harts are not considered to
         be halted.
 
+        This register may not be present in systems with fewer than 2 harts.
+
         The LSB reflects the halt status of hart \{hartsel[19:5],5'h0\}, and the
         MSB reflects halt status of hart \{hartsel[19:5],5'h1f\}.
 
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -615,7 +615,8 @@
         is halted or not. Unavailable/nonexistent harts are not considered to
         be halted.
 
-        This register may not be present in systems with fewer than 2 harts.
+        This register might not be present if fewer than 2 harts are connected
+        to this DM.
 
         The LSB reflects the halt status of hart \{hartsel[19:5],5'h0\}, and the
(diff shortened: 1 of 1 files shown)


timsifive:
@markuslauterbach, can you comment?

xml/dm_registers.xml
```diff
@@ -615,6 +615,8 @@
         is halted or not. Unavailable/nonexistent harts are not considered to
         be halted.
 
+        This register may not be present in systems with fewer than 2 harts.
```
pdonahue-ventana: "May not" is ambiguous, possibly meaning "shall not" and possibly meaning "may."  How about something like:
This register must be present in systems with 2 or more harts and may be in systems with 1 hart.

Also, is this statement related to the number of harts in the system or the number of harts attached to this DM?

markuslauterbach:
I can't really comment on whether it's justified to make a backwards-incompatible change here, but I agree that the change itself would be reasonable. So far my debug driver does not use haltsum at all.

xml/dm_registers.xml
timsifive: Does "might not" instead of "may not" work?

timsifive:
Thank you, @markuslauterbach. Knowing it won't be a problem for you either makes me more confident that this change is OK.

xml/dm_registers.xml
pdonahue-ventana: That sounds good.

My other question still applies.  Should it be something like "This register might not be present if fewer than 2 harts are attached to the DM." or do you think it's correct as written?

timsifive: You are right on that point. I've pushed a change to reflect all this.

PR #506, opened 2019-12-03
Add resume groups.

timsifive:
Mostly inspired by e-mail discussion with Subject:"resume groups in
addition to halt groups".

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -571,33 +571,48 @@
         This register contains DM control and status bits that didn't easily
         fit in \RdmDmcontrol and \RdmDmstatus. All are optional.
 
-        <field name="0" bits="31:11" access="R" reset="0" />
+        If halt groups are not implemented, then \FdmDmcsTwoGroup will always
+        be 0 when \FdmDmcsTwoGrouptype is 0.
+
+        If resume groups are not implemented, then \FdmDmcsTwoGrouptype will
+        remain 0 even after 1 is written there.
+
+        The triggers referred to in \FdmDmcsTwoExttrigger might be the same,
+        or might be completely separate sets, because it is not generally
+        useful to automatically halt some cores when other cores resume
+        or vice versa.
+
+        <field name="0" bits="31:12" access="R" reset="0" />
+        <field name="grouptype" bits="11" access="WARL" reset="0">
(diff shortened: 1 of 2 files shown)


timsifive:
@markusgoehrle

xml/dm_registers.xml
```diff
@@ -571,33 +571,48 @@
...
+        If halt groups are not implemented, then \FdmDmcsTwoGroup will always
+        be 0 when \FdmDmcsTwoGrouptype is 0.
+
+        If resume groups are not implemented, then \FdmDmcsTwoGrouptype will
+        remain 0 even after 1 is written there.
+
+        The triggers referred to in \FdmDmcsTwoExttrigger might be the same,
+        or might be completely separate sets, because it is not generally
+        useful to automatically halt some cores when other cores resume
+        or vice versa.
```
ernie-sifive: This could probably be phrased better.  I think you are saying that the external triggers available to add to halt groups may or may not be distinct from the external triggers available to add to resume groups.  In other words, external trigger 0 for halt groups might not be the same as external trigger 0 for resume groups.

xml/dm_registers.xml
```diff
@@ -571,33 +571,48 @@
...
-            When \FdmDmcsTwoHgselect is 1, contains the halt group of the external
+            When \FdmDmcsTwoHgselect is 1, contains the group of the external
             trigger selected by \FdmDmcsTwoExttrigger.
 
             Writes only have an effect if \FdmDmcsTwoHgwrite is also written 1.
 
             An implementation may tie any number of upper bits in this field to
-            0. If halt groups aren't implemented, then this entire field
-            is 0.
+            0. If groups aren't implemented, then this entire field is 0.
```
ernie-sifive: Should we specifically state that the number of halt groups need not be equal to the number of resume groups?  For example, writing a value of 2 in the "group" field might be legal if "grouptype" is being written to 0 but not if "grouptype" is being written to 1.  Maybe the "WARL" is sufficient for the spec but in this case, it's not just this field alone that's WARL - the legal values for this field depend on other fields.

timsifive:
I've addressed @ernie-sifive's comments above. I didn't explicitly add lines saying what happens if halt/resume happen at the same time, but I stuck with the "process of halting"/"process of resuming" language. Does that work?

PR #512, opened 2020-01-16
Adding version encoding for 0.14 spec.

bdwyatt:
I'm of the view that there are sufficient changes in 0.14 to warrant a distinct encoding for the version. This pull request simply adds that encoding to the spec.

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -100,8 +100,8 @@
             2: There is a Debug Module and it conforms to version 0.13 of this
             specification.
 
-	    3: There is a Debug Module and it conforms to version 0.14 of this
-	    specification.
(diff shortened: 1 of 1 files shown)


bdwyatt:
Yeah, I probably should have raised this as an issue for discussion first, but the goal of this is just to put it on the radar and make sure we consider it before finalizing 0.14.

rbohn:
There are changes since 0.13 that are backwards incompatible. The specific case I have in mind is https://github.com/riscv/riscv-debug-spec/pull/507 . This isn't theoretical: following the old FdmSbcsSbautoincrement text to the letter is incompatible with some versions of commercial debugger software available today. The debugger software needs to be able to discover which behavior it is potentially dealing with.

Issue #519, opened 2020-04-01
debugger should be able to request that a hart stay powered on


timsifive:
subject:"Abstract command related questions and suggestions"

Robert Chyla:
> Some existing processors provide a logic to prevent (selective) core[s] from going into some deep-sleep/not-clocked/not-powered modes. This is usually done in SoC specific way, but maybe commonly agreed 'RISC-V debug logic' should have dedicated bit for it? This way debugger may be relieved from doing 'magic' in SoC specific way. Some SoC may ignore it, but if someone is 'compliant' debugging of hard (but practical ...) conditions will be pleasant and predictable.

timsifive:
Is this intended to prevent halted harts from becoming unavailable, or running ones?

For halted harts, we can simply say that hardware should attempt to keep halted harts available. For running harts, we could add 2 bits to dmcs2: setkeepavailable and clearkeepavailable. They would set/clear the keepavailable status for all currently selected harts. If keepavailable is set, we request that hardware keep the harts available, if possible.

Thoughts, @pdonahue-ventana?

pdonahue-ventana:
This looks like a good way to request keepavailable to be set/cleared.

The actual availability is usually managed by something (one of the harts or a management processor or even some hardware state machine) that may take some time to actually see the change to keepavailable.  To avoid the race condition, it would be useful to have an acknowledgement mechanism like on dmactive.

It's a bit more complicated than dmactive since this affects multiple harts.  How about treating setkeepavailable like allkeepavailable on reads and clearkeepavailable is anykeepavailable on reads?  Something like:
> setkeepavailable: Writing 0 has no effect.  Writing 1 requests that the keepavailable bit be set for all selected harts.  On reads, this returns the AND of the actual keepavailable bits for each selected hart.  After writing 1, the debugger must poll until this bit reads as 1, indicating that the setkeepavailable request has been acknowledged for all selected harts and all of them have seen that keepavailable=1.

> clearkeepavailable: Writing 0 has no effect.  Writing 1 requests that the keepavailable bit be cleared for all selected harts.  On reads, this returns the OR of the actual keepavailable bits for each selected hart.  After writing 1, the debugger must poll until this bit reads as 0, indicating that the clearkeepavailable request has been acknowledged for all selected harts and all of them have seen that keepavailable=0.

timsifive:
[...]

PR #520, opened 2020-04-02
Debuggers should know when harts are unavailable.

### LINKED ISSUE #461: behavior when harts become unavailable during an abstract command
From Marc Gauthier:
The very possibility that a hart can become unavailable while halted is unsettling.  It can significantly increase debugger software complexity, if the debugger ends up having to constantly check whether the hart disappeared under its feet, perhaps completely changing state, or else the debugger might start doing something completely inconsistent based on false assumptions about hart stability.   It may help to establish some ground rules:  that if a hart becomes unavailable while in Debug Mode, then becomes available again, either it retains the same state it had when it became unavailable, or it has been reset and its havereset bit indicates it.  Even the case of retaining the same state might not be tractable, as the debugger does not know, given race conditions, what actions it might have taken while the hart was in reality unavailable; so perhaps that scenario ought to be disallowed completely, in other words if a hart becomes unavailable while halted, it must be reset to become available again.
Are there specific scenarios in mind?  (It seems an appropriate point to recommend that harts never be power-shutoff while dmactive=1 preferably, or at minimum while halted.  Other architectures have such mechanisms, to force power to remain ON while debugging hardware is active.)
The alternative to appear as hung seems incomplete:  debugger software can check and see that the hart is unavailable (via dmstatus.{any,all}unavail), so the hart will appear unavailable, not hung.  (Unless availability ping pongs on and off randomly, not a recommended possibility.)

timsifive:
Make unavail sticky, needing to be acknowledged with new ackunavail bit.

This is a backwards incompatible change. I think it's OK because
implementations that support unavailable harts are rare (I don't know of
any), and the change only means that debuggers must be updated. If
hardware doesn't implement this then everything will still work,
although users might not be notified when harts become unavailable.

Fixes #461.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -175,10 +175,10 @@ continuous hart index space. In order to let the debugger discover all harts,
 they must show up as unavailable even if there is no chance of them ever
 becoming available.
 
-While a hart is unavailable, {\tt unavail} is set for that hart. The bit
-remains set once the hart is available again, until the debugger acknowledges
-it using \FdmDmcontrolAckunavail. There are no guarantees about the state of
-the hart when it comes back.
+While a hart is unavailable, {\tt unavail} is set for that hart. If
+\FdmDmstatusStickyunavail is set, then the bit remains set once the hart is
+available again, until the debugger acknowledges it using
+\FdmDmcontrolAckunavail. There are no guarantees about the state of the hart when it comes back.
 
 Harts are running when they are executing normally, as if no debugger was
 attached. This includes being in a low power mode or waiting for an interrupt,
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -175,6 +175,11 @@ continuous hart index space. In order to let the debugger discover all harts,
 they must show up as unavailable even if there is no chance of them ever
 becoming available.
 
+While a hart is unavailable, {\tt unavail} is set for that hart. The bit
+remains set once the hart is available again, until the debugger acknowledges
+it using \FdmDmcontrolAckunavail. There are no guarantees about the state of
+the hart when it comes back.
+
 Harts are running when they are executing normally, as if no debugger was
 attached. This includes being in a low power mode or waiting for an interrupt,
 as long as a halt request will result in the hart being halted.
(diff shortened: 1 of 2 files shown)


rtwfroody:
I imagined that unavail would remain set until it's acknowledged, so it's possible to have unavail set together with running/halted.

I'm puzzling to see if we can make hardware backwards compatible at least without using an extra bit. I'm not very concerned about debuggers needing to be updated because there are relatively few of them and they are relatively easy to change.

timsifive:
I imagined that unavail would remain set until it's acknowledged, so it's possible to have unavail set together with running/halted.

I'm puzzling to see if we can make hardware backwards compatible at least without using an extra bit. I'm not very concerned about debuggers needing to be updated because there are relatively few of them and they are relatively easy to change.

RobertChyla:
Let me comment - maybe it will go through this time.
I can drop debugger compatibility request (extra bit), however some people may be forced to use older debugger (from many reasons).
But it is really important to handle situation, when hart is leaving 'unavail' state for very short period of time. Maybe stop request should force transition from 'unavail' into 'stopped' and if core/hart is in 'stopped' state some SoC may prevent to put it into 'unavail' (by blocking clock/power gating)?
This may also life of simple debuggers when core/hart will never 'disappear' once stopped.
Again - forgive me lack of precise terms - my knowledge of exact bits faded away.

timsifive:
This proposal explicitly addresses harts briefly becoming unavailable. In the "Files changed" tab above you can see the exact language I wrote.

timsifive:
This doesn't really address @ernie-sifive 's feedback that the intro to Section 3.4 implies only one of these bits is set. We don't even really mention these bits, and I'm struggling with a concise way to describe them: something that's helpful and not even more confusing.

Maybe what's needed is a state diagram/table...

timsifive:
I've tweaked this a little, but I rescind my earlier comment about things not being clear. As it is, the states are defined in Section 3.4, while the description of what exactly the bits do is described in the register definition.

Thoughts?

pdonahue-ventana:
This seems to address #461 where the debugger can know if a hart unexpectedly became unavailable.  For a complete solution, can there also be a mechanism to address #519 to prevent this from happening to start with (similar to ARM's CORENPDRQ)?

ernie-sifive (approved):
This change looks good to me.

PR #526, opened 2020-06-12
Explicitly allow uni-directional external triggers.

timsifive:
Also document a convention for trigger number by direction.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -263,8 +263,8 @@ external triggers are referred to by a number. Commonly, external triggers
 are capable of sending a signal from the system into the DM, as well as
 receiving a signal from the DM to take their own action on. It is also
 allowable for an external trigger to be input-only or output-only. By
-convention, external triggers 0--7 are bidirectional, triggers 8--11 are
-input-only, and triggers 12--15 are output-only.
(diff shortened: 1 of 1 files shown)


debug_module.tex
```diff
@@ -259,6 +257,15 @@ \section{Halt Groups, Resume Groups, and External Triggers}
...
 if other harts in the group are currently running.
 
+External triggers are abstract concepts that can signal the DM and/or receive
+signals from the DM. This configuration is done through \RdmDmcsTwo, where
+external triggers are referred to by a number. Commonly, external triggers
+are capable of sending a signal from the system into the DM, as well as
+receiving a signal from the DM to take their own action on. It is also
+allowable for an external trigger to be input-only or output-only. By
+convention, external triggers 0--7 are bidirectional, triggers 8--11 are
+input-only, and triggers 12--15 are output-only.
```
ernie-sifive, line 267: So "by convention" means that an implementation is not required to use this numbering scheme, right?

ernie-sifive: I realized that the group field of dmcs2 is 5 bits.  If the intention was to use the 2 MSBs to indicate the trigger direction, then the numbering would be 0-15 bidirectional, 16-23 input, 24-31 output.

Issue #448, opened 2019-04-24
Name of the document


pdonahue-ventana:
The draft version 0.14.0 of the debug spec is titled "RISC-V External Debug Support".

Since the functionality described in the document provides for debugging via software with no external debugger (via action=0) and there is no competing RISC-V proposal for non-external debug, it seems that the name of the document should be changed from "RISC-V External Debug Support" to the more general "RISC-V Debug Support".

timsifive:
"Debug" could be interpreted to be a lot broader than "External Debug," but I don't have a strong opinion.
If you can get some other people to agree that this is a good change I'm happy to change it.

pdonahue-ventana:
> "Debug" could be interpreted to be a lot broader than "External Debug,"

Exactly.  Because the spec addresses a broader set of issues than external debug, the title is misleading.  If it described only OpenOCD-style debugging then the current title is fitting.  Since it also addresses gdb-style debugging, a more generic name is appropriate.

I'll ask on tech-debug.

pashenden:
Works for me.

PR #534, opened 2020-07-13
Rename document, "ISA vs Non-ISA Specifications"

timsifive:
See https://docs.google.com/document/d/1bSJSnPfL17kDTL-CUB4xE32Zjp77RTAk1XOKXQF_3Qc/edit

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ make
 ```
 
 Contributing
-------------------
+------------
 
 There are various ways to contribute to this spec. You can use a combination of them to get your idea across.
 Please note that pull requests will only be reviewed/accepted from RISC-V Foundation members.
@@ -24,7 +24,7 @@ Please note that pull requests will only be reviewed/accepted from RISC-V Founda
 3. Discuss higher-level questions or ideas on the riscv-debug-group mailing list: https://groups.google.com/a/groups.riscv.org/forum/#!forum/debug
 
 For More Information
-------------------
+--------------------
 
 Additional information can be found at
 https://github.com/riscv/debug-taskgroup
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-RISC-V Debug Specification
-==========================
+Debug Specification for RISC-V
+==============================
 
 You may be looking for one of the following pre-built PDFs:
 * [Latest draft](https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-draft.pdf)
--- a/riscv-debug-spec.tex
+++ b/riscv-debug-spec.tex
@@ -69,7 +69,7 @@
 
 \begin{document}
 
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
-RISC-V Debug Specification
-==========================
+Debug Specification for RISC-V
```
ernie-sifive, line 1: Should this be "External Debug Specification for RISC-V"?  That's what is being called within the doc.

riscv-debug-spec.tex
```diff
@@ -69,7 +69,7 @@
 
 \begin{document}
 
-\title{RISC-V External Debug Support\\
+\title{External Debug Support for RISC-V\\
```
ernie-sifive, line 72:
```suggestion
\title{External Debug Specification for RISC-V\\
```

riscv-debug-spec.tex
```diff
@@ -123,8 +123,8 @@
 Andy Wright,
 and Bryan Wyatt.
 
-\markboth{RISC-V External Debug Support Version \versionnum}
-{RISC-V External Debug Support Version \versionnum}
+\markboth{External Debug Support for RISC-V Version \versionnum}
+{External Debug Support for RISC-V Version \versionnum}
```
ernie-sifive, line 127:
```suggestion
\markboth{External Debug Specification for RISC-V Version \versionnum}
{External Debug Specification for RISC-V Version \versionnum}
```

README.md
pdonahue-ventana: This document doesn't only describe external debug (in particular, action=0 causes trap-based debug where no external debugger is present).  So since we're renaming the document anyway, I'll re-raise [my point from last year](https://github.com/riscv/riscv-debug-spec/issues/448) that the document name should not have the word "external".  I've talked to at least a few people who thought that the only way to do debug on RISC-V is with an external debugger.

ernie-sifive: Good point, no objection.  The name should be consistent throughout.

timsifive: I'll do that.

pdonahue-ventana:
A question occurred to me this morning: Is the debug spec not part of the ISA?

It specifies:
- several CSRs
- a new instruction
- a new mode

The latter two may be hidden during normal use so that's arguable.*  But the CSRs are definitely visible and seem to be part of the ISA.

\* There is a slightly more stringent requirement that dret is an illegal instruction outside of Debug Mode whereas the base architecture treats it as UNSPECIFIED.  But it's indistinguishable whether dret trapped due to this requirement or if it trapped as part of UNSPECIFIED.

timsifive:
Good points. I'll ask Mark if he can clarify.

timsifive:
I haven't heard any clarification on what is meant by ISA vs. non-ISA, so closing this PR.

PR #536, opened 2020-07-29
Add abstractcs.relaxedpriv

pdonahue-ventana:
New dmcontrol.relaxedpriv bit as discussed in https://lists.riscv.org/g/tech-debug/topic/75549024

### DIFF
--- a/core_debug.tex
+++ b/core_debug.tex
@@ -20,6 +20,8 @@ How Debug Mode is implemented is not specified here.
     Debug Mode and the following apply:}
 \item All operations are executed at machine mode privilege level, except that
     \FcsrMcontrolMprv in \Rmstatus may be ignored according to \FcsrDcsrMprven.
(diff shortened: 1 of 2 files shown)


ernie-sifive:
May I suggest that this new control bit be put in `abstractcs` instead of `dmcontrol` since it specifically applies to that function?  This will also solve the problem of what happens if you change the bit during execution of an abstract command or the program buffer (`abstractcs.cmderr` will be set to 1).

pdonahue-ventana:
> May I suggest that this new control bit be put in `abstractcs` instead of `dmcontrol`

Good point.  I'll make changes (probably tomorrow) to move it to abstractcs[11].

pdonahue-ventana:
Take a look now.  I changed the name of this PR, moved the bit to abstractcs, and fixed the LaTeX error.

PR #542, opened 2020-08-27
aamvirtual: Clarification for systems without address translations

JanMatCodasip:
Adding brief clarification of permitted "aamvirtual" bit behavior on systems without address translation, per the following discussion: https://lists.riscv.org/g/tech-debug/topic/72759746

### DIFF
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -191,6 +191,10 @@
 
             1: Addresses are virtual, and translated the way they would be from
             M-mode, with \FcsrMcontrolMprv set.
+
+            Debug Modules on systems without address translation (i.e. virtual addresses equal physical)
+            may optionally allow \FacAccessmemoryAamvirtual set to 1, which would produce the same result as
+            that same abstract command with \FacAccessmemoryAamvirtual cleared.
         </field>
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
Although this is really a detail, I believe it is worth to have the output of the above discussion noted in the spec.

Please feel free to edit/reword the commit as needed.

xml/abstract_commands.xml
```diff
@@ -191,6 +191,10 @@
 
             1: Addresses are virtual, and translated the way they would be from
             M-mode, with \FcsrMcontrolMprv set.
+
+            Debug Modules on systems without address translation (i.e. virtual addresses equal physical)
+            may optionally allow \FacAccessmemoryAamvirtual bit set to 1, which would produce the same result as
+            the same abstract command with \FacAccessmemoryAamvirtual bit cleared.
```
timsifive, line 197: Our convention seems to be to say just "\FacAccessmemoryAamvirtual" instead of "\FacAccessmemoryAamvirtual bit". Not all the spec does that, but most does, and it is more concise while just as readable.

JanMatCodasip: Thanks, I have reworded it as suggested.

PR #543, opened 2020-09-01
Added exttrigger capability as type 7

pdonahue-ventana:
Enables cross triggering between debug and other blocks such as hardware performance monitors.
Skipped over type 6 assuming #538 will be approved.


trigger.tex
```diff
@@ -69,6 +69,8 @@ \section{Actions}
 \hline
 2 -- 5 & Reserved for use by the trace specification. \\
 \hline
+8 -- 9 & Signal the firing of the trigger over an external debug trigger output to other blocks within the hart (e.g. as countable events to hpmcounters).  Generate a pulse on external debug trigger output 0 or 1 (respectively). \\
```
timsifive, line 72: I'd omit "over an external debug trigger output". Is it important that this is a pulse? That sounds like an implementation detail, but if there's already something elsewhere that says hpmcounters can have inputs that are pulsed then it would make sense to use the same definition here.

xml/hwbp_registers.xml
```diff
@@ -549,6 +552,37 @@
...
+        <field name="hit" bits="XLEN-6" access="WARL" reset="0">
+            If this bit is implemented, the hardware sets it when this
+            trigger matches. The trigger's user can set or clear it at any
+            time. It is used to determine which
+            trigger(s) matched.  If the bit is not implemented, it is always 0
+            and writing it has no effect.
+        </field>
+        <field name="0" bits="XLEN-7:10" access="R" reset="0" />
+        <field name="select" bits="9:6" access="WARL" reset="0">
+            Selects one of up to 16 external debug trigger inputs that would
```
timsifive, line 577: Would it be better to make this a 16-bit field, allowing the trigger to fire when any of the selected inputs sends a signal?

xml/hwbp_registers.xml
```diff
@@ -549,6 +552,37 @@
         </field>
     </register>
 
+    <register name="External Trigger" short="exttrigger" address="0x7a1">
+        This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 7.
+
+        This trigger fires when a selected external debug trigger input
+        asserts (e.g. when a pulse is received).  One of up to 16 external
```
timsifive, line 559: I'd use "signals" instead of "asserts," and remove the parenthetical.

timsifive (changes requested):
Looks like a reasonable way to implement this feature. Comments below.

Please don't submit riscv-debug-draft.pdf in your pull requests. This leads to annoying conflicts, and generally git isn't great at maintaining lots of binary history.

I manually rebuild the PDF once a month, just before the meeting. That seems to be enough to give people who don't want to build a PDF something to use.

trigger.tex
pdonahue-ventana: I updated this.

xml/hwbp_registers.xml
pdonahue-ventana: Sounds good.  The downside is that there are fewer bits for other purposes but I don't think there's a need for a large number of additional bits.

xml/hwbp_registers.xml
pdonahue-ventana: I updated this and specified that it fires if "any" selected input signals to be explicit that this is an OR function.

timsifive (approved):
Looks good. Thanks! I'll merge this Friday unless there are more comments. @ernie-sifive, do you have anything to add?

xml/hwbp_registers.xml
```diff
@@ -549,6 +552,37 @@
...
+    <register name="External Trigger" short="exttrigger" address="0x7a1">
+        This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 7.
+
+        This trigger fires when any selected external debug trigger input
+        signals.  Up to 16 external
+        debug trigger inputs coming from other blocks within the hart,
+        (e.g. signaling an hpmcounter overflow) can be selected.  Hardware
+        may support none or just a few external debug trigger inputs
+        (starting with external debug trigger input 0 and continuing
+        sequentially).  Unsupported inputs are hardwired to be inactive.
```
ernie-sifive, line 564: I wonder if we should use a name other than "external debug trigger" to avoid confusing these with the external triggers in the debug module referred to in dmcs2 and used for halt and resume groups.  Maybe something like intra-hart logic trigger?

pdonahue-ventana: Sure.  Intra-hart logic trigger is fine with me.

There's the description but also the register name.  This register name should be similar to the description and should probably avoid using the same name as the dmcs2 exttrigger field (which it currently conflicts with).  What should this register be called?
"intrahartlogictrigger" is too long.
"ihlt" is too short.
"ihltrigger"?
"iharttrigger"?  (The unicode "I❤️trigger" is probably out. 😄)
"intrahart"?

Or maybe there's a description other than "intra-hart logic trigger" that lends itself to a more natural name.

timsifive: This is a tricky one. Intra-hart seems limiting. These triggers could be connected to something outside the RISC-V core altogether, and that should be fine.

They really are external triggers, going to a hart. The external triggers in dmcs2 are also external triggers, going to a Debug Module. Should we rename both? And then the direction matters, too.

dmcs2.exttrigger refers to input/output triggers to/from the DM.
exttrigger.select refers to input triggers to the hart.

I'm not arriving at a great name here, but just sharing some thoughts.

timsifive:
@pdonahue-ventana, what are your thoughts on the naming issue?

ernie-sifive:
How about "Hart Logic Trigger" or simply "Logic Trigger", then explicitly state that the logic signal can come from a source within the hart or from outside the hart.  Problem is this suggests that they have to be logic signals and not abstractions that might be from a message-passing network or something else.

Another possibility is to rename existing external triggers to "DM External Trigger" or "Debug External Trigger" and then call these new triggers "Hart External Trigger".

timsifive:
What if we just call them "dm trigger" and "hart trigger?" I think external doesn't really add anything, and if somebody wants to wire these up to sources that are not external that's probably also fine.

pdonahue-ventana:
I think Ernie's "DM External Trigger" sounds good since it's for things external to the DM.  I'd similarly call the new one "TM External Trigger" since it's for things that are external to the TM, regardless of whether they're coming from something associated with this hart (like HPM or trace) or from outside the hart.  I prefer this over "hart trigger" because I consider the internal TM triggers (load/store, pc, single step, etc.) to be hart triggers.

Thoughts?

ernie-sifive:
@pdonahue-ventana I agree - we need terminology that indicates these entities are signal inputs rather than a block containing registers and comparators that can cause breakpoints (which is what I would call a "trigger").

timsifive:
So "DM external trigger" and "TM external trigger" then?

pdonahue-ventana:
Sounds good to me.  Does this mean that this register should be tmexttrigger and the dmcs2 field should change from exttrigger to dmexttrigger (plus some minor description changes)?

pdonahue-ventana:
That change just now was to resolve merge conflicts.  I won't change names until I hear back.

ernie-sifive (approved):
Thanks, that looks good to me.

timsifive (changes requested):
Just 1 comment.

xml/hwbp_registers.xml
```diff
@@ -49,6 +49,9 @@
             to a type 2 trigger, but provides additional functionality and
             should be used instead of type 2 in newer implementations.
 
+            7: The trigger is a trigger source external to the TM.  The
```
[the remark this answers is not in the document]
pdonahue-ventana: Good catch.  I updated the name of chapter 5.  An acronym section would definitely be nice but I'll leave that to another ambitious person.

timsifive (approved):
Looks good. I'll merge this tomorrow unless there are more comments.

PR #546, opened 2020-09-08
Fix links to point to fields in the correct registers.

pdonahue-ventana:
When doing mcontrol6 work, I noticed that icount and etrigger have references to fields in mcontrol rather than to fields in themselves.  I'm making this a separate PR since these changes are really unrelated to mcontrol6.


timsifive (approved):
Good catch! I'll merge this tomorrow, unless there are more comments.

PR #549, opened 2020-09-12
Hypervisor support

pdonahue-ventana:
Added support for the H extension.  Also:
- Clarify that the backwards-compatible mscontext is optional.
- Clarify that registers are "accessible" in specified modes, not just "writable"
- Prohibit hcontext alias when the H extension isn't supported since this is in the CSR address space that's normally just for hypervisor.


xml/hwbp_registers.xml
```diff
@@ -155,21 +155,27 @@
...
+            0. If the H extension is not supported, it's recommended to implement
+            no more than 6 bits on RV32, and 13 on RV64.  If the H extension is
+            supported, it's recommended to implement no more than 7 bits on RV32
+            and 14 on RV64.
         </field>
     </register>
 
     <register name="Supervisor Context" short="scontext" address="0x5a8">
-        This optional register is only writable in S mode, M mode and Debug Mode.
+        This optional register is only accessible in S mode, M mode and Debug Mode.
```
ernie-sifive, line 178: I think this register is also accessible in HS mode and VS mode, right?

xml/hwbp_registers.xml
```diff
@@ -820,17 +869,27 @@
...
-            1: This trigger will only match if the low bits of
-            \RcsrMcontext equal \FcsrTextraThirtytwoMvalue.
+            100: This trigger will only match if the low bits of
+            \RcsrMcontext/\RcsrHcontext equal \FcsrTextraThirtytwoMhvalue.
+
+            x01: This trigger will only match if the low bits of
+            \RcsrMcontext/\RcsrHcontext equal {\FcsrTextraThirtytwoMhvalue, mhselect[2]}.
+
+            x10: This trigger will only match if VMID in satp equals the lower VMIDMAX
+            (defined in the Privileged Spec) bits of {\FcsrTextraThirtytwoMhvalue, mhselect[2]}.
```
ernie-sifive, line 885: The braces in these two descriptions are not being rendered in the PDF.

xml/hwbp_registers.xml
```diff
@@ -857,7 +916,7 @@
         This is the layout of {\tt textra} if XLEN is 64. The fields are defined
         above, in \RcsrTextraThirtytwo.
 
-        <field name="mvalue" bits="63:51" access="WARL" reset="0">
+        <field name="mhvalue" bits="63:51" access="WARL" reset="0">
         </field>
         <field name="mselect" bits="50" access="WARL" reset="0">
```
ernie-sifive, line 921: Should be called mhselect and should be 3 bits wide in bits 50:48.

ernie-sifive (commented):
Thanks for doing this!  I have a few minor comments.

xml/hwbp_registers.xml
pdonahue-ventana: I missed that.  I'll push the fix in a minute.

xml/hwbp_registers.xml
gfavor: First off, at a minimum this should say "S/HS mode", i.e. they are different names - based on the context of whether the H-extension is implemented and enabled - for the same privilege level.  If the H extension is enabled, then S-mode is now called HS-mode and certain additional instructions and CSRs become available (on top of all the standard S-mode arch functionality).

Regarding VS mode, the H-extension generally (i.e. in most but not all cases) takes the approach that there  are s* and vs* versions of what originally were just s* CSRs.  They are physically distinct registers.  In HS-mode both are accessible and the s* registers are active (in having whatever effects they are defined to have).  In VS-mode vs* CSRs are inaccessible directly, but they become "active" and are accessed as s* CSRs (in place of the normal s* CSRs which are inactive in VS-mode).  In other words, the vs* CSRs masquerade as the s* CSRs in VS-mode.

In this case we either add a vscontext CSR that behaves as just described (but is that cost worth it?), or we stick with just the current scontext CSR and make it accessible in both S/HS and VS modes.  I believe in past email discussion we decided to take the latter approach.  The hypervisor will context switch scontext when switching between VMs (just as it would have done with vscontext).  If scontext is not being used within HS and U modes, then that's it.  Otherwise the hypervisor also needs to context switch scontext between HS/U mode usage and VS/VU mode usage.

So, in short, scontext should be accessible in S/HS mode, VS mode, M mode, and Debug mode.

pdonahue-ventana: I added that it's accessible in VS mode.  I also changed the mode filter bits in various CSRs to refer to S/HS mode instead of just S mode (since HS and S share the same filter bit depending on whether hypervisor is supported).

ernie-sifive (approved):
Looks good to me.

xml/hwbp_registers.xml
```diff
@@ -820,18 +871,27 @@
...
+        <field name="mhselect" bits="25:23" access="WARL" reset="0">
+            000: Ignore \FcsrTextraThirtytwoMhvalue.
 
-            1: This trigger will only match if the low bits of
-            \RcsrMcontext equal \FcsrTextraThirtytwoMvalue.
+            100: This trigger will only match if the low bits of
+            \RcsrMcontext/\RcsrHcontext equal \FcsrTextraThirtytwoMhvalue.
+
+            x01: This trigger will only match if the low bits of
+            \RcsrMcontext/\RcsrHcontext equal \{\FcsrTextraThirtytwoMhvalue, mhselect[2]\}.
```
timsifive, line 884: I assume this means 001 or 101. It's probably more clear to write "001/101" explicitly instead of relying on x01. Alternatively, you could explain what is meant by x01 in the Register Definition Format section.

timsifive (changes requested):
Looks good overall, just one comment.

Thanks again for taking this on.

xml/hwbp_registers.xml
pdonahue-ventana: Even ignoring the wildcard, the rest of the document consistently uses decimal instead of binary.  So I changed it to list all the decimal values.  This means that the decimal values are not in order but the section is short enough that I don't think it's hard to comprehend and this keeps things reasonably compact.

I also fixed an incorrect reference to satp here.

timsifive (approved):
Looks good. Thank you. I'll merge this tomorrow unless there are more comments.

Issue #551, opened 2020-09-16
ephemeral link to github


scottj97:
Change 2f1c133e6b06fe2c9cb605e9fd15d5e20fff91d6 introduced a symlink to `https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/devices/debug/Debug.scala#L170` in the spec. Today this points to `class DebugExtTriggerIn` which I suppose is intended.

But the link is to the master branch of rocket-chip, which will change over time. Next year, line 170 might have some unrelated code, leading to confusion.

What if the link pointed to a specific git commit of that file? e.g. `https://github.com/chipsalliance/rocket-chip/blob/db3003aae25b4bdbc4189c1680ab7ff5c313976f/src/main/scala/devices/debug/Debug.scala#L170`?

That way, a year from now, the link will point to an out-of-date file, but at least it will still be pointing to the intended class in the code instead of some random line.

ernie-sifive:
Addressed in https://github.com/riscv/riscv-debug-spec/pull/552

PR #552, opened 2020-09-16
Update rocket-chip link in appendix to specific commit for permanence

ernie-sifive:
Addresses issue #551


timsifive (approved):
Looks good!

PR #553, opened 2020-09-25
Remove duplicate list item

ernie-sifive:
Debug Mode characteristics list had a duplicate item.


pdonahue-ventana:
I don't understand this change (unless the goal is to require implementations to allow branching within the program buffer which is not backwards-compatible).  It's not a duplicate of the following item because one constrains branches and loops within the program buffer and one constrains branches from the program buffer to somewhere outside the program buffer.

Some implementations may distinguish between branches/loops inside the program buffer and branches that escape from the program buffer to somewhere else.  Because of that, the second sentence of each item prohibits combining these two items into a unified item that control transfer instructions may be illegal regardless of destination.

pdonahue-ventana:
Sorry, I thought this had been merged instead of closed.  I think you already caught the problem.

Issue #555, opened 2020-10-05
Interaction of icount and tcontrol


scottj97:
# Spec

When `icount.m==1`: "When set, every instruction completed or exception taken in M-mode decrements count by 1."

For `icount.count`: "When count is decremented to 0, the trigger fires."

When `tcontrol.mte==0`: "Triggers with action=0 do not match/fire while the hart is in M-mode."

# Questions

Use case: debugging M-mode software using icount breakpoints.

When `tcontrol.mte==0`, clearly the breakpoint exception should not be taken regardless of the count.

Should count still be decremented when executing instructions when `tcontrol.mte==0`? It seems the spec requires it. But then what happens to the breakpoint exception? It cannot be taken because of `mte`.

It seems bad to just drop the exception and never take it. The debugger would then run away without ever stopping as expected.

## Options

When we do `mret` which sets `tcontrol.mte==1`, we could take the breakpoint exception at that time. This would imply that the hardware icount register will have to remember the pending trigger.

Alternatively, when executing in M-mode with `tcontrol.mte==0`, stop decrementing count (if `action==0`), since it can't trap anyway.

ernie-sifive:
I think your second option is the right answer.

One purpose of `mte` is to prevent triggers to M-mode from occurring within an M-mode trap handler because this would overwrite `mepc` and render the system unrecoverable.

When using icount triggers from an M-mode debugger, clearly the counter can't be enabled until the M-mode debugger performs `mret` to return to the system code.  You don't want icount counting instructions within the debugger.  So I think using `mte` to enable icount both to count and to trigger is the solution. 

This will have the side effect of preventing stepping of M-mode trap handlers from an M-mode debugger, but M-mode trap handlers are already un-debuggable.

agrobman:
I think the spec is over protecting here by preventing breakpoints in exception handlers. Yes, breakpoints in handlers prolog and epilogs will destroy CPU state and not recoverable, but  for handlers written in C with prolog/epilog saving/ restoring mstatus/mpc user still can safely set breakponts in the handler bodies. 
Even BP setting at the beginning /end of the handler can be useful sometimes, when user wants to know if a specific exception/interrupt ever happens

scottj97:
Wouldn't such interruptible handlers set tcontrol.mte and/or mstatus.mie to enable breakpoints?

[...]

PR #557, opened 2020-10-06
Don't decrement icount.count when exception is blocked by tcontrol.mte

scottj97:
See #555.

I have not tested this change by building a PDF (or anything else).

cc @ernie-sifive

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -727,9 +727,9 @@
 
         <field name="m" bits="9" access="WARL" reset="0">
             When set, every instruction completed in or trap taken from
-            M-mode decrements \FcsrIcountCount by 1. But if tcontrol is
-            implemented, \FcsrTcontrolMte is 0, and \FcsrIcountAction is
-            0, no decrement occurs.
+            M-mode decrements \FcsrIcountCount by 1, unless \FcsrIcountAction
+            is 0 and M-mode breakpoints are disabled via one of the mechanisms
+            described in Section~\ref{sec:mmtrigger}.
         </field>
         <field name="0" bits="8" access="R" reset="0" />
         <field name="s" bits="7" access="WARL" reset="0">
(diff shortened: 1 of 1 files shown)


ernie-sifive (approved):
This looks good to me.

timsifive (changes requested):
I'd write this as:
            When set, every instruction completed in or trap taken from
            M-mode decrements \FcsrIcountCount by 1, unless \FcsrTcontrolMte and
            \FcsrIcountAction are both 0.

I don't think any existing behavior is contingent on tcontrol being implemented. If it's not implemented, then mte is assumed to always be 0, which is the simpler behavior. Making triggers work in M mode is really a fancy corner case. Generally if you're debugging M mode software you're using an external debugger.

scottj97:
> 
> 
> I'd write this as:
> When set, every instruction completed in or trap taken from
> M-mode decrements \FcsrIcountCount by 1, unless \FcsrTcontrolMte and
> \FcsrIcountAction are both 0.
> 
> I don't think any existing behavior is contingent on tcontrol being implemented. If it's not implemented, then mte is assumed to always be 0, which is the simpler behavior. Making triggers work in M mode is really a fancy corner case. Generally if you're debugging M mode software you're using an external debugger.

I don't think mte is assumed to be 0 if tcontrol is not implemented. If it were, this statement in 5.4 would not make sense: "The simple solution is to have the hardware prevent triggers with action=0 from firing while in M mode and while MIE in mstatus is 0. Its limitation is that interrupts might be disabled at other times when a user might want triggers to fire."

If mte were assumed to be 0 then there would be no need to consider MIE as the trap would already be blocked.

scottj97:
Though that does bring up an interesting point. If an implementation uses the `mstatus.MIE` mechanism to block M-mode breakpoints, perhaps that should also suppress the decrement of `count`, for the same reasons.

scottj97:
I've pushed another commit that simplifies the description by referring to section 5.4.

ernie-sifive (approved):
This wording is better, thanks.

timsifive (approved):
I'll merge this tomorrow, pending more comments.

scottj97:
I guess this should have the "Backwards Incompatible Change" label?

scottj97:
Note that this has been reversed by #574 since that change makes this problem moot.

PR #560, opened 2020-10-07
Explicitly add custom1-custom14

pdonahue-ventana:


xml/dm_registers.xml
```diff
@@ -950,6 +950,76 @@
...
+    <register name="Custom Features 13" short="custom13" address="0x7d">
+      This optional register may be used for non-standard features. Future
+      version of the debug spec will not use this address.
+    </register>
+
+    <register name="Custom Features 14" short="custom14" address="0x7e">
+      This optional register may be used for non-standard features. Future
+      version of the debug spec will not use this address.
+    </register>
+
```
ernie-sifive, line 1022: I wonder if we should do this in a way similar to the data0 - data11 registers.  There is only one entry in the Debug Module Registers table and one register description that starts with "data0 through data11 are basic read/write registers that..."

pdonahue-ventana: The data0-data11 registers are already in the XML (which is indirectly used to generate the PDF) but they're commented out: https://github.com/riscv/riscv-debug-spec/blob/master/xml/dm_registers.xml#L522

That was done in 3140efa24a39e8fe98ea768fb1262c267f70cbbb by @timsifive for the data registers and the program buffer registers.  I think that all registers should be listed, at least in table 3.8.  The .h files should probably also have them all.

timsifive: It makes sense to list the registers in Table 3.8, but giving each their own subsubsection with identical text just takes up space for no reason. Except if we do that Table 3.8 runs off the end of the page. I will try to fix registers.py to accommodate that.

pdonahue-ventana: I agree that both custom\* and data\* registers should be in the tables but don't need lots of identical sections.

I tried uncommenting the data registers and then they showed up in the table but the addresses were blank.  I assume that the script was written to handle the `<register></register>` syntax and didn't like the `<register ... />` syntax but I didn't dig deeper.

timsifive:
Closing in favor of #579.

PR #561, opened 2020-10-09
Add support for the A extension

pdonahue-ventana:
Followup to https://lists.riscv.org/g/tech-debug/topic/76138908

I tried to provide flexibility for implementations that:
- don't return data on AMOs to x0 (and, for similar reasons, for loads to x0)
- have different ways of handling failing SC instructions
- do store data comparison in various places with relation to where AMOs are performed


xml/hwbp_registers.xml
```diff
@@ -220,6 +220,18 @@
...
+        \begin{tightlist}
+        \item LR instructions are loads
+        \item successful SC instructions are stores
+        \item it is \unspecified\ whether failing SC instructions are stores or not
+        \item Each AMO instruction is both a load and a store
+        \end{tightlist}
+
+        If the destination register of any load or AMO is x0 then it is
+        \unspecified\ whether a load data trigger will match.  Whether store
+        data triggers match on AMOs is \unspecified.
```
ernie-sifive, line 234: Is it clear enough what is meant by "load data trigger" and "store data trigger"?  It is to me but it might be more precise to say "... a trigger with \FcsrMcontrolSelect = 1".

pdonahue-ventana: I think it would be nice to have a glossary with terms like this.  Every time I read a passage like that I have to go look up what select=1 or timing=0 or action=1 means.  The descriptive terms would make it much easier to read.

That being said, it's a non-trivial effort which is beyond the scope of this particular PR so I'll take your advice.

ernie-sifive (approved):
:+1:

xml/hwbp_registers.xml
```diff
@@ -220,6 +220,19 @@
         hold all valid virtual addresses but it need not be capable of holding
         other values.
 
+        If the A extension is supported, then trigger behavior is as follows
+        for the load and store bits:
+        \begin{tightlist}
+        \item LR instructions are loads
```
timsifive, line 226: LR -> {\tt lr}

xml/hwbp_registers.xml
```diff
@@ -220,6 +220,19 @@
...
+        \item successful SC instructions are stores
```
timsifive, line 227: SC -> {\tt sc}

xml/hwbp_registers.xml
```diff
@@ -220,6 +220,19 @@
         hold all valid virtual addresses but it need not be capable of holding
         other values.
 
+        If the A extension is supported, then trigger behavior is as follows
+        for the load and store bits:
+        \begin{tightlist}
```
timsifive, line 225: I'm not sure what exactly tightlist does, but I'm guessing it's like steps which I made for this document. So write:
`\begin{steps}{If the A extension is supported, then trigger behavior is as follows for the load and store bits:}`
which also eliminates the empty line between the text and the first list entry.

xml/hwbp_registers.xml
```diff
@@ -220,6 +220,19 @@
...
+        \item LR instructions are loads
+        \item successful SC instructions are stores
+        \item it is \unspecified\ whether failing SC instructions are stores or not
```
timsifive, line 228: SC -> {\tt sc}

xml/hwbp_registers.xml
```diff
@@ -220,6 +220,19 @@
...
+        \item Each AMO instruction is both a load and a store
+        \end{tightlist}
+
+        If the destination register of any load or AMO is x0 then it is
```
timsifive, line 232: x0 -> \Rzero

timsifive (changes requested):
Content looks good to me. I just have some formatting feedback.

xml/hwbp_registers.xml
pdonahue-ventana: I was using all-caps since that's what the unprivileged and privileged specs do when referring to instructions.  They use lower-case only in code snippets.  Except in appendix B of the unprivileged spec.

But I'll change it to be consistent with the rest of this document.  Maybe someday RVI will have a techpubs department that gets everything to have a similar look and feel across documents and, as with appendix B, within the same document.

xml/hwbp_registers.xml
pdonahue-ventana: I really think that this should be an unordered list and I picked tightlist over itemize to reduce empty space.

xml/hwbp_registers.xml
pdonahue-ventana: I was trying to be consistent with the A extension:
> More complex implementations might also implement AMOs at memory controllers, and can optimize away
fetching the original value when the destination is x0.

In some situations (not this particular case), that prevents confusion such as Zicsr's "For both CSRRS and CSRRC, if rs1=x0, then the instruction will not write to the CSR at all" (rs1=zero would introduce ambiguity between the register number and the register value).  Rather than sometimes calling it x0 (to prevent confusion) and sometimes calling it zero (when there is no confusion), it's best to consistently use the unambiguous name.  All the architecture documents I've seen call it x0 consistently.

xml/hwbp_registers.xml
pdonahue-ventana: I reduced the empty space.

timsifive (approved):
Thanks. github says there's a conflict in the PDF. Can you resolve that, so I can merge?

I think the simplest way is to merge `master` into this branch, resolving the PDF conflict in favor of `master`. But if you want to rewrite history so you never modified the PDF file in the first place that's also fine.

xml/hwbp_registers.xml
timsifive: If x0 is the standard, then I'm happy with a change to the macro to render it as x0. Optionally we could change the name of the macro too.

pdonahue-ventana:
Sorry.  I always build the file to make sure there are no errors, to eyeball it, etc.  I accidentally put it in one of the commits.  I'll fix it.

PR #564, opened 2020-10-14
Use official RISC-V terminology

pdonahue-ventana:
I had a conversation today with someone who was confused by "clobber" so I propose to use the official term.


ernie-sifive (approved):
Thanks, looks good!

timsifive (approved):
I'll merge this tomorrow, pending other comments.

PR #566, opened 2020-10-14
require polling of dmactive low as well as dmactive high transitions

benscotstaveley:
This proposed spec change would allow a DM time to deactivate gracefully by requiring a debugger to poll after writing 0 to dmcontrol.dmactive.  Currently the spec only states that a debugger should poll after writing 1 to this bit.

Prior discussion can be found in my post to the tech-debug mailing list on 2020.10.13, Subject "How to reliably inform debugger of completion of DM inactivation?".

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -237,6 +237,12 @@
         </field>
         <field name="dmactive" bits="0" access="R/W" reset="0">
             This bit serves as a reset signal for the Debug Module itself.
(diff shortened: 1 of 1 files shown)


ernie-sifive (approved):
Thanks, this change looks good to me!

timsifive (approved):
Looks good. I'll merge it tomorrow unless there are more comments.

This change is backwards-incompatible, but it's a trivial change for debuggers that allows some hardware implementations to be simplified a little.

Issue #559, opened 2020-10-06
Details of icount.count field


scottj97:
I'm thinking about what happens when an icount breakpoint exception occurs. If the handler does not touch the CSRs, is it acceptable for the trap to reoccur immediately after `mret`? If not, what mechanism prevents this?

The spec describes when `count` should be decremented, and "When count is decremented to 0, the trigger fires."

What if count is already 0? Is it decremented anyway, wrapping around to 0x3fff (or whatever platform-defined maximum)?

If the answer is that it should clamp to 0, what happens when a CSRW writes count=0? Should it take a trap immediately, or is the breakpoint effectively disabled? One could argue that the spec does not allow a trap since writing 0 directly is not "decremented to 0".

ernie-sifive:
I think the trigger occurs when count=1 and an instruction or trap occurs in one of the enabled modes.  Once the breakpoint trap is taken, count remains at 0 (or m/s/u stay at 0).  The spec doesn't say, but if your debugger mrets with count=0, I would think count would decrement from 0 to 0x3fff, etc., until it traps again.

If non-debugger code sets count to 0, I would say this should not trap (unless count was 1 just before).

scottj97:
When you say "count remains at 0" I assume that's because of `tcontrol.mte=0` (or `mstatus.mie=0`)?

I think there's a practical problem with this approach. Let's say I write count=2, m/s/u = 0/0/1, then mret to U-mode. The first U-mode instruction completes successfully, decrementing count to 1. The second instruction has a load page fault exception that goes to an M-mode handler. The count needs to decrement, since a trap was taken from U-mode, and the count is now 0, but we can't take the breakpoint exception because the page fault has set `tcontrol.mte=0`.

The count will stay at 0 until the mret -- but when/how does the breakpoint exception occur? 0x3fff instructions later? Not good.

If, instead, we took a breakpoint exception any time count was 0, the mret in this case would immediately trap back to the debugger, before running a single U-mode instruction and before decrementing (and wrapping) count, which seems like the most desirable behavior.

That has the drawback that writing count to 0 would also cause a trap when enabled. But maybe that's okay?

scottj97:
Is there any implementation of icount triggers that I could refer to? Or a software debugger implementation that can take advantage of icount-type triggers?

ernie-sifive:
Offhand, it seems like your approach would work.

timsifive:
Decrementing the count should never wrap.

[...]

PR #574, opened 2020-10-23
Add pending state/bit to icount.

timsifive:
See mailing list. Subject:"On icount triggers"

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -739,7 +739,7 @@
         </field>
         <field name="pending" bits="8" access="R/W" reset="0">
             This bit becomes set when \FcsrIcountCount is decremented from 1
-            to 0. It is cleared when the trigger fired.
+            to 0. It is cleared when the trigger fires.
         </field>
         <field name="s" bits="7" access="WARL" reset="0">
             When set, enable this trigger in S/HS-mode.
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -693,9 +693,8 @@
 
         If \FcsrIcountCount is not 0, then every instruction completed or
         trap taken from a mode where the trigger is enabled decrements
-        \FcsrIcountCount by 1, unless \FcsrIcountAction is 0, the current
-        mode is M-mode, and M-mode breakpoints are disabled via one of the
-        mechanisms described in Section~\ref{sec:mmtrigger}. When
+        \FcsrIcountCount by 1.
+        When
         \FcsrIcountCount is decremented from 1 to 0 then \FcsrIcountPending
         becomes set. When \FcsrIcountPending is set, the trigger fires just
         before any further instructions are executed in a mode where the
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -408,9 +408,13 @@
         <field name="0" bits="5" access="R" reset="0" />
         <field name="s" bits="4" access="WARL" reset="0">
             When set, enable this trigger in S/HS-mode.
+            This bit is hard-wired to 0 if the hart does not support
+            S-mode.
         </field>
         <field name="u" bits="3" access="WARL" reset="0">
             When set, enable this trigger in U-mode.
+            This bit is hard-wired to 0 if the hart does not support
+            U-mode.
         </field>
         <field name="execute" bits="2" access="WARL" reset="0">
             When set, the trigger fires on the virtual address or opcode of an
@@ -662,9 +666,13 @@
         <field name="0" bits="5" access="R" reset="0" />
         <field name="s" bits="4" access="WARL" reset="0">
             When set, enable this trigger in S/HS-mode.
+            This bit is hard-wired to 0 if the hart does not support
+            S-mode.
         </field>
         <field name="u" bits="3" access="WARL" reset="0">
             When set, enable this trigger in U-mode.
+            This bit is hard-wired to 0 if the hart does not support
+            U-mode.
         </field>
         <field name="execute" bits="2" access="WARL" reset="0">
             When set, the trigger fires on the virtual address or opcode of an
@@ -716,9 +724,13 @@
 
         <field name="vs" bits="26" access="WARL" reset="0">
             When set, enable this trigger in VS-mode.
+            This bit is hard-wired to 0 if the hart does not support
+            virtualization mode.
         </field>
         <field name="vu" bits="25" access="WARL" reset="0">
             When set, enable this trigger in VU-mode.
+            This bit is hard-wired to 0 if the hart does not support
+            virtualization mode.
         </field>
         <field name="hit" bits="24" access="WARL" reset="0">
             If this bit is implemented, the hardware sets it when this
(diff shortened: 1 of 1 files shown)


scottj97:
This was also discussed in #559

xml/hwbp_registers.xml
```diff
@@ -726,17 +732,17 @@
         </field>
 
         <field name="m" bits="9" access="WARL" reset="0">
-            When set, every instruction completed in or trap taken from
-            M-mode decrements \FcsrIcountCount by 1.
+            When set, enable this trigger in M-mode.
+        </field>
+        <field name="pending" bits="8" access="R/W" reset="0">
```
scottj97, line 740: Does this bit need to be writable?

xml/hwbp_registers.xml
```diff
@@ -726,17 +732,17 @@
...
+            This bit becomes set when \FcsrIcountCount is decremented from 1
+            to 0. It is cleared when the trigger fired.
```
scottj97, line 742: s/fired/fires/ (?)

xml/hwbp_registers.xml
```diff
@@ -726,17 +732,17 @@
...
         </field>
-        <field name="0" bits="8" access="R" reset="0" />
         <field name="s" bits="7" access="WARL" reset="0">
-            When set, every instruction completed in or trap taken from
-            S/HS-mode decrements \FcsrIcountCount by 1.
+            When set, enable this trigger in S/HS-mode.
```
scottj97, line 745: Is there somewhere else that explains that both committed instructions and traps should decrement the count? I can't find that in this icount section.

scottj97 (changes requested):

xml/hwbp_registers.xml
scottj97: Nevermind; I see this explained on line 686 in the merged file.

xml/hwbp_registers.xml
```diff
@@ -683,33 +683,42 @@
...
         <field name="vs" bits="26" access="WARL" reset="0">
-          When set, enable this trigger in VS-mode.
-          This bit is hard-wired to 0 if the hart does not support
-          virtualization mode.
+            When set, enable this trigger in VS-mode.
         </field>
         <field name="vu" bits="25" access="WARL" reset="0">
-          When set, enable this trigger in VU-mode.
-          This bit is hard-wired to 0 if the hart does not support
-          virtualization mode.
```
pdonahue-ventana, line 712: Removing this will make these bits consistent with `u` and `s` which are also optional modes.  However, the spec is not consistent across registers.  For instance, `dcsr` says that `ebreaku` is hardwired to 0 if U mode isn't supported but none of the trigger registers say the same about the `u` bit.  All of the `vs` and `vu` bits in the other registers are inconsistent with `s` and `u` in those same registers (which is my fault) and this change only fixes one.

My preference would be to leave these sentences in and add similar sentences to all the existing `u` and `s` bits so there is no confusion about legal values on these WARL fields.

xml/hwbp_registers.xml
timsifive: My thought was that an M-mode debugger might want to clear the bit for some reason, and surely somebody might have a clever use for setting it. I don't think it costs the hardware anything to make it writable given that it already exists, so I made it writable. If I'm wrong about hardware and making the bit writable makes implementations notably more complicated, I'm fine with it being read-only.

xml/hwbp_registers.xml
timsifive: Fixed.

xml/hwbp_registers.xml
timsifive: I would like not to have to mention for every feature that if the feature doesn't exist, the bits are wired to 0. That's the default for (I think) every bit we define anywhere in this spec. We already have some language that talks about the fact that all triggers are WARL at https://github.com/riscv/riscv-debug-spec/blob/master/trigger.tex#L176. Is there a place where we can make a general statement, instead of replacing the same boilerplate over and over?

I guess I don't feel super strongly about this. It just feels like we're repeatedly stating the obvious. But maybe that's better than for somebody (who likely did not read the entire document) to be unsure.

pdonahue-ventana: The privileged spec explicitly describes what happens for each bit if it's affected by a lack of S-mode or U-mode, etc.  Personally, I never take for granted that people will have enough understanding to do what I think is obvious.  On the other hand, what I think is obvious is sometimes not the correct thing to do because I am the one who lacks understanding.  Making things explicit helps get everyone "on the same page" at the expense of making the spec have an extra page or two.

xml/hwbp_registers.xml
```diff
@@ -683,33 +683,42 @@
...
-        If the hardware limits \FcsrIcountCount to 1, and changes mode bits instead of
-        decrementing \FcsrIcountCount, this register can be implemented with just 2
-        bits. One for \FcsrIcountU, and one for \FcsrIcountM and \FcsrIcountS tied together.  If
-        only the external debugger or only a software monitor needs to be
-        supported, a single bit is enough.
+        If \FcsrIcountCount is not 0, then every instruction completed or
+        trap taken from a mode where the trigger is enabled decrements
+        \FcsrIcountCount by 1, unless \FcsrIcountAction is 0, the current
+        mode is M-mode, and M-mode breakpoints are disabled via one of the
+        mechanisms described in Section~\ref{sec:mmtrigger}. When
```
scottj97, line 690: After this change introducing the pending bit, I wonder if this is even necessary anymore. I had suggested this behavior since otherwise the breakpoint exception would get lost when the count reached 0 and exceptions were blocked. But now with the pending bit, that exception would not get lost anymore.

xml/hwbp_registers.xml
timsifive: I'm making it explicit.

xml/hwbp_registers.xml
timsifive: I don't quite follow you. I thought pending is orthogonal to the mechanisms that might disable triggers in M-mode. Can you explain a bit more what you mean?

scottj97: This change (to stop count decrementing when exceptions were blocked) was #555. I opened that issue because at the time, if count decremented to 0 and breakpoint exceptions were blocked, the breakpoint would get dropped on the floor and never taken. Stopping the decrementing when exceptions were blocked was a way to avoid dropping that breakpoint exception completely.

But now, with this new pending bit, if the count still decrements even when exceptions are blocked, the pending bit will turn on and stay on until the exception can be taken, so it will no longer get dropped on the floor.

So I think #555 is moot and should be undone.

timsifive: I agree. There is maybe some use to disabling counting while in M-mode, but it makes the language more cumbersome and presumably the implementation as well. For the single step case #555 is no longer needed, and I cannot really imagine why somebody would want to count M-mode instructions except the ones executed while M-mode triggers are enabled.

scottj97 (approved):

ernie-sifive (approved):

PR #575, opened 2020-10-24
Triggers affect harts, not the system.

pdonahue-ventana:
For instance, there may be heterogeneous systems with some control processor harts that only support M mode and some full-featured harts that support more.


ernie-sifive (approved):
Seems right to me...

PR #576, opened 2020-10-24
Remove all uses of the ambiguous term "may not"

pdonahue-ventana:
Here's my attempt at removing the ambiguous term.  Double-check that I didn't interpret any of these incorrectly.


debug_module.tex
```diff
@@ -108,7 +108,7 @@ \section{Reset Control} \label{reset}
 These bits must be set regardless of the cause of the reset.
 The {\tt havereset} bits for the selected harts
 can be cleared by writing 1 to \FdmDmcontrolAckhavereset in \RdmDmcontrol.
-The {\tt havereset} bits may or may not be cleared
+The {\tt havereset} bits may be cleared
```
ernie-sifive, line 111: To me, the "may or may not" language seems clearer. Perhaps "might" or "might or might not"?

xml/dm_registers.xml
```diff
@@ -592,7 +592,7 @@
         If resume groups are not implemented, then \FdmDmcsTwoGrouptype will
         remain 0 even after 1 is written there.
 
-        The external triggers available to add to halt groups may or may not
+        The external triggers available to add to halt groups may
         be distinct from the external triggers available to add to resume groups.
```
ernie-sifive, line 596: How about "... halt groups may be the same as or distinct from the external triggers..."

ernie-sifive (commented):
Thanks, Paul, this does help a lot!

trigger.tex
```diff
@@ -10,7 +10,7 @@ \chapter{Trigger Module}
 
 A hart can be compliant with this specification without implementing any
 trigger functionality at all, but if it is implemented then it must conform to
-this section. If triggers aren't implemented, the CSRs may not exist at all and
+this section. If triggers aren't implemented, the CSRs must not exist at all and
```
timsifive, line 13: This was intended as "might" although I'm not sure why you'd want CSRs implemented if they're all tied to 0 (or whatever).

timsifive (approved):
Looks OK to me, although Ernie's comments for extra clarity are probably justified.

debug_module.tex
pdonahue-ventana: OK.  I was going with IEEE terminology but that's basically not followed anywhere else so there's no point in doing that here.  I just pushed a change that follows your suggestion.

xml/dm_registers.xml
pdonahue-ventana: OK.  Done.

trigger.tex
pdonahue-ventana: I'm glad I made this PR since it'll allow us to resolve this ambiguity.

In the privileged architecture, it says:
> Attempts to access a non-existent CSR raise an illegal instruction exception.

The reason for that is to provide a discovery mechanism.  Step 1 of the enumeration recipe in section 5.1 of the debug spec relies on that.  I think that we're all in agreement on that.

However, your point is that it could alternatively hard-wire all of the CSRs to 0 (or maybe just hard-wire tselect!=0 and/or tdata1.type=0 and/or tinfo.info=1).  That would cause the first trigger to exit the loop at the various other exit points.  That would give the same result as exiting in step 1.

I don't see a lot of value in having two ways of doing it but there may be other things I'm not considering.  Backward-compatibility?  The decode logic (which handles illegal instructions) may not know if there's a trigger module attached?

The options are "might not" (allowing both ways of doing things) or "must not" (allowing only the trapping form of discovery) or "should not" (allowing both but telling implementations to prefer the trapping mechanism).

pdonahue-ventana: I see now that tselect says:
> It is optional if no triggers are implemented.

That means that we can't say "must not" unless we want to change the requirements (and there's no strong argument for that).  That argues for "might not" or possibly "should not".  Just keep it simple and go with "might not"?

pdonahue-ventana: I just went with "might not" as the easiest solution.  I think that they should not exist but if software follows the recipe then it doesn't really matter and maybe somebody can save a couple of gates in the decoder or something.

timsifive (approved):
I'll merge this tomorrow, unless there's more discussion.

timsifive:
I'll merge tomorrow, unless there are more comments.

PR #577, opened 2020-10-24
Use consistent term when referring to impebreak

ernie-sifive:
Most references are "implicit ebreak" but there is one instance of "implied ebreak".  Make all of them "implicit".


timsifive (approved):
Thanks!

PR #578, opened 2020-10-26
textra32/64 also affects the new mcontrol6 triggers

pdonahue-ventana:
Forgot this in #538


timsifive (approved):
Thanks.

pdonahue-ventana:
I always found textra64 to be confusing since it referred to textra instead of textra32.  I just pushed another commit that, in my judgement, makes things clearer.

PR #579, opened 2020-10-28
List all DM registers in Table 3.8.

timsifive:
But don't give them each a description if it's redundant.

The main change here is that registers.py now generates an "index" table
that can be split across pages.

This change supersedes #560.

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -962,9 +962,9 @@
     </register>
 
     <register name="Custom Features 0" short="custom0" address="0x70">
-        The optional {\tt custom} registers may be used for non-standard
-        features. Future version of the debug spec will not use these
-        addresses.
+        The optional \RdmCustomZero through \RdmCustomFifteen registers may
+        be used for non-standard features. Future versions of the debug spec
+        will not use these addresses.
     </register>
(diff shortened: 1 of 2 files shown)


timsifive:
@pdonahue-ventana, what do you think?

xml/dm_registers.xml
```diff
@@ -966,13 +962,24 @@
...
-        version of the debug spec will not use this address.
-    </register>
-
-    <register name="Custom Features 15" short="custom15" address="0x7f">
-        This optional register may be used for non-standard features. Future
-        version of the debug spec will not use this address.
-    </register>
+        The optional {\tt custom} registers may be used for non-standard
+        features. Future version of the debug spec will not use these
+        addresses.
```
ernie-sifive, line 967: To be consistent with the data0-11 and progbuf0-15 register blocks, should this say something like "The optional \RdmCustomZero through \RdmCustomFifteen may be used..."

pdonahue-ventana: I agree.  Also, "Future versions of the debug spec" instead of "version".  Otherwise it looks great.

timsifive: Done. Thanks.

ernie-sifive (approved):

PR #585, opened 2020-11-03
Change quick access exceptions to halt the target.

### LINKED ISSUE #462: exceptions in quick access
From Marc Gauthier:
The sequence shows that if any program buffer instruction takes an exception, the hart resumes execution immediately.  This is generally fatal to the program being debugged, as it will corrupt its state.  Most program buffer sequences save some registers, do an operation, and restore the registers (very little can be done otherwise).  If the operation hits an exception and the hart resumes immediately, registers are not restored, resulting in program state corruption.  This effectively means the operation must not take an exception.  But if it must not take an exception anyway – leaving the hart halted is much better than corrupting its state, no matter how bad halting may be.  Also note that not being able to take an exception is particularly limiting:  it means the debugger cannot offer the user arbitrary access to memory, such as to browse virtual address spaces etc, because the debugger has no way to know whether an access at a particular address will succeed.  Attempting it anyway and giving the responsibility to the user is fraught with peril:  failure is silently fatal (the worst kind of bug, especially when introduced by a debugger).

At least if the hart stays halted, the debugger can clean things up and resume correct execution.

A clean and non-intrusive solution is an optional proper trap mechanism inside the program buffer:  a way to designate where in the program buffer to jump to if an exception occurs.  Or if the implementation supports branches within the program buffer, start execution at the 2nd program buffer word, and jump to the 1st program buffer word on exception; the instruction at the first word can branch to the  appropriate restore sequence.  And either way, if a second exception occurs, leave the hart halted.

timsifive:
Fixes #462.

### DIFF
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -129,9 +129,12 @@
         \item If the hart is halted, the command  sets \FdmAbstractcsCmderr to ``halt/resume'' and does not continue.
         \item Halt the hart. If the hart halts for some other reason (e.g. breakpoint), the command
             sets \FdmAbstractcsCmderr to ``halt/resume'' and does not continue.
-        \item Execute the Program Buffer. If an exception occurs, \FdmAbstractcsCmderr is set to ``exception''
-            and the program buffer execution ends, but the quick access command continues.
-        \item Resume the hart.
+        \item Execute the Program Buffer.
+            If an exception occurs, \FdmAbstractcsCmderr is set to
+            ``exception,'' the Program Buffer execution ends, and the hart is
+            halted with \FcsrDcsrCause set to 3.
+        \item If the Program Buffer executed without an exception, then
+            resume the hart.
         \end{steps}
 
         Implementing this command is optional.


ernie-sifive (approved):

xml/abstract_commands.xml
```diff
@@ -129,9 +129,12 @@
         \item If the hart is halted, the command  sets \FdmAbstractcsCmderr to ``halt/resume'' and does not continue.
         \item Halt the hart. If the hart halts for some other reason (e.g. breakpoint), the command
             sets \FdmAbstractcsCmderr to ``halt/resume'' and does not continue.
-        \item Execute the Program Buffer. If an exception occurs, \FdmAbstractcsCmderr is set to ``exception''
-            and the program buffer execution ends, but the quick access command continues.
-        \item Resume the hart.
+        \item Execute the Program Buffer.
+            If an exception occurs, \FdmAbstractcsCmderr is set to
+            ``exception,'' the Program Buffer execution ends, and the hart is
+            halted with \FcsrDcsrCause set to 3.
```
pdonahue-ventana, line 135: dcsr.cause=3 is "The debugger requested entry to Debug Mode using haltreq."  That doesn't seem to accurately reflect what happened.  On the other hand, there's not a lot of free encoding space for a new cause.  What are your thoughts?

timsifive: Yes, nothing really fits. I just realized I should mention that it also sets cmderr and then there is no ambiguity why the halt occurred.

timsifive: Oh, it already says that. And right there, no less.

I think this isn't ideal, but I don't want to use another dcsr.cause encoding (we have 2 left) for such a rare case.

pdonahue-ventana: I agree that it's not ideal but it's probably fine.  The debugger ought to know whether it has been doing quick accesses and/or haltreq.  If the hart is halted with cause=3 and cmderr=1 then either:
- it made a haltreq and something else went wrong before or after the halt
- it did a quick access and got an exception

If the debugger wants to do a quick access and then a haltreq, I think it can disambiguate the two scenarios as long as it reads cmderr in between.

PR #588, opened 2020-11-05
Added sbytemask field to textra32 and textra64

bruceable:
sbytemask field in textra32 and textra64 provides byte-granular comparison of the svalue trigger value to the scontext CSR

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -1036,7 +1036,7 @@
 
         <field name="0" bits="47:41" access="R" reset="0" />
 
-        <field name="sbytemask" bits="41:36" access="WARL" reset="0">
+        <field name="sbytemask" bits="40:36" access="WARL" reset="0">
             0=compare, 1=don't compare, i.e byte compare is masked out
             bit 36 = first (low) byte compare of svalue (bits [7:0])
             bit 37 = second byte compare of svalue (bits [15:8])
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -967,12 +967,12 @@
         tie any number of upper bits to 0. The $|select|$ bits may only support
         0 (ignore).
 
-        Byte-granular comparison of the \RcsrScontext to \FcsrTextraThirtytwoSvalue
+        Byte-granular comparison of \RcsrScontext to \FcsrTextraThirtytwoSvalue
         allows \RcsrScontext to be defined to include more than one element of comparison.
         For example, software instrumentation can program the \RcsrScontext value to be
         the concatenation of different ID contexts such as process ID and thread ID.
-        The user can then program byte compares to include one or more of the contexts
-        in the compare. For \RcsrTextraThirtytwo there are two byte-compare mask bits.
+        The user can then program byte compares based on \FcsrTextraThritytwoSbytemask
+        to include one or more of the contexts in the compare.
         
         Byte masking only applies to \RcsrScontext comparison; i.e when \FcsrTextraThirtytwoSselect is 1.
         
@@ -1002,7 +1002,7 @@
             When the least significant bit of this field is 1, it causes bits 7:0
             in the comparison to be ignored, when \FcsrTextraThirtytwoSselect=1.
             When the next most significant bit of this field is 1, it causes bits 15:8
-            to be ignored in the comparison, when \FcsrTextraThirtytwoSselect=1
+            to be ignored in the comparison, when \FcsrTextraThirtytwoSselect=1.
         </field>
 
         <field name="svalue" bits="17:2" access="WARL" reset="0">
@@ -1030,13 +1030,12 @@
         5, or 6 and XLEN=64. The fields are defined
         above, in \RcsrTextraThirtytwo.
 
-        Byte-granular comparison of the \RcsrScontext to \FcsrTextraSixtyfourSvalue in
+        Byte-granular comparison of \RcsrScontext to \FcsrTextraSixtyfourSvalue in
         \RcsrTextraSixtyfour allows \RcsrScontext to be defined to include
         more than one element of comparison.  For example, software instrumentation
         can program the \RcsrScontext value to be the concatenation of different ID contexts
-        such as process ID and thread ID.  The user can then program byte compares
-        to include one or more of the contexts in the compare. For \RcsrTextraSixtyfour
-        there are 5 byte-compare mask bits.
+        such as process ID and thread ID.  The user can then program byte compares based on
+        \FcsrTextraSixtyfourSbytemask to include one or more of the contexts in the compare.
         
         Byte masking only applies to \RcsrScontext comparison; i.e when \FcsrTextraSixtyfourSselect is 1.
         
@@ -1050,14 +1049,10 @@
         <field name="sbytemask" bits="40:36" access="WARL" reset="0">
             When the least significant bit of this field is 1, it causes bits 7:0
             in the comparison to be ignored, when \FcsrTextraSixtyfourSselect=1.
-            When the next most significant bit of this field is 1, it causes bits 15:8
-            to be ignored in the comparison, when \FcsrTextraSixtyfourSselect=1
-            When the next most significant bit of this field is 1, it causes bits 23:16
-            to be ignored in the comparison, when \FcsrTextraSixtyfourSselect=1
-            When the next most significant bit of this field is 1, it causes bits 31:24
-            to be ignored in the comparison, when \FcsrTextraSixtyfourSselect=1
-            When the next most significant bit of this field is 1, it causes bits 33:32
-            to be ignored in the comparison, when \FcsrTextraSixtyfourSselect=1
+            Likewise, the second bit controls the comparison of bits 15:8,
+            third bit controls the comparison of bits 23:16,
+            fourth bit controls the comparison of bits 31:24, and
+            fifth bit controls the comparison of bits 33:32.
         </field>
         <field name="svalue" bits="35:2" access="WARL" reset="0">
         </field>
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -974,6 +974,7 @@
         such as process ID and thread ID.  The user can then program byte compares of
         the trigger to include one or more of the contexts in the compare. For textra32
         there are two byte-compare mask bits; for textra64 there are 5 byte-compare mask bits.
+        Byte masking only applies to scontext comparison; i.e when sselect==1.
         
         <field name="mhvalue" bits="31:26" access="WARL" reset="0">
             Data used together with \FcsrTextraThirtytwoMhselect.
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -1019,8 +1034,16 @@
...
+        <field name="0" bits="47:41" access="R" reset="0" />
 
+        <field name="sbytemask" bits="41:36" access="WARL" reset="0">
+            0=compare, 1=don't compare, i.e byte compare is masked out
+            bit 36 = first (low) byte compare of svalue (bits [7:0])
+            bit 37 = second byte compare of svalue (bits [15:8])
+            bit 38 = third byte compare of svalue (bits [23:16])
+            bit 39 = fourth byte compare of svalue (bits [31:24])
+            bit 40 = fifth byte compare of svalue (bits [33:32])
+        </field>
```
ernie-sifive, line 1046:
```suggestion
        <field name="sbytemask" bits="40:36" access="WARL" reset="0">
            0=compare, 1=don't compare, i.e byte compare is masked out
            bit 36 = first (low) byte compare of svalue (bits [7:0])
            bit 37 = second byte compare of svalue (bits [15:8])
            bit 38 = third byte compare of svalue (bits [23:16])
            bit 39 = fourth byte compare of svalue (bits [31:24])
            bit 40 = fifth byte compare of svalue (bits [33:32])
        </field>
```

xml/hwbp_registers.xml
```diff
@@ -967,6 +967,14 @@
         tie any number of upper bits to 0. The $|select|$ bits may only support
         0 (ignore).
 
+        Byte-granular comparison of the scontext CSR to the trigger's svalue in
+        textra32 and textra64 registers allows scontext to be defined to include
+        more than one element of comparison.  For example, software instrumentation
+        can program the scontext value to be the concatenation of different ID contexts
+        such as process ID and thread ID.  The user can then program byte compares of
+        the trigger to include one or more of the contexts in the compare. For textra32
+        there are two byte-compare mask bits; for textra64 there are 5 byte-compare mask bits.
```
ernie-sifive, line 976: What about when sselect==2 (ASID comparison)?  Would the byte mask apply in that case?  This text only describes scontext comparison and needs to explicitly state whether the masking applies to ASID.

xml/hwbp_registers.xml
```diff
@@ -987,10 +995,17 @@
...
+        <field name="sbytemask" bits="19:18" access="WARL" reset="0">
+            0=compare, 1=don't compare byte, i.e. byte is masked out
+            bit 18 = lower byte compare of svalue (bits [7:0])
+            bit 19 = upper byte compare of svalue (bits [15:8])
+        </field>
 
         <field name="svalue" bits="17:2" access="WARL" reset="0">
-            Data used together with \FcsrTextraThirtytwoSselect.
+            Data used together with \FcsrTextraThirtytwoSselect to compare with the scontext CSR
+            which is ANDed with all other trigger comparisons.
```
ernie-sifive, line 1008: I would suggest leaving this text unchanged since svalue can compare to either scontext or ASID depending on sselect.  The text already makes it clear (I think) that all comparisons associated with a trigger must all match (ANDed) for the trigger to match.

ernie-sifive (approved):
Looks good to me.

xml/hwbp_registers.xml
```diff
@@ -987,7 +996,16 @@
             If the H extension is not supported, the only legal values are 0 and 4.
         </field>
 
-        <field name="0" bits="22:18" access="R" reset="0" />
+        <field name="0" bits="22:20" access="R" reset="0" />
+        
+        <field name="sbytemask" bits="19:18" access="WARL" reset="0">
+            0=compare, 1=don't compare byte, i.e. byte is masked out
+            bit 18 = lower byte compare of svalue (bits [7:0])
+            bit 19 = upper byte compare of svalue (bits [15:8])
```
timsifive, line 1004: We use capitalization and punctuation in field descriptions. When mentioning the behavior of values we use : instead of = (see e.g. https://github.com/riscv/riscv-debug-spec/blob/master/xml/dm_registers.xml#L12).

I don't think we have any examples yet of where each bit inside a field has a different meaning. The closest is hawindow, where we say "I.e. bit 0 refers to hart $\RdmHawindowsel * 32$, while bit 31 refers to hart $\RdmHawindowsel * 32 + 31$." (https://github.com/riscv/riscv-debug-spec/blob/master/xml/dm_registers.xml#L347).

Putting that all together, I think we want language like:
When the least-significant bit of this field is 1, it causes bits 7:0 in the comparison to be ignored when \FcsrTextraThirtytwoSselect=1. When the next most significant bit of this field is 1, causes bits 15:8 to be ignored in the comparison when \FcsrTextraThirtytwoSselect=1, and so on.

xml/hwbp_registers.xml
```diff
@@ -1019,8 +1037,19 @@
         <field name="mhselect" bits="50:48" access="WARL" reset="0">
         </field>
 
-        <field name="0" bits="47:36" access="R" reset="0" />
+        <field name="0" bits="47:41" access="R" reset="0" />
 
+        <field name="sbytemask" bits="40:36" access="WARL" reset="0">
+            0=compare, 1=don't compare, i.e byte compare is masked out
```
timsifive, line 1043: Same as above.

xml/hwbp_registers.xml
```diff
@@ -987,7 +996,16 @@
...
-        <field name="0" bits="22:18" access="R" reset="0" />
+        <field name="0" bits="22:20" access="R" reset="0" />
+        
+        <field name="sbytemask" bits="19:18" access="WARL" reset="0">
+            0=compare, 1=don't compare byte, i.e. byte is masked out
+            bit 18 = lower byte compare of svalue (bits [7:0])
+            bit 19 = upper byte compare of svalue (bits [15:8])
+            
+            Byte masking of svalue only applies to scontext comparison; i.e.
+            when sselect==1.
```
timsifive, line 1007: I think I mentioned that adequately in my suggestion enough. If not, use \FcsrTextraThirtytwoSselect=1 instead of sselect==1.

timsifive (changes requested):
The feature looks good. I have a bunch of comments about formatting etc, however.

xml/hwbp_registers.xml
```diff
@@ -967,6 +967,15 @@
         tie any number of upper bits to 0. The $|select|$ bits may only support
         0 (ignore).
 
+        Byte-granular comparison of the \RcsrScontext to \FcsrTextraThirtytwoSvalue
```
timsifive, line 970: Just "\RcsrScontext" instead of "the \RcsrScontext"

xml/hwbp_registers.xml
```diff
@@ -967,6 +967,15 @@
...
+        allows \RcsrScontext to be defined to include more than one element of comparison.
+        For example, software instrumentation can program the \RcsrScontext value to be
+        the concatenation of different ID contexts such as process ID and thread ID.
+        The user can then program byte compares to include one or more of the contexts
+        in the compare. For \RcsrTextraThirtytwo there are two byte-compare mask bits.
```
timsifive, line 975: No need to talk explicitly about textra32 because this paragraph is in the section that defines textra32, and the size is already clear from the figure that is placed directly below this block of text.

xml/hwbp_registers.xml
timsifive: This paragraph should refer to \FcsrTextraThirtytwoSbytemask since that is what implements the comparison.

xml/hwbp_registers.xml
```diff
@@ -987,7 +996,14 @@
...
         </field>
 
-        <field name="0" bits="22:18" access="R" reset="0" />
+        <field name="0" bits="22:20" access="R" reset="0" />
+        
+        <field name="sbytemask" bits="19:18" access="WARL" reset="0">
+            When the least significant bit of this field is 1, it causes bits 7:0
+            in the comparison to be ignored, when \FcsrTextraThirtytwoSselect=1.
+            When the next most significant bit of this field is 1, it causes bits 15:8
+            to be ignored in the comparison, when \FcsrTextraThirtytwoSselect=1
```
timsifive, line 1005: Needs a period at the end of the sentence.

xml/hwbp_registers.xml
```diff
@@ -1014,13 +1030,35 @@
         5, or 6 and XLEN=64. The fields are defined
         above, in \RcsrTextraThirtytwo.
 
+        Byte-granular comparison of the \RcsrScontext to \FcsrTextraSixtyfourSvalue in
```
timsifive, line 1033: Same feedback as textra32 version.

xml/hwbp_registers.xml
```diff
@@ -1014,13 +1030,35 @@
...
         </field>
 
-        <field name="0" bits="47:36" access="R" reset="0" />
+        <field name="0" bits="47:41" access="R" reset="0" />
 
+        <field name="sbytemask" bits="40:36" access="WARL" reset="0">
+            When the least significant bit of this field is 1, it causes bits 7:0
+            in the comparison to be ignored, when \FcsrTextraSixtyfourSselect=1.
+            When the next most significant bit of this field is 1, it causes bits 15:8
+            to be ignored in the comparison, when \FcsrTextraSixtyfourSselect=1
```
timsifive, line 1054: Need a period at the end of these sentences.

xml/hwbp_registers.xml
```diff
@@ -1014,13 +1030,35 @@
...
+            When the next most significant bit of this field is 1, it causes bits 23:16
+            to be ignored in the comparison, when \FcsrTextraSixtyfourSselect=1
+            When the next most significant bit of this field is 1, it causes bits 31:24
+            to be ignored in the comparison, when \FcsrTextraSixtyfourSselect=1
+            When the next most significant bit of this field is 1, it causes bits 33:32
+            to be ignored in the comparison, when \FcsrTextraSixtyfourSselect=1
```
timsifive, line 1060: At some point we'll have to find a nicer way to write this, but I don't have any good ideas right now. Probably can merge like this and fix it later.

timsifive (changes requested):

xml/hwbp_registers.xml
```diff
@@ -967,6 +967,15 @@
         tie any number of upper bits to 0. The $|select|$ bits may only support
         0 (ignore).
 
+        Byte-granular comparison of \RcsrScontext to \FcsrTextraThirtytwoSvalue
+        allows \RcsrScontext to be defined to include more than one element of comparison.
+        For example, software instrumentation can program the \RcsrScontext value to be
+        the concatenation of different ID contexts such as process ID and thread ID.
+        The user can then program byte compares based on \FcsrTextraThritytwoSbytemask
```
timsifive, line 974: Typo: FcsrTextraThritytwoSbytemask -> FcsrTextraThirtytwoSbytemask

timsifive (changes requested):
Looks good, except for the typo.

I think it still won't pass the checks even with this fix because it depends on #591. You can fix this in your PR by merging latest master into your bytemask branch, but I'm OK skipping that step and just merging this once the typo is fixed.

timsifive (approved):
I'll merge this tomorrow, unless there is more discussion.

PR #589, opened 2020-11-07
Fix #587: Mention vs and vu in count field

### LINKED ISSUE #587: icount breakpoints clearing modes
For the `count` field when `tdata1` is type 3, the specification says:

> When count is decremented to 0, the trigger fires. Instead of changing count from 1 to 0, it is also acceptable for **hardware to clear m, s, and u**. This allows count to be hard-wired to 1 if this register just exists for single step

If the Hypervisor is implemented, I think vs and vu must be cleared also.

pdonahue-ventana:


pdonahue-ventana:
> Can you add this description: "Mention vs and vu in count field."

Done

timsifive (approved):

PR #592, opened 2020-11-10
Add keepalive feature.

timsifive:
Addresses #519. Went with the simple version without reading the keepalive values.

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -22,7 +22,14 @@
 
             This must be 1 when \FdmAbstractcsProgbufsize is 1.
         </field>
-        <field name="0" bits="21:20" access="R" reset="0" />
+        <field name="allkeepalive" bits="21" access="R" reset="0">
+            This field is 1 when all currently selected harts have
+            \Fkeepalive set.
+        </field>
+        <field name="anykeepalive" bits="20" access="R" reset="0">
+            This field is 1 when at least one currently selected harts has
+            \Fkeepalive set.
+        </field>
         <field name="allhavereset" bits="19" access="R" reset="-">
             This field is 1 when all currently selected harts have been reset
             and reset has not been acknowledged for any of them.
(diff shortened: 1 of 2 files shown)


xml/dm_registers.xml
```diff
@@ -222,7 +231,19 @@
             The high 10 bits of \Fhartsel: the DM-specific index of the hart to
             select. This hart is always part of the currently selected harts.
         </field>
-        <field name="0" bits="5:4" access="R" reset="0" />
+        <field name="setkeepalive" bits="5" access="W1" reset="-">
+            This optional field writes \Fkeepalive for all currently selected
+            harts, unless \FdmDmcontrolClrkeepalive is simultaneously set to
+            1.
```
ernie-sifive, line 237:
```suggestion
            This optional field sets \Fkeepalive for all currently selected
            harts, unless \FdmDmcontrolClrkeepalive is simultaneously set to
            1.
```

ernie-sifive (approved):
I think this simpler approach is the right thing to do.  Just one minor nit...

timsifive:
@pdonahue-ventana does this look OK?

pdonahue-ventana:
@RobertChyla: You made the original request so what are your thoughts?

This allows a debugger to request that harts should stay available but there's no feedback about whether this feature is implemented or if it will be respected.  As such, I think that debuggers are not simplified since they'll still have to deal with things going unavailable unexpectedly.  At least now with stickyunavail they know that this happened.  But it's a nice feature to at least allow the debugger to make the request.

I think that some sort of discovery mechanism would be an improvement to at least know if the DM supports making the request (and whether the power controller will honor it may be beyond the scope of the DM spec).

timsifive:
I contacted Robert by e-mail, and he suggested the slight rewording, and said "it is OK as it is now."

Even if the feature is supported, some hardware will not be able to guarantee that the hart will remain available, so debuggers will still need to deal with harts disappearing. If we had a config structure I'd just say stick some bits in there that describe possible behavior, although that still doesn't handle the case where the DM itself doesn't even know whether the system it is in can honor the request.

Seeing how we're freezing on Monday, I'd like to merge this as-is, and you can open an issue for 1.1 to address discoverability. Thoughts?

RobertChyla:
I was surprised that my innocent email paragraph end-up as spec change.
I did not wanted to overload with allowing discovery nor 'confirmation' that setting is effective as this can be really outside of DM. This may be better handled by Config Structure (for DM and/or Power Controller).

pdonahue-ventana:
> Seeing how we're freezing on Monday, I'd like to merge this as-is, and you can open an issue for 1.1 to address discoverability. Thoughts?

OK

Issue #550, opened 2020-09-14
Interaction of mcontrol.hit with mcontrol.chain?


scottj97:
If a trigger with `mcontrol.chain=1` matches, but the next trigger does not match, no exception happens.

Should the first trigger's `mcontrol.hit` bit be set in that case?

ernie-sifive:
Yes, I believe so.  The spec says "If this bit is implemented, the hardware sets it when this trigger matches."  When a chain is configured, the triggers compute their own match result independently, then if all match, the action of the last trigger is performed.

scottj97:
I suppose that's clear enough. No spec change needed.

scottj97:
There's a similar issue with the opposite case: the first trigger in the chain does not match, but the second one would. The description of chain=1 says "While this trigger does not match, it prevents the trigger with the next index from matching."

Combined with the description of hit ("If this bit is implemented, the hardware sets it when this trigger matches") suggests that the second trigger should not set its hit bit, even if it would have absent the chain.

But that's a little different than what @ernie-sifive said above.

ernie-sifive:
Your interpretation seems valid.  The text in the Spec suggests that the `hit` bit of the second trigger in a chain should not indicate a match if the first trigger did not match.

scottj97:
Since there's a lot of good discussion around the debug spec lately, I'd like to revisit this issue.

The behavior described above ("temperature coding") is rather vague in the spec, and I wonder if it is actually deliberate (and desirable?) or just an accident.

Would it be acceptable, or could we modify the spec to make it acceptable, for the chained triggers to set all their hit bits at once, only when the entire chain matches?

timsifive:
This could lead to trouble. If hit is set even when the entire chain doesn't match, you could end up with all the hit bits being set even if the entire chain never matched. Now the debugger cannot use the hit bit to see which trigger hit.

scottj97:
@timsifive that's not how I interpreted it. Each trigger in the chain that does not match prevents the next from matching (recursively). The result is a temperature coding where the first 0-N hit bits will end up set. All N will be set if the entire chain matched and a trigger exception was taken. Otherwise, only the first one(s) will possibly be set, which could allow the debugger to see how "close" it came to triggering the breakpoint exception.

This is possibly useful, but seems accidental, which is why I want to have the discussion.

From my selfish hardware design perspective, I would prefer that all hit bits are set only them the entire chain matches; otherwise none of the hit bits in the chain get set.

[...]

PR #593, opened 2020-11-10
hit must be set on fire, may be set on match

timsifive:
This helps some hardware implementations be simpler.  Addresses #550.


ernie-sifive (approved):

scottj97 (approved):
Many thanks

Issue #570, opened 2020-10-15
Clarify trigger priority for AMO instructions


JamesKenneyImperas:
Table 5.2 gives the priority of triggers with respect to standard exceptions. Could you please clarify how this works with AMO instructions other than `amoswap`?

For such instructions, the value stored by the instruction is not known until after the load. This means that a _mcontrol/mcontrol6 store data before trigger_ cannot fire at the indicated point but must happen later. I think the table needs to be enhanced to discuss this case, and also to explicitly state whether `amoswap` behaves differently to other AMO instructions in this respect.

Thanks.

ernie-sifive:
I think triggers on AMO data value, whether load or store, typically have timing=1 (after) if a system can support AMO store data value triggers at all.  The AMO operation may be performed in the memory system so the core may not ever see the store value.

Assuming the trigger unit sees both the load and store value, then with timing=1 the priority is clear from Table 5.2 -- the mcontrol exception is taken as the highest priority.

Supporting timing=0 with select=1 is complicated so not typically supported.  Store data value is uncomplicated as long as the system doesn't support the A extension.  Otherwise, if the breakpoint/debug exception ends up being taken, then the hardware must revert the system to the state it was in before the triggering instruction, including restoring any changed memory contents and registers.  If the spec needs to address such systems, we could update Table 5.2 to replace "mcontrol store data before" with "mcontrol non-AMO store data before" and add "mcontrol AMO store data" at the same priority as "mcontrol load data before".

Ernie

JamesKenneyImperas:
Thanks Ernie. I think text based on your response here should be included in the specification near table 5.2, especially the point that _triggers may not be supported for AMO operations at all_ - as far as I can see, AMO operation constraints are not discussed anywhere in the specification, and this came as a complete surprise to me. From my naive viewpoint, setting a breakpoint on a lock operation sounds like something that users might regularly attempt to do, so it should be made very clear that this might not be possible.

pdonahue-ventana:
What are your thoughts on https://github.com/riscv/riscv-debug-spec/pull/561?

JamesKenneyImperas:
I'm not an expert on debug hardware at all, but that change looks to be starting to address this issue, when combined with Ernie's comment above.

[...]

PR #595, opened 2020-11-12
AMO operations may be ignored by mcontrol triggers.

timsifive:
Also some formatting cleanup to make things more consistent.

Addresses #570.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -508,13 +508,17 @@
         \item {\tt lr} instructions are loads
         \item successful {\tt sc} instructions are stores
         \item it is \unspecified\ whether failing {\tt sc} instructions are stores or not
-        \item Ideally each AMO instruction is a load for the read portion of
-        the operation and a store for the write part of the operation.
-        Implementations may choose to implement only the load part, only the
-        store part, or neither.
+        \item Each AMO instruction is a load for the read portion of the
+        operation. The address is always available to trigger on, although
+        the value loaded might not be, depending on the hardware
+        implementation.
+        \item Each AMO instruction is a store for the write portion of the
+        operation. The address is always available to trigger on, although
+        the value stored might not be, depending on the hardware
+        implementation.
         \end{steps}
 
-        If the destination register of any load or AMO is x0 then it is
+        If the destination register of any load or AMO is \Rzero then it is
         \unspecified\ whether a load trigger with \FcsrMcontrolSelect=1 will
         match.  Whether store triggers with \FcsrMcontrolSelect=1 match on
         AMOs is \unspecified.
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -223,15 +223,16 @@
         hold all valid virtual addresses but it need not be capable of holding
         other values.
 
-        If the A extension is supported, then trigger behavior is as follows
-        for the load and store bits:
-        \vspace{-0.2in}
-        \begin{tightlist}
+        \begin{steps}{If the A extension is supported, then trigger behavior is as follows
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -223,15 +223,16 @@
...
         \item {\tt lr} instructions are loads
         \item successful {\tt sc} instructions are stores
         \item it is \unspecified\ whether failing {\tt sc} instructions are stores or not
-        \item Each AMO instruction is both a load and a store
-        \end{tightlist}
+        \item Ideally each AMO instruction is a load for the read portion of
+        the operation and a store for the write part of the operation.
+        Implementations may choose to implement only the load part, only the
+        store part, or neither.
+        \end{steps}
```
ernie-sifive, line 235: Is "ideally" the right word here?  Maybe "nominally"?  I would also qualify the implementation's choices with something like "_In certain cases,_ implementations may choose..." since it is not allowed to remove the load or store in most cases.

pdonahue-ventana: Is it useful to allow implementations to not implement:
- load address triggers on AMOs
- store address triggers on AMOs
- load value triggers on AMOs where the value is actually written to the register file (so the hart knows the value)

Wouldn't any hart (no matter how it's implemented) already have all of the information for all of those?

Note that the later paragraph already says that it's an implementation option whether to implement load value triggers on AMOs where the value is not written (for fire-and-forget AMO operations) and whether to implement store value triggers at all (for far atomics).

pdonahue-ventana:
@JamesKenneyImperas might have opinions since he filed #570.

xml/hwbp_registers.xml
timsifive: Rereading what Paul wrote originally, I think my PR isn't intended to actually change anything.

The text below makes slightly more sense to me, but I don't think it changes anything. Thoughts?

* Each AMO instruction is a load for the read portion of the operation. The address is always available to trigger on, although the value loaded might not be, depending on the hardware implementation.
* Each AMO instruction is a store for the read portion of the operation. The address is always available to trigger on, although the value stored might not be, depending on the hardware implementation.

ernie-sifive: That seems good to me!

pdonahue-ventana: Except that "read" should be "write" on the second item, that looks OK to me.

ernie-sifive (approved):

scottj97 (changes requested):
It looks like mcontrol6 was updated by 5181cfc7958a355e5a382f5ae5db993fc889253a but not mcontrol?

scottj97 (approved):

xml/hwbp_registers.xml
```diff
@@ -492,28 +497,32 @@
...
+        the value loaded might not be, depending on the hardware
+        implementation.
+        \item Each AMO instruction is a store for the write portion of the
+        operation. The address is always available to trigger on, although
+        the value stored might not be, depending on the hardware
+        implementation.
+        \end{steps}
 
-        If the destination register of any load or AMO is x0 then it is
+        If the destination register of any load or AMO is \Rzero then it is
```
pdonahue-ventana, line 525: I still think that it's confusing to use the ABI name.  x0 is unambiguous while "zero" could be referring to x0 or it could mean that the destination register contains the value 0.

timsifive: On my mental TODO list is to change that macro to render x0 instead of zero. It will still get the font that we use for registers, and then the entire document will be consistent.

pdonahue-ventana: OK.  Works for me.

Issue #572, opened 2020-10-22
Clarify chaining of triggers of type 3, 4 and 5


JamesKenneyImperas:
The specification gives examples of trigger chains using type 2 (or 6) triggers. Types 3, 4 and 5 have no `chain` field. Can you please clarify the specification for triggers of these types in this regard? Specifically:

1. Is it legal to terminate a chain with a type 3, 4 or 5 trigger?
2. If it is legal, does the `timing` of all previous entries in the chain have to be `after` for the chain to match?
3. Can you enhance the documentation to include an example where this might be useful?

Thanks.

timsifive:
The intent was to only chain type 2 (or 6) triggers, and not to mix them with different types. I'll clarify that later.

timsifive:
@JamesKenneyImperas, can this be closed?

JamesKenneyImperas:
Yes, closing.

PR #596, opened 2020-11-12
Chains should all have the same type.

timsifive:
but not generally useful. Addresses #572.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -376,9 +376,8 @@
             trigger will be taken if and only if all the triggers in the chain
             match at the same time.
 
-            It is allowed to terminate a chain with a trigger with
-            $|type| \neq 2$. In most situations this results in a
-            chain that can never match.
+            Debuggers should not terminate a chain with a trigger with a
+            different type. It is undefined when exactly such a chain fires.
 
             Because \FcsrMcontrolChain affects the next trigger, hardware must zero it in
             writes to \RcsrMcontrol that set \FcsrTdataOneDmode to 0 if the next trigger has
@@ -651,9 +650,8 @@
             trigger will be taken if and only if all the triggers in the chain
             match at the same time.
 
-            It is allowed to terminate a chain with a trigger with
-            $|type| \neq 6$. In most situations this results in a
-            chain that can never match.
+            Debuggers should not terminate a chain with a trigger with a
+            different type. It is undefined when exactly such a chain fires.
 
             Because \FcsrMcontrolSixChain affects the next trigger, hardware must zero it in
             writes to \RcsrMcontrolSix that set \FcsrTdataOneDmode to 0 if the next trigger has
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -376,6 +376,10 @@
             trigger will be taken if and only if all the triggers in the chain
             match at the same time.
 
+            It is allowed to terminate a chain with a trigger with
+            $|type| \neq 2$. In most situations this results in a
+            chain that can never match.
+
```
ernie-sifive, line 382: I worry that this wording isn't precise enough.  Is what you are really saying is that if there is a trigger chain that includes a trigger with a type other than 2, it is allowed for an implementation to never match the chained trigger?

pdonahue-ventana:
It can presumably match if the chain has a mixture of type=2 and type=6.  Should each say that it can be chained with the other or should this be discouraged?  (I think it should probably be discouraged but perhaps this language was intended exactly to do that.)

xml/hwbp_registers.xml
timsifive: I phrased this to (hopefully) make hardware implementations simple so they don't have to look at the trigger type of the next trigger. Debuggers can set the chain bit if they want, and if it turns out to make sense, so much the better. But I don't want the hardware to have to do a bunch of extra checks in case a debugger is trying to be clever.

If you chain type 2/6 to:
* icount (3), unclear what exactly happens with the new pending bit, because we don't define a "match" for icount.
* itrigger (4), we don't define match either. I think typically the hart either takes an interrupt or executes an instruction, so probably never matches.
* etrigger (5), the same as for itrigger. You never really do a load/store *and* take an exception. A memory exception means the load/store didn't happen.
* tmexttrigger (7), don't define match, I assume this is something that's evaluated in between instruction executions, so again can't match together.

Maybe I should have written:
Debuggers should not terminate a chain with a trigger with a different type. It is implementation-specific when exactly such a chain fires.

timsifive:
@pdonahue-ventana, I see mcontrol6 as deprecating mcontrol. Maybe we should explicitly state that we suggest new implementations only implement mcontrol6.

xml/hwbp_registers.xml
ernie-sifive: Or... a trigger that chains more than one trigger type has undefined behavior.

timsifive: Yes, but we don't actually have a definition of what we mean by "undefined." It's fine for these triggers to fire at strange times, but it's not OK for such a setup to lock up the trigger unit.

pdonahue-ventana:
Actually, mcontrol6 already says "This replaces mcontrol in newer implementations and serves to provide additional functionality."  So maybe the existing proposal is fine since mentioning the mixture of 2 and 6 might encourage something that is already discouraged.

ernie-sifive (approved):
Looks good. At first I wondered why the text only mentioned terminating a chain but then remembered that types other than 2 and 6 don’t have chain bits.

Issue #583, opened 2020-11-02
Is it too late to modify tdata register formats and semantics to improve usability?


JamesKenneyImperas:
I want to start this issue with an apology: I have only recently looked at the Trigger Module part of this specification in any detail, so the points I'm making here should really have been raised a long time ago - I'm sorry about that. 

I am very concerned that the current specification of the trigger registers, especially `tdata1` and `tdata2`, makes their use very unwieldy. Clearly these have evolved significantly over time (I have not followed any of the history of that) but I think the resulting specification will cause unnecessary difficulties for all future implementations and programmers. If at all possible, I think this should be revised before submission for ratification.

I think that when designing a CSR interface, several principles should be observed:

1. Wherever possible, field positions and meanings should match for all XLEN. In this specific case, it would be better to locate the `type` and `dmode` fields in the lower 32 bits of the `tdata1` register, not at the top. This makes it much easier to write code that can be compiled for any XLEN and still work.  The one exception to this is that it sometimes makes sense to put a flag at bit XLEN-1 to facilitate branching.

2. When a register has different views (like `tdata1`), fields with the same meaning should appear at the same location so that it is easy to write common code for all views. This isn't true for many of the fields in `tdata1`: the worst offender is the `hit` bit, which can be found at positions 20, 22, 24 or XLEN-6, but many fields have at least two different positions, depending on view.

3. As a minor point, where fields have analogues in existing CSRs, it makes sense to locate them in the same place. For example, the `m`, `s` and `u` mode bits might be more logically placed at bits 3, 1 and 0 where they line up with similar mode bits in `mie`, but aligning the whole group to any nibble would help.

4. The interface should be easily programmable without pitfalls for the unwary. The semantic definition mentioned in issue #573 seems to interlock `tdata1` and `tdata2` in an undesirable manner, with each only being writable if the other is in a particular state. I think the rules listed there should be modified: perhaps any write to `tdata1` that changes `type` should automatically clear `tdata2`?

5. The implications of operations with different XLEN need to be considered and clearly stated. It isn't clear to me how exactly the triggers behave when `DXLEN=64` and `MXLEN=32`, for example, especially when both M mode and the debugger are independently programming triggers for their own purposes.

Once again, I apologise for raising these things so late in the day. Please feel free to close this issue if it is too late to do anything, or if all these things have previously been addressed in other issues.

pdonahue-ventana:
[...]

PR #597, opened 2020-11-13
Clarify trigger CSR behavior when XLEN changes.

timsifive:
Addresses #583.


xml/hwbp_registers.xml
```diff
@@ -3,9 +3,13 @@
...
 
-    In this section XLEN means MXLEN when in M-mode, and DXLEN when in Debug Mode.
-    Note that this makes several of the fields in \RcsrTdataOne move around based
-    on the current execution mode and value of MXLEN.
+    In this section XLEN means MXLEN when in M-mode, and DXLEN when in Debug
+    Mode. On systems where XLEN is writable, this is handled as follows.
+    Fields retain their values regardless of XLEN, which only affects where
+    in the register these fields appear (e.g. \FcsrTdataOneType). Some fields
+    are wider when XLEN is 64 than when it is 32 (e.g.
+    \FcsrTextraThirtytwoSvalue). The high bits in such fields retain their
```
ernie-sifive, line 11: So if I’ve set svalue to something with high bits set and then xLen changes, I can’t ever set another breakpoint using scontext since there is no way to clear the upper bits of svalue.  Is that really what we want?

timsifive: If software changes XLEN, it can deal with the complexity of making sure any triggers it set will work with the new XLEN. (Note that you will always be able to access the high bits from the debugger, since DXLEN never changes.) I think this is essential because any change between M-Mode to D-Mode can be an XLEN change, and going and zeroing out a bunch of bits every time that happens sounds like a side effect that could be worse, and there's nothing software can do about it.

ernie-sifive: Ok, two questions then.  Suppose DXLEN=64 and MXLEN can be changed between 32 and 64.  The underlying state will then store 34 bits for svalue.

1. While MXLEN=64, M-mode writes tdata2 such that upper bits of svalue are non-zero.  Then MXLEN is changed to 32.  Then M-mode writes tdata2.  Does the underlying state of svalue[35:16] retain its state or is it set to 0?

2. If svalue[33:16] retains its value, what part of svalue is used when comparing svalue to scontext?  The spec says "match if the low bits of scontext equal svalue".  Does this mean svalue[15:0] or svalue[33:0]?

xml/hwbp_registers.xml
```diff
@@ -3,9 +3,13 @@
     and Debug Mode to prevent untrusted user code from causing entry into Debug
     Mode without the OS's permission.
 
-    In this section XLEN means MXLEN when in M-mode, and DXLEN when in Debug Mode.
-    Note that this makes several of the fields in \RcsrTdataOne move around based
-    on the current execution mode and value of MXLEN.
+    In this section XLEN means MXLEN when in M-mode, and DXLEN when in Debug
+    Mode. On systems where XLEN is writable, this is handled as follows.
```
pdonahue-ventana, line 7: "XLEN" isn't writable.  MXL and SXL and UXL and VSXL are possibly writable and affect XLEN in various modes.  You can, for instance, have programmable MXL which allows MXLEN to change.  But you can also have hard-wired values which only allow for MXLEN=64 and SXLEN=32.  So XLEN changes every time you drop into S mode without anything being programmable.

Since these registers are only accessible in M mode, perhaps only MXL matters for the purposes of software reads.  But also consider that dropping into 32b S mode might affect things like how many ASID bits match on tdata3 even though S mode can't see tdata3 directly.  So I think that the concern here is whether XLEN can change, not whether anything is writable.

xml/hwbp_registers.xml
timsifive: 1. I think it makes sense to clear the upper bits if a field is written in 32-bit mode. Cheap to implement, and more likely to do what the users want.
2. In the referred issue, James argued that all bits should still matter. But zeroing out the high bits nicely solves this for the specific case where the trigger was written.

I'll update the PR so the inaccessible bits are zeroed out.

pdonahue-ventana: Why is DXLEN defined to be the widest supported XLEN instead of just DXLEN=MXLEN?  If MXLEN=SXLEN=UXLEN=32 then why does a debugger need to see the upper 32b that the software being debugged can't see?  That state doesn't affect anything.

timsifive: DXLEN is the widest supported XLEN because it seemed simpler that way. Part of it was me being ornery at the notion that XLEN can change at all. But if XLEN can change every time the hart runs, the debugger needs to discover it every time the hart halts. I'm not convinced it was the right decision, but it seems like a big thing to change right now.

pdonahue-ventana: I wasn't proposing to change it one business day before the freeze.  I just want to understand the history and if I was missing some important benefit.  It's a bit weird but not the weirdest thing I've seen. 😄

xml/hwbp_registers.xml
timsifive: Good point that XLEN isn't writable.

I think we're OK otherwise, but not 100% sure everything is always right when XLEN changes.

xml/hwbp_registers.xml
timsifive: More details at https://github.com/riscv/riscv-debug-spec/pull/298

ernie-sifive (approved):
Thanks, this looks good!

xml/hwbp_registers.xml
```diff
@@ -3,9 +3,15 @@
...
-    on the current execution mode and value of MXLEN.
+    In this section XLEN means MXLEN when in M-mode, and DXLEN when in Debug
+    Mode. On systems where those values of XLEN can differ, this is handled
+    as follows.
+    Fields retain their values regardless of XLEN, which only affects where
+    in the register these fields appear (e.g. \FcsrTdataOneType). Some fields
+    are wider when XLEN is 64 than when it is 32 (e.g.
+    \FcsrTextraThirtytwoSvalue). The high bits in such fields retain their
+    value but are not readable when XLEN is 32. A write to a field when XLEN
+    is 32 clears any inaccessible bits in that field.
```
pdonahue-ventana, line 14: This is per-field?  Clearly, if you `csrrw` when XLEN=32 then it clears the upper 32b (unless, perhaps, there are some fields that only exist if XLEN=64 in which case there was no write to that field).  What about these cases:
- In XLEN=32, you `csrrs mcontrol` and only change sizelo?  Does sizehi (which is inaccessible) get cleared?
- In XLEN=32, you `csrrs textra32` and only change svalue.  Do the upper 7 bits of mhvalue not get cleared?

(Just trying to poke holes in things to make sure they're robust)

timsifive: How about:
A modification of a register when XLEN is 32 clears any inaccessible bits in that register.

pdonahue-ventana: Perfect.  That seems straightforward to specify, to implement in hardware, and for software/debuggers to understand.  I'm not sure anything would be gained by keeping the other state around in some unmodified fields.

JamesKenneyImperas: I think this looks good. Apologies again for throwing a spanner in the works so late in the day.

Issue #412, opened 2018-12-14
itrigger mechanism change under CLIC proposal


chuanhua:
There is a new CLIC proposal that will change the local interrupt mechanism as follows:
[CLIC proposal](https://github.com/sifive/clic-spec/blob/master/clic.adoc)

So the corresponding itrigger mechanism in the debug spec needs to be changed as well. Here is a proposal to change the itrigger mechanism in the following site:
[itrigger mechanism change proposal under CLIC.](https://github.com/sifive/clic-spec/issues/29)

Please see if this corresponding change makes sense or not.

Thanks.

Chuanhua, Andes Technology

timsifive:
What's the status of the fast interrupt task group? If it looks like they have something that's likely to be ratified, it's probably time to see how we should accommodate it in the debug spec.

Kevin-Andes:
Hi,

I am Kevin Chen and the co-chair of the Fast Interrupt task group. Yes, our CLIC proposal has been stable for quite some time and pretty much ready to be ratified. Also, Chuanhua's proposal looks quite straightforward and useful.

Thanks.

            Regards,
            Kevin

timsifive:
Where can I see the latest working version of the fast interrupt document?

Kevin-Andes:
Hi Tim,

Sorry for my slow reply. Please feel free to take a look at our fast interrupt proposal in the following github:

    https://github.com/riscv/riscv-fast-interrupt

Thanks.

    Regards,
    Kevin

Kevin-Andes:
Hi Tim,

Just to clarify, are you suggesting to add thousands of registers (2 per interrupt) for infrequent cases that need triggering on more than 4 interrupts?  My first impression is that this sounds like a significant cost to pay for an uncommon use case...  Maybe we can try to think more and explore some cheaper alternatives.

Thanks.
Kevin

timsifive:
More or less, although I think of it as adding one bit of state per interrupt that can be configured. I assume that a typical CLIC implementation will have maybe 16 possible interrupts, so it would add 16 bits of state in the CLIC, and 1 bit in the Debug Module.. @chuanhua's proposal would require 10 bits of state in the DM for that case, on RV32, and then you're limited to triggering on a max of 2 interrupt sources.

Are my numbers off?

Kevin-Andes:
[...]

PR #599, opened 2020-11-13
Add clic bit to tmexttrigger.

timsifive:
Addresses #412.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -966,7 +966,7 @@
             and writing it has no effect.
         </field>
         <field name="0" bits="XLEN-7:23" access="R" reset="0" />
-        <field name="clic" bits="22" access="WARL" reset="0">
+        <field name="intctl" bits="22" access="WARL" reset="0">
             This optional bit, when set, causes this trigger to fire whenever an attached
             interrupt controller signals a trigger.
         </field>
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -965,7 +965,11 @@
             trigger(s) matched.  If the bit is not implemented, it is always 0
             and writing it has no effect.
         </field>
-        <field name="0" bits="XLEN-7:22" access="R" reset="0" />
+        <field name="0" bits="XLEN-7:23" access="R" reset="0" />
+        <field name="clic" bits="22" access="WARL" reset="0">
+            When set, this trigger will fire whenever the attached CLIC
+            signals a trigger.
```
pdonahue-ventana, line 971: It should be stated that the CLIC is optional.  I assume this is hard-wired if there's no CLIC.

Should this be a bit more generic and talk about an "interrupt controller" so that some CLIC competitor can use this (or even just since the existing CLIC will apparently change names)?  Something like:
```suggestion
            This optional bit, when set, causes this trigger to fire whenever an attached
            interrupt controller signals a trigger.
```

ernie-sifive: The generic wording looks good to me.  Also need to change the name of the field from "clic" to something else... "ic" or "intctrl"?

ernie-sifive (approved):

pdonahue-ventana (approved):
Looks OK to me.

Issue #584, opened 2020-11-03
Make list of functional changes between 0.13 and 1.0


timsifive:
Highlight the backwards incompatible changes.

JamesKenneyImperas:
I have a list of things that have changed for the trigger module in front of me. I'm certain that you know this better that I do, but I'll jot them down as a reminder:

1. scontext CSR address reassigned to 0x5a8;
2. hcontext CSR introduced at 0x6a8 as alias of mcontext when H present;
3. optional mscontext CSR in old scontext CSR location;
4. vu and vs added to icount, itrigger and etrigger trigger types when H present;
5. new trigger type mcontrol6;
6. new match types 8, 9, 12, 13;
7. textra32/textra64 mvalue/mselect renamed mhvalue/mhselect, and functionality expanded when H present;
8. textra32/textra64 sselect filtering by vsatp.ASID where appropriate;
9. tcontrol.sxce and tcontrol.hcxe added (backwards incompatible).

PR #600, opened 2020-11-19
Document changes since 0.13.

timsifive:
Excluding clarifications, which don't change the spec.


JamesKenneyImperas:
Hi Tim,

I find it difficult to read all the changes in this form, but I think some important things are missing. For example, the address of the `scontext` CSR has changed from 0x7AA to 0x5A8, which is completely incompatible. Is this mentioned?

I appended a list of possibilities to issue #584; did you see this?

James.

ernie-sifive (approved):
This list looks complete to me, given that clarifications are not included in the list.

@JamesKenneyImperas, the scontext address change is mentioned in the list.  It is not backward-incompatible since the original CSR address is still allowed.  (#535). The text in the list for this item should probably also mention the addition of scxe and hcxe.

timsifive:
> I find it difficult to read all the changes in this form

What could we change to make it easier to read?

JamesKenneyImperas:
Hi Tim,

This was down to my inability to use github properly, so please ignore.

Sorry,

James.

Issue #533, opened 2020-07-09
clarification of completion of program buffer is output for fence


silabs-PaulZ:
Hi @timsifive ,

I have some clarification questions regarding rule 9) in Chapter 4 RISC-V Debug, Section 4.1:
[“Completing Program Buffer execution is considered output for the purpose of fence instructions.”](https://github.com/riscv/riscv-debug-spec/blob/2210002c3724a2f91c63900eb79acfcceeaf4c40/core_debug.tex#L38)

Here is a copy of the riscv-spec section 2.7 Memory Ordering Instructions on the FENCE instruction for reference:
_The FENCE instruction is used to order device I/O and memory accesses as viewed by other RISCV
harts and external devices or coprocessors. Any combination of device input (I), device output
(O), memory reads (R), and memory writes (W) may be ordered with respect to any combination
of the same. Informally, no other RISC-V hart or external device can observe any operation in the
successor set following a FENCE before any operation in the predecessor set preceding the FENCE._

Are the following statements and my understanding correct:
From the debug requirement 9), an exit of the program buffer will set an internal ‘device output’ type flag. An ensuing FENCE instruction will then order the next successor access if it is dependent on a predecessor ‘device output’ access type. Flag is cleared if access has completed.

Exiting the program buffer can be accomplished with:
•	ebreak (jump back to debug halt address)
•	dret (exit debug and jump back to main execution)
•	exception (jump to debug exception address)

Also, can you clarify [‘this address'](https://github.com/riscv/riscv-debug-spec/blob/2210002c3724a2f91c63900eb79acfcceeaf4c40/implementations.tex#L57) in sectionA.2 Execution Based. Is 'this address' assumed to be the address of the DM or the debug exception address? I would assume that all access to DM be an I/O for fence instruction.

Regards,
-PaulZ

timsifive:
Can you see if #601 clears things up?

PR #601, opened 2020-11-30
Clarify details around fence and progbuf.

timsifive:
Addresses #533.

### DIFF
--- a/core_debug.tex
+++ b/core_debug.tex
@@ -37,10 +37,6 @@ How Debug Mode is implemented is not specified here.
     and {\tt dret}.  (To change the privilege level, the debugger can write
     \FcsrDcsrPrv and \FcsrDcsrV in \RcsrDcsr). The only exception is {\tt ebreak}, which ends
     execution of the Program Buffer when executed.
-\item \label{fence} Completing Program Buffer execution is considered output
-    for the purpose of {\tt fence} instructions.
-    This includes executing an {\tt ebreak} or {\tt dret} instruction, and
-    any exceptions that may be encountered.
 \item All control transfer instructions may act as illegal instructions if
     their destination is in the Program Buffer. If one such instruction acts as
     an illegal instruction, all such instructions must act as illegal
--- a/core_debug.tex
+++ b/core_debug.tex
@@ -37,8 +37,10 @@ How Debug Mode is implemented is not specified here.
     and {\tt dret}.  (To change the privilege level, the debugger can write
     \FcsrDcsrPrv and \FcsrDcsrV in \RcsrDcsr). The only exception is {\tt ebreak}, which ends
     execution of the Program Buffer when executed.
-\item \label{fence} Completing Program Buffer execution is considered output for the purpose
-    of {\tt fence} instructions.
+\item \label{fence} Completing Program Buffer execution is considered output
+    for the purpose of {\tt fence} instructions.
+    This includes executing an {\tt ebreak} or {\tt dret} instruction, and
+    any exceptions that may be encounterd.
 \item All control transfer instructions may act as illegal instructions if
     their destination is in the Program Buffer. If one such instruction acts as
     an illegal instruction, all such instructions must act as illegal
--- a/core_debug.tex
+++ b/core_debug.tex
@@ -40,7 +40,7 @@ How Debug Mode is implemented is not specified here.
 \item \label{fence} Completing Program Buffer execution is considered output
     for the purpose of {\tt fence} instructions.
     This includes executing an {\tt ebreak} or {\tt dret} instruction, and
-    any exceptions that may be encounterd.
+    any exceptions that may be encountered.
 \item All control transfer instructions may act as illegal instructions if
     their destination is in the Program Buffer. If one such instruction acts as
     an illegal instruction, all such instructions must act as illegal
--- a/implementations.tex
+++ b/implementations.tex
@@ -56,7 +56,9 @@ the Debug Module. The code there causes the hart to
 write to the Debug Module indicating an exception.
 Then the hart jumps back to the park loop.
 The DM infers from the write that there was an exception, and sets \FdmAbstractcsCmderr appropriately.
-Before entering the park loop, the hart will typically execute a {\tt fence} instruction, then write to the Debug Module indicating that the abstract command has finished.  The {\tt fence} ensures that any effects from the abstract command take effect before the DM returns \FdmAbstractcsBusy to 0.
+Typically the hart will execute a {\tt fence} instruction before entering the
+park loop, to ensure that any effects from the abstract command take effect
+before the DM returns \FdmAbstractcsBusy to 0.
 
 To resume execution, the debug module sets a flag which causes the hart to execute a {\tt dret}.
 When {\tt dret} is executed, \Rpc is restored from \RcsrDpc and normal execution resumes at the
--- a/implementations.tex
+++ b/implementations.tex
@@ -57,8 +57,9 @@ write to the Debug Module indicating an exception.
 Then the hart jumps back to the park loop.
 The DM infers from the write that there was an exception, and sets \FdmAbstractcsCmderr appropriately.
 Typically the hart will execute a {\tt fence} instruction before entering the
-park loop, to ensure that any effects from the abstract command take effect
-before the DM returns \FdmAbstractcsBusy to 0.
+park loop, to ensure that any effects from the abstract command, such as a
+write to \RdmDataZero, take effect before the DM returns \FdmAbstractcsBusy
+to 0.
 
 To resume execution, the debug module sets a flag which causes the hart to execute a {\tt dret}.
 When {\tt dret} is executed, \Rpc is restored from \RcsrDpc and normal execution resumes at the
--- a/implementations.tex
+++ b/implementations.tex
@@ -54,8 +54,6 @@ Program Buffer code) the hart returns to its park loop. If an exception is
 encountered, the hart jumps to a debug trap address within
 the Debug Module. The code there causes the hart to
 write to the Debug Module indicating an exception.
-This write is considered output for {\tt fence} instructions (see \#\ref{fence}
-on page \pageref{fence}).
 Then the hart jumps back to the park loop.
 The DM infers from the write that there was an exception, and sets \FdmAbstractcsCmderr appropriately.
 
--- a/implementations.tex
+++ b/implementations.tex
@@ -52,9 +52,9 @@ otherwise the DM causes a {\tt ebreak} to execute immediately.
 When {\tt ebreak} is executed (indicating the end of the
 Program Buffer code) the hart returns to its park loop. If an exception is
 encountered, the hart jumps to a debug trap address within
-the Debug Module. The code at that address causes the hart to
-write to an address in the Debug Module which indicates exception.
-This address is considered I/O for {\tt fence} instructions (see \#\ref{fence}
+the Debug Module. The code there causes the hart to
+write to the Debug Module indicating an exception.
+This write is considered I/O for {\tt fence} instructions (see \#\ref{fence}
 on page \pageref{fence}).
 Then the hart jumps back to the park loop.
 The DM infers from the write that there was an exception, and sets \FdmAbstractcsCmderr appropriately.
--- a/implementations.tex
+++ b/implementations.tex
@@ -56,6 +56,7 @@ the Debug Module. The code there causes the hart to
 write to the Debug Module indicating an exception.
 Then the hart jumps back to the park loop.
 The DM infers from the write that there was an exception, and sets \FdmAbstractcsCmderr appropriately.
+Before entering the park loop, the hart will typically execute a {\tt fence} instruction, then write to the Debug Module indicating that the abstract command has finished.  The {\tt fence} ensures that any effects from the abstract command take effect before the DM returns \FdmAbstractcsBusy to 0.
 
 To resume execution, the debug module sets a flag which causes the hart to execute a {\tt dret}.
 When {\tt dret} is executed, \Rpc is restored from \RcsrDpc and normal execution resumes at the
--- a/implementations.tex
+++ b/implementations.tex
@@ -54,7 +54,7 @@ Program Buffer code) the hart returns to its park loop. If an exception is
 encountered, the hart jumps to a debug trap address within
 the Debug Module. The code there causes the hart to
 write to the Debug Module indicating an exception.
-This write is considered I/O for {\tt fence} instructions (see \#\ref{fence}
+This write is considered output for {\tt fence} instructions (see \#\ref{fence}
 on page \pageref{fence}).
 Then the hart jumps back to the park loop.
 The DM infers from the write that there was an exception, and sets \FdmAbstractcsCmderr appropriately.


core_debug.tex
```diff
@@ -37,8 +37,10 @@ \section{Debug Mode} \label{debugmode}
     and {\tt dret}.  (To change the privilege level, the debugger can write
     \FcsrDcsrPrv and \FcsrDcsrV in \RcsrDcsr). The only exception is {\tt ebreak}, which ends
     execution of the Program Buffer when executed.
-\item \label{fence} Completing Program Buffer execution is considered output for the purpose
-    of {\tt fence} instructions.
+\item \label{fence} Completing Program Buffer execution is considered output
+    for the purpose of {\tt fence} instructions.
+    This includes executing an {\tt ebreak} or {\tt dret} instruction, and
+    any exceptions that may be encounterd.
```
ernie-sifive, line 43:
```suggestion
    any exceptions that may be encountered.
```

ernie-sifive (approved):
LGTM with typo fix

implementations.tex
```diff
@@ -52,9 +52,9 @@ \section{Execution Based} \label{execution_based}
 When {\tt ebreak} is executed (indicating the end of the
 Program Buffer code) the hart returns to its park loop. If an exception is
 encountered, the hart jumps to a debug trap address within
-the Debug Module. The code at that address causes the hart to
-write to an address in the Debug Module which indicates exception.
-This address is considered I/O for {\tt fence} instructions (see \#\ref{fence}
+the Debug Module. The code there causes the hart to
+write to the Debug Module indicating an exception.
+This write is considered I/O for {\tt fence} instructions (see \#\ref{fence}
```
pdonahue-ventana, line 57: Isn't it just considered O?

core_debug.tex
pdonahue-ventana, line 43: The act of clearing abstractcs.busy indicates PB completion, not the execution of these instructions.  I think that the rule should refer to the act of clearing that bit rather than this proposed sentence for a few reasons:
1. The sentence "includes" those things but is not necessarily exhaustive.  Although I can't think of any other cases, clearing abstractcs.busy includes these 3 cases and any others that we may be missing.
2. If an execution-based implementation executes a dret in the PB then it's UNSPECIFIED so arguably this sentence doesn't apply (or maybe it does if the dret actually terminates PB execution).  It just seems cleaner to talk about the observable act of abstractcs.busy becoming 0 than having to know whether rule 8 overrides rule 9.

silabs-PaulZ: Hi @timsifive ,
I am still a bit confused about the need to have this rule. Any 'device output' or 'memory write' should follow the rules for the FENCE instructions, whether or not the hart is in normal mode, debug mode, or running code from PB or any other location. If there is no exception to following the FENCE instruction as defined in the main RISC-V specification, then re-stating it here seems confusing to me. Perhaps you are suggesting that FENCE is not valid within PB execution but is valid only when finishing PB execution, which I doubt is the intention.

If though you are suggesting that completing the program buffer will automatically invoke a FENCE operation then that rule should be stated differently. But I don't think you are suggesting this behavior either.

Perhaps you are wanting to give a rule (or a note or a suggestion) that the debug (rom) code shall(..or shoud) include a FENCE instruction as the first instruction to be executed.

If you plan to keep the existing rule, could you consider changing the word " output " to " 'device output' " or '" memory write' " to be inline with the FENCE instruction terminology?

Please also refer to the comments posted on the RISC-V tech-debug listserve with the email subject title "clarification request: completion of program buffer is output for fence". One comment was that this rule should be more of a note.

implementations.tex
timsifive:
```suggestion
This write is considered output for {\tt fence} instructions (see \#\ref{fence}
```

timsifive:
For context, this language was originally added in #305. I think fence is much more defined now than it was back then.

The goal is for a debugger to be able to write memory using the program buffer, and have that write happen near instantly. E.g. in my debugger window I should be able to write to a memory-mapped LED, and that LED should visibly change state almost right away. It seems slow to force the debugger to execute fences through the program buffer every time it executes, so maybe an implicit fence is what we want? What is the best way to address this problem? (But OpenOCD does fence/fence.i often, so maybe I'm forgetting the real history behind this language.)

The spec is currently frozen, so we can only clarify things right now.

silabs-PaulZ:
Assuming the processor correctly adheres to the FENCE instruction requirements, is it fair to say that there is no additional or unique hardware or modifications of the debugger or processor to enable this rule?

pdonahue-ventana:
To satisfy rule 9, implementations must guarantee that the fence instruction enforces order between PB memory accesses and whatever channel they use to communicate PB completion to the DM.  For implementations as described in appendix A, that happens naturally (as mentioned in A.2).  For implementations where the hart signals PB completion to the DM using a sideband communication channel, rule 9 seems to me like it requires that these implementations add completion semantics to the RISC-V fence instruction which normally provides no such guarantee.

This was also discussed back in July: https://lists.riscv.org/g/tech-debug/topic/75725318#146

Silabs-ArjanB:
Hello @pdonahue-ventana ,

>>To satisfy rule 9, implementations must guarantee that the fence instruction enforces order between PB 
>> memory accesses and whatever channel they use to communicate PB completion to the DM. 

In the discussion on https://lists.riscv.org/g/tech-debug/topic/75725318#146 there is still the major unanswered question whether rule 9 relates to (only) "order all the memory accesses in the program buffer with following memory accesses" or that rule 9 relates to "full execution of the program buffer has completed". If the first interpretation applies, then rule 9 should just be removed (it is just regular fence behavior). If the second interpretation applies, then the question becomes: Why is this needed and how can this be detected? 

>> For implementations as described in appendix A, that happens naturally (as mentioned in A.2).

Could you please explain why an implementation according to appendix A '_naturally_' signals 'completing program buffer execution' so that a fence instruction can enforce ordering for that (implied) '_naturally signaled completion signal_'? (I understand that this works for the mentioned exception case in which an actual write happens to the DM, but how is this supposed to work in the normal scenario in which an ebreak is executed indicating the end of the program buffer code?)

pdonahue-ventana:
> In the discussion on https://lists.riscv.org/g/tech-debug/topic/75725318#146 there is still the major unanswered question whether rule 9 relates to (only) "order all the memory accesses in the program buffer with following memory accesses" or that rule 9 relates to "full execution of the program buffer has completed". If the first interpretation applies, then rule 9 should just be removed (it is just regular fence behavior). If the second interpretation applies, then the question becomes: Why is this needed and how can this be detected?

Rule 9 considers program buffer completion as a memory operation for purposes of fences and the memory consistency model.  It's not just normal fence behavior because completion of the program buffer isn't a memory operation so it normally wouldn't participate in the memory consistency model.  (Note that in an implementation such as appendix A that completion of the program buffer does involve a memory operation which would naturally be fenced but there may be implementations that implement the program buffer but do not communicate completion via a store.)

> Could you please explain why an implementation according to appendix A 'naturally' signals 'completing program buffer execution' so that a fence instruction can enforce ordering for that (implied) 'naturally signaled completion signal'? (I understand that this works for the mentioned exception case in which an actual write happens to the DM, but how is this supposed to work in the normal scenario in which an ebreak is executed indicating the end of the program buffer code?)

Appendix A describes this (perhaps incompletely):
https://github.com/riscv/riscv-isa-sim/blob/master/debug_rom/debug_rom.S

After an exception, there's a [store](https://github.com/riscv/riscv-isa-sim/blob/master/debug_rom/debug_rom.S#L50) that indicates that there was an exception (as described in A.2).  Then it does an [ebreak](https://github.com/riscv/riscv-isa-sim/blob/master/debug_rom/debug_rom.S#L51).  After either that ebreak or the ebreak that normally happens at the end of the PB, it enters the park loop which performs a [store](https://github.com/riscv/riscv-isa-sim/blob/master/debug_rom/debug_rom.S#L34) which tells the DM it's in the park loop and that the PB is completed (if it had previously been executing the PB).  This park loop store is what the fence naturally operates on.

We at my company have spent many engineer-hours trying to understand rule 9 over the last several months, likely more than any other single topic in the debug architecture.  It definitely does impose a requirement that is not trivially met on all implementations.  It's possible that the requirement it imposes is useful for some purpose.  The requirement it imposes does not seem to be useful for guaranteeing that an LED lights up (except on implementations where fences have special properties above and beyond the minimum required by the architecture).

timsifive:
> We at my company have spent many engineer-hours trying to understand rule 9 over the last several months, likely more than any other single topic in the debug architecture. It definitely does impose a requirement that is not trivially met on all implementations. It's possible that the requirement it imposes is useful for some purpose. The requirement it imposes does not seem to be useful for guaranteeing that an LED lights up (except on implementations where fences have special properties above and beyond the minimum required by the architecture).

I had never imagined this would be so complex. Given that, you probably have the best understanding of any of us of actually what the implications are.

You say that LED doesn't light up even with this rule. Is that because the write might be cached? Is the requirement useful for anything?

Rereading #305, the thinking may be that this requirement is necessary for a subsequent cache flush instruction to be guaranteed to flush this write. Of course the debugger can always put a fence in the program buffer, although if it will do that every time it writes memory then it does make sense to make that behavior implicit.

silabs-PaulZ:
Hi @pdonahue-ventana , thank you for that clarification. Obviously, many have struggled with this rule (or ignored it) and I would hope we re-write this to make it easier and more straightforward.

In the execution based example (A.2), you are interpreting the phrase 'Completing Program Buffer' in a non-precise manner, where the ‘device output’ store event occurs after the ebreak instruction. I’m not opposed to the non-precise behavior, but the spec is now suggesting otherwise. In particular, this pull request seems to add a more precise definition of when the program buffer execution terminates as “…executing ‘ebreak’ or ‘dret’ and any exceptions”.  

Moreover, I’m not sure how executing a ‘dret’ from the program buffer would properly ensure the ‘device output’ event since the core will immediately jump back out of debug mode and not execute the ‘device output’ store instruction in the park loop. Granted, you would not typically want to execute ‘dret’ from the program buffer and instead call ‘dret’ from evaluating a resume command from the park loop, but it is allowed.

The overall purpose of this rule is still not clear to me. If you have instructions in a program buffer that you execute that does not itself perform a ‘device output’ store, then why would we need to generate a spurious ‘device output’ event when completing the program buffer.  If the instructions in a program buffer do execute a ‘device output’ store, then again why would we need to generate an additional ‘device output’ event when completing the program buffer. The ensuing fence instruction in the park loop will guarantee that the real intended ‘device output’ store is ordered.

For the case of executing dret from the program buffer, the core returns to normal non-debug execution. There could be an outstanding ‘device output’ store from the program buffer execution that may not be ordered properly as the normal code will most likely not have a fence instruction. (again, it is probably not a good idea to have dret in program buffer)

I would recommend a different rule or guideline:
A fence instruction should be executed or invoked prior to beginning execution of a new debug command and before exiting debug mode to ensure proper data ordering.

pdonahue-ventana:
> Rereading #305, the thinking may be that this requirement is necessary for a subsequent cache flush instruction to be guaranteed to flush this write.

If you have an implementation with bespoke cache flush instructions (which are not yet in the architecture) then all you need is to fence between previous stores and the flush.  I don't see how the order of program buffer completion with respect to the store and the flush comes into play.

> Of course the debugger can always put a fence in the program buffer, although if it will do that every time it writes memory then it does make sense to make that behavior implicit.

Rule 9 doesn't introduce any implicit fence and therefore doesn't guarantee that all loads/stores in the PB are visible before the PB completion is visible.  If the debugger wants all loads/stores to be visible before PB completion then it must add a `fence iorw,o` to the end of the PB (or have a second PB iteration with a fence which will guarantee the order of previous loads/stores with respect to the completion of this second PB iteration but not with respect to the first).

Just like normal software needs to add fences if it wants to force a particular order, the debugger must add fences if it wants to force a particular order.

pdonahue-ventana:
I really want to caution everyone to not assume any particular implementation such as appendix A.  To paraphrase John Marshall (legendary 19th century American jurist), we must never forget that it is an architecture we are expounding.

Some implementations will not execute any instructions in debug mode other than those instructions in the PB (and the implicit ebreak, if any).  All other debug mode functionality may be based on wires and hardware state machines.  In particular:
- The hart may indicate to the DM that it is halted via a wire, not a store.
- The DM may tell the hart to start executing the PB via another wire, not by having a park loop read the DM.
- The hart may indicate to the DM that the PB is complete and whether there was an exception via other wires, not stores to the DM.
- The DM may tell the hart to exit debug mode via yet another wire, not by having a park loop read the DM.
- After learning that it has been instructed to exit debug mode, the hart may do so by setting pc=dpc and resuming instruction fetch, not by having a `dret` instruction go through the pipe.

In short, even if implementing the program buffer, the architecture:
- does not require a park loop
- does not require any particular behavior for the `dret` instruction in debug mode (explicitly stated in 4.7)

> In the execution based example (A.2), you are interpreting the phrase 'Completing Program Buffer' in a non-precise manner, where the ‘device output’ store event occurs after the ebreak instruction. I’m not opposed to the non-precise behavior, but the spec is now suggesting otherwise. In particular, this pull request seems to add a more precise definition of when the program buffer execution terminates as “…executing ‘ebreak’ or ‘dret’ and any exceptions”.

I have always interpreted "Completing Program Buffer" to mean that abstractcs.busy is observed to be 0.  It seems like the only thing that is visible to another hart and/or observer.  That's why this PR concerns me.

> Moreover, I’m not sure how executing a ‘dret’ from the program buffer would properly ensure the ‘device output’ event since the core will immediately jump back out of debug mode and not execute the ‘device output’ store instruction in the park loop. Granted, you would not typically want to execute ‘dret’ from the program buffer and instead call ‘dret’ from evaluating a resume command from the park loop, but it is allowed.

Executing a `dret` from the PB may jump out of debug mode or it may cause an exception or, since it's UNSPECIFIED (both in section 4.7 and in section 4.1 rule 8), it can do absolutely anything (nop, store, multiply, branch-to-self).

> The overall purpose of this rule is still not clear to me.

I am also struggling.

> If you have instructions in a program buffer that you execute that does not itself perform a ‘device output’ store, then why would we need to generate a spurious ‘device output’ event when completing the program buffer. If the instructions in a program buffer do execute a ‘device output’ store, then again why would we need to generate an additional ‘device output’ event when completing the program buffer.

Fences can also order normal loads and stores with respect to device output stores.  So `fence w,o` in the PB should make sure that other harts see any prior stores before they see abstractcs.busy=0 (as long as they fence their loads properly).

> The ensuing fence instruction in the park loop will guarantee that the real intended ‘device output’ store is ordered.

This assumes an implementation with a park loop and that the park loop contains a fence.

> For the case of executing dret from the program buffer, the core returns to normal non-debug execution. There could be an outstanding ‘device output’ store from the program buffer execution that may not be ordered properly as the normal code will most likely not have a fence instruction. (again, it is probably not a good idea to have dret in program buffer)

This assumes that executing `dret` from the PB returns to non-debug execution.  Assuming an implementation where that is true then, as I commented above, this PR is particularly dangerous because it says that you're able to force ordering between memory operations and a dret instruction (rather than abstractcs.busy) and dret is normally not part of the memory consistency model.

timsifive:
> Just like normal software needs to add fences if it wants to force a particular order, the debugger must add fences if it wants to force a particular order.

That is one solution, but I think debuggers always want to force the order. (Imagine a series of writes to a memory-mapped device.) In that case it makes sense to have the hardware "just do it," which I assume is cheap to implement and saves the debugger from having to issue fence instructions all the time.

> I really want to caution everyone to not assume any particular implementation such as appendix A.

Yes. Thanks.

If we did not have this rule, and a debugger would do 2 writes in separate program buffer executions, is there any guarantee about the ordering of those writes?

pdonahue-ventana:
> That is one solution, but I think debuggers always want to force the order. (Imagine a series of writes to a memory-mapped device.)

That device case is already handled by the memory consistency model:
> memory access *a* precedes memory access *b* in global memory order if *a* precedes *b* in program order and one or more of the following holds:
> ...
> 3. *a* and *b* are accesses to the same strongly-ordered I/O region
> ...
> 5. *a* and *b* are accesses to I/O regions associated with the same channel (except for channel 0)

(I'm assuming all addresses for a memory-mapped device are either in the same I/O region or on the same channel.)

> If we did not have this rule, and a debugger would do 2 writes in separate program buffer executions, is there any guarantee about the ordering of those writes?

My observations:
1. Without fences, there is no ordering guarantee on your scenario whether this rule exists or not (other than special cases like the I/O region example above).
2. With `fence ow,ow`, there is an ordering guarantee for those stores whether this rule exists or not (since you only asked about the ordering of stores with respect to each other, not ordering of them with respect to PB completion).
3. If this rule transforms into a requirement for an implicit `fence iorw,iorw` then 2 applies.

But if there's an implicit fence then:
1. That doesn't seem like a clarification to rule 9.  It seems like an entirely new requirement.
2. Does this also apply to abstract memory accesses?
3. Does this also apply to memory accesses performed during single-stepping?

For comparison, ARMv8 says that how memory is accessed doesn't change in debug state (their equivalent of debug mode), and they explicitly include the memory order model in that statement.  They only support a single-instruction program buffer which means that fences are even more painful for their debuggers (and I don't see any DMB instructions in their version of OpenOCD and the only DSB is when they modify their mode register).  In their memory-access mode, they have essentially a program buffer with two hard-wired instructions and neither is a fence.

Is there really a need to add hardware requirements so that debuggers have stricter ordering than normal software?  Especially in light of the progress axiom.

timsifive:
> Is there really a need to add hardware requirements so that debuggers have stricter ordering than normal software? Especially in light of the progress axiom.

It would be nice not to have to. Without an extra rule, is there anything that a debugger can do (using the program buffer) to flush a write to a memory-mapped device?

pdonahue-ventana:
The earlier comment was:
> in my debugger window I should be able to write to a memory-mapped LED, and that LED should visibly change state almost right away.

Is this desire different than a user running a LED control program directly on the hart and then expecting to see it turn on shortly after giving the command to do so?

The program buffer can do anything that normal software can do other than:
- wfi
- branches/jumps
- ecall, mret, sret, uret, dret
- auipc

Therefore, the debugger should be able to do whatever the LED control program does to force the LED to light up.  Of course, the progress axiom means that no special action of any kind is required by either a program or a debugger to force the LED to *eventually* light up.

I expect that any reasonable implementation would consider "eventually" to be on the order of milliseconds or less, not on the order of seconds or more.  So implementations will generally force the write out to the LED controller and I also assume that the LED controller would have fairly low latency between seeing the write and lighting the LED.  If an implementation hasn't pushed the write out to the LED controller within 100ms or so (which is pushing the limits of what humans perceive as "instantaneous"), I wonder if it would ever push it out and whether such an implementation complies with the progress axiom.

silabs-PaulZ:
I agree with @pdonahue-ventana  . I am still in favor of removing this requirement. I am also against adding hardware to force a fence or any other mechanism to ensure the timeliness of a write action.

If fence is required, then that fence can be part of the program buffer code. 

If LED write needs to be guaranteed in a timely fashion, perhaps after writing the register the user could execute a fence followed with a read of the same register within the program buffer. This can be done with multiple commands if necessary. Bottom line is that this can all be done with SW (in the program buffer) just as it is done outside of debug mode without adding special hardware.

timsifive:
2 thoughts:

1. The LED example is not the worst case because, as Paul said, I/O regions shouldn't be cached. A worse code is in a multihart system, with one hart halted and another running. The debugger changes memory through the halted hart, and expects the running one to instantly see this change. I'm not sure if this changes anything for our discussion but I wanted to mention it.
2. I don't even know what to write for the Appendix about this, so I took it out. @ernie-sifive, can you enlighten me with what should be said there?

implementations.tex
```diff
@@ -52,10 +52,8 @@ \section{Execution Based} \label{execution_based}
...
 Program Buffer code) the hart returns to its park loop. If an exception is
 encountered, the hart jumps to a debug trap address within
-the Debug Module. The code at that address causes the hart to
-write to an address in the Debug Module which indicates exception.
-This address is considered I/O for {\tt fence} instructions (see \#\ref{fence}
-on page \pageref{fence}).
+the Debug Module. The code there causes the hart to
+write to the Debug Module indicating an exception.
 Then the hart jumps back to the park loop.
 The DM infers from the write that there was an exception, and sets \FdmAbstractcsCmderr appropriately.
```
ernie-sifive, line 58:
```suggestion
the Debug Module. The code there causes the hart to
write to the Debug Module indicating an exception.
Then the hart jumps back to the park loop.
The DM infers from the write that there was an exception, and sets \FdmAbstractcsCmderr appropriately.
Before entering the park loop, the hart will typically execute a {\tt fence} instruction, then write to the Debug Module indicating that the abstract command has finished.  The {\tt fence} ensures that any effects from the abstract command take effect before the DM returns \FdmAbstractcsBusy to 0.
```

ernie-sifive (commented):
Here's a suggestion for some wording in the appendix regarding use of fence.

implementations.tex
timsifive: That's a good starting point. Thanks. I was going to extend it by saying that the fence should happen both in the park loop and the exception handler, but rocket-chip does not have a fence at the start of the exception handler: https://github.com/chipsalliance/rocket-chip/blob/master/scripts/debug_rom/debug_rom.S#L64

The comment on the `fence` in _entry also doesn't match what we're saying here.

I think the language we're writing here is correct, and rocket-chip either isn't, or has some hack that's not visible in debug_rom.S to make things work out.

ernie-sifive: A rocket-chip exception writes to EXCEPTION, then does an ebreak which goes back to _entry, which then executes the fence.  The rocket-chip comment is just saying that if the abstract command writes something to DATA (such as a register read value), the fence ensures that the value in DATA is visible before the write to HALTED clears the abstractcs.busy bit.  Without the fence, conceivably the write to DATA could be delayed relative to HALTED.

timsifive: Got it. Thanks for explaining.

timsifive:
@pdonahue-ventana, can you take a look at this final form? I just want to make sure I got it right.

timsifive:
Pinging @pdonahue-ventana

pdonahue-ventana (approved):
This rationale is a bit different than what I thought it was.  I had to ponder this a bit.

Section 3.7 says: "If a command returns results, the Debug Module must ensure they are placed in the data registers before busy is cleared."  That certainly seems like a necessary requirement.

I now understand that rule 9 was trying to guarantee the 3.7 behavior for ROM-based implementations while imposing an unnecessary requirement on non-ROM implementations.  I think that this PR removes the unnecessary fence requirement while still preserving the necessary requirement in section 3.7.  In the appendix it provides guidance to ROM-based implementations on how to meet the 3.7 requirement using a fence.  Other implementations can do whatever they want as long as they comply with 3.7 (and ROM-based implementations don't have to use a fence if they have some other magic mechanism).

So I think this is good.

PR #602, opened 2020-12-02
Add version 1.0 value to dtmcs.version field

ernie-sifive:
I noticed that dtmcs.version did not have a value specified to indicate spec version 1.0.

### DIFF
--- a/xml/jtag_registers.xml
+++ b/xml/jtag_registers.xml
@@ -73,6 +73,8 @@
             0: Version described in spec version 0.11.
 
             1: Version described in spec version 0.13.
+            
+            2: Version described in spec version 1.0.
 
             15: Version not described in any available version of this spec.
         </field>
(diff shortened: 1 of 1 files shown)


timsifive:
Did the DTM behavior change since 0.13? I think we can use the same encoding as before and say it applies to DTM 0.13 and DTM 1.0.

pdonahue-ventana:
Things definitely changed from 0.13:
```
git diff b4f1f439a57afe04aab61b8b2c42e490f3aaaf58 -- jtagdtm.tex
git diff b4f1f439a57afe04aab61b8b2c42e490f3aaaf58 -- xml/jtag_registers.xml
```

I didn't scour it for all substantial changes but a quick scan shows that at least the dmihardreset description changed to require DTM registers to be reset.  That wasn't required in the dmihardreset description in 0.13 but maybe it was described elsewhere.

timsifive:
> Things definitely changed from 0.13:

That language changed happened in #508, which is titled as a clarification. Reading it, it sounds like a clarification to me and not a change in how things work.

timsifive (approved):

pdonahue-ventana:
That's fine and I'm OK with the proposal if you think that the diffs are all just clarifications.

Issue #569, opened 2020-10-15
Clarify data value for triggers on sign-extending loads


JamesKenneyImperas:
For a data value trigger on a sign extending load, what value is used for value comparisons? For example, if an `lb` instruction loads the value `0x80` from memory, is the data value considered to be `0x80` or `0xffffff80` for the purposes of the trigger?

Thanks.

ernie-sifive:
Hi James,

I think the number of bits of *tdata2* that are compared to the load value
is determined by the *size* field.  If you have a trigger conditioned by
*size=1*, then only 8 bits are compared.  As stated in the text, *size=0*
with *select=1* (data value trigger) has undefined behavior.

Ernie

JamesKenneyImperas:
Hello Ernie,

Are you saying that, in the example I gave, only the low 8 bits of the loaded value would be compared with the lower 8 bits of tdata2?

If so, can you please also clarify what the _lower half_ and _upper half_ references mean in match types 4 and 5? Are these 4-bit nibbles in this case, or something else?

Thanks.

ernie-sifive:
Yes, only 8 bits should be compared if size=1.

The spec is a little vague on the masking definition.  I would take it to mean that the 8-bit data value is extended to XLEN (doesn't matter if zero-extended or sign-extended), then the mask is applied, then part of the result is compared to part of the value according to `size`.  The mask is the upper half of tdata2 (16 bits with XLEN=32 or 32 bits with XLEN=64) and the value is the lower half of tdata2.

So in your example with size=1 and XLEN=32:

- match=4 means `trigger = (value[7:0] & tdata2[23:16]) == tdata2[7:0]`.
- match=5 means `trigger = 1`.  (Upper 16 bits are don't-care when size=1)

JamesKenneyImperas:
Thanks for the clarification. I think it would be helpful to state this behavior explicitly in the specification - I'll leave this issue open as a reminder.

tianshi-xty:
Hi Ernie，

“match=5 means trigger = 1. (Upper 16 bits are don't-care when size=1)”
Can you please explain why trigger = 1 when upper 16 bits are don't-care when size=1？If we don't care about  upper 16 bits of datavalue, is it more reasonable to make trigger = 0?

JamesKenneyImperas:
I think Ernie's second example probably isn't right; assuming `datavalue[7:0]` is zero-extended to 32 bits, the logic is:

[...]

PR #604, opened 2020-12-03
Clarify how many bits mcontrol/mcontrol6 compare.

timsifive:
Addresses #569.


scottj97 (changes requested):
Shouldn't this only apply when select=1 (match data, not address)?

ernie-sifive (changes requested):
I agree with @scottj97 -- the bit masking only applies when data value is being compared.

pdonahue-ventana:
What about things like match=1, tdata2=0x00007FFF, and we match on a byte load?  We're simultaneously comparing only [7:0] and only [XLEN-1:16].  I suspect that it's the intersection of those two (so nothing is compared in this example) but that should be clear.

Also, if match=4 or 5 what value is "the value" described there?

timsifive:
> Shouldn't this only apply when select=1 (match data, not address)?

Yes, good point.

timsifive:
@pdonahue-ventana, I think I addressed your question in the latest change. Can you take a look?

xml/hwbp_registers.xml
```diff
@@ -276,14 +276,21 @@
...
+            This bit determines the contents of the XLEN-bit value to be
+            compared.
+
             0: Perform a match on the lowest virtual address of the access.  In
             addition, it is recommended that the trigger also matches if any of
             the other accessed virtual addresses match.
             (E.g. on a 32-bit read from 0x4000, the lowest address is 0x4000
             and the other addresses are 0x4001, 0x4002, and 0x4003.)
+            If the virtual address is smaller than DXLEN, the high bits
+            contain 0.
```
pdonahue-ventana, line 288: This sentence isn't clear to me.  I think that you mean that if XLEN<DXLEN then address[DXLEN-1:XLEN]=0 for purposes of the comparison.  But I don't think this works for the case where DXLEN=64 and MXLEN=32 and M mode sets up a trigger with action!=1.  The M-mode write to tdata2 is defined to sign-extend [63:32].  If address[63:32]=0 always but tdata[63:32] is sign-extended then you can't match on any address with [31]=1.

The other two options are:
1. Sign-extend the address before comparison.  But if you have MXLEN=64 and SXLEN=32 and satp.MODE=Bare then the top byte of what S mode can access is VA=PA=0x00000000FFFFFFFF, not VA=PA=0xFFFFFFFFFFFFFFFF (which is likely a vacant PMA region).

2. Say that no bits above the current XLEN are compared.  In the MXLEN=64 and SXLEN=32 case, this means that you will ignore tdata2[63:32] which can also be weird because address 0x12345678 matches against tdata2=0xfeedface12345678.

I bet that 2 would be more popular with implementations so they don't have to have 33 bits fanout from addr[31] (for no real benefit).

xml/hwbp_registers.xml
```diff
@@ -276,14 +276,21 @@
...
 
             1: Perform a match on the data value loaded or stored, or the
             instruction executed.
+            In the value to compare, any bits beyond the size of the data
+            access will contain 0.
```
pdonahue-ventana, line 293: Similar thoughts here.  If DXLEN=64 and MXLEN=32 and you write tdata2 from M mode with tdata2[31]=1 then you can never match on a 32b load since the load data will have [63:32]=0 but tdata2[63:32]=0xFFFFFFFF.

xml/hwbp_registers.xml
```diff
@@ -434,6 +441,10 @@
             13: Matches when \FcsrMcontrolMatch$=5$ would not match.
 
             Other values are reserved for future use.
+
+            When \FcsrMcontrolSelect=1 and access size is N, a match only
+            compares the low N bits of the value to the low N bits of
+            \RcsrTdataTwo.
```
pdonahue-ventana, line 447: I initially thought that this was fine but, to keep up what is undoubtedly my reputation as Mr. Corner-case: If match=4 then what is "the lower half of the value" for a load of 8b or 16b?  How about defining match=4 as "matches when the lower XLEN/2 bits of the value ANDed with the upper XLEN/2 bits of tdata2 match the lower XLEN/2 bits of tdata2"?  (Similar for match=5.)  Then it might be better for this sentence to say "... a match only compares the low N bits" (removing "of the value to the low N bits of tdata2" which implies that it's the low N bits of both things being compared, whatever those things may be).

I think that this sentence means that the newly-added sentence in select=1 is unnecessary because if you're not comparing those bits then it doesn't matter if you zero-extend or not.  Perhaps the new select=0 sentence should be removed and a sentence added down here like: "When select=0, a match only compares the low XLEN bits."

xml/hwbp_registers.xml
timsifive: Option 2 makes sense to me. The quirk you point out is perhaps a little odd, but at the same time it doesn't make sense to include bits in the comparison which cannot be part of the virtual address.

xml/hwbp_registers.xml
timsifive: I think that's also solved by limiting matches to looking at the low XLEN bits of everything.

xml/hwbp_registers.xml
timsifive: I appreciate you're bringing up the corner cases. I wish they didn't exist, but that's largely a problem of my own making. (Except for XLEN being writable. I had nothing to do with that.)

This language about halfs is not precise enough, evidently. I'll make it more precise and more ugly.

pdonahue-ventana (commented):
This looks good.  I tried to poke more holes in it but I can't think of any.

xml/hwbp_registers.xml
```diff
@@ -276,14 +276,18 @@
...
         <field name="select" bits="19" access="WARL" reset="0">
-            0: Perform a match on the lowest virtual address of the access.  In
-            addition, it is recommended that the trigger also matches if any of
-            the other accessed virtual addresses match.
+            This bit determines the contents of the XLEN-bit compare values.
+
+            0: There is at least one compare value and it contains the lowest
+            virtual address of the access.
+            In addition, it is recommended that there are additional compare
+            values for the other accessed virtual addresses match.
```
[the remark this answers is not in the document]
timsifive: Good point. Done.

timsifive:
@ernie-sifive, github still says you tagged this as "needs changes." Can you take a look and see if I missed anything?

ernie-sifive (approved):

PR #608, opened 2020-12-24
Update version of priv spec to one with hypervisor

ernie-sifive:
I was informed that the reference to the Privileged Spec is outdated and received a suggestion to change it to point to a version that includes hypervisor.


timsifive (approved):
Thanks!

PR #610, opened 2020-12-28
Define some more terms/acronyms.

timsifive:
Many acronymos contributed by Bruce Ableidinger.

### DIFF
--- a/introduction.tex
+++ b/introduction.tex
@@ -20,22 +20,71 @@ functionality.
 
 \section{Terminology}
 
-A \emph{platform} is a single integrated circuit consisting of one or more
-\emph{components}. Some components may be RISC-V cores, while others may have a different
-function. Typically they will all be connected to a single system bus.
-A single RISC-V core contains one or more hardware threads, called
-\emph{harts}.
+\begin{description}[style=nextline]
+    \item[AMO]
+        Atomic Memory Operation.
+    \item[BIST]
+        Built-In Self Test.
+    \item[BYPASS]
+        JTAG instruction that selects a single bit data register, also called BYPASS.
+    \item[component]
+        A RISC-V core, or other part of a platform.
+        Typically all components will all be connected to a single system
+        bus.
+    \item[CSR]
+        Control Status Register.
+    \item[DM]
+        Debug Module.
+    \item[DMI]
+        Debug Module Interface.
+    \item[DR]
+        Data Register.
+    \item[DTM]
+        Debug Transport Module.
+    \item[DXLEN]
+        Debug XLEN, which is the widest XLEN a hart supports, ignoring the
+        current value of \Fmxl in \Rmisa.
+    \item[GPR]
+        General Purpose Register.
+    \item[hart]
+        A hardware thread in a RISC-V core.
+    \item[IDCODE]
+        32-bit Identification CODE, and a JTAG instruction that returns the IDCODE value.
+    \item[IR]
+        Instruction Register.
+    \item[JTAG]
+        Joint Test Action Group.
+    \item[MIE]
+        Machine Interrupt Enable register; also a bit in mstatus.
+    \item[NAPOT]
+        Naturally Aligned Powers-Of-Two.
+    \item[NMI]
+        Non-Maskable Interrupt.
+    \item[OpenOCD]
+        Open On-Chip Debugger.
+    \item[PC]
+        Program Counter.
+    \item[platform]
+        A single integrated circuit consisting of one or more \emph{components}.
+    \item[physical address]
+        An address that is directly usable on the system bus.
+    \item[RV32]
+        RISC-V core with XLEN of 32.
+    \item[RV64]
+        RISC-V core with XLEN of 64.
+    \item[SBA]
+        System Bus Access.
+    \item[TAP]
+        Test Access Port.
+    \item[XLEN]
+        Width of the core's integer register file.
+    \item[virtual address]
+        An address as a hart sees it. If there is address translation this
+        may be different from the physical address. If there is no
+        translation then it will be the same.
+\end{description}
 
-\emph{DXLEN} of a hart is its widest supported XLEN, ignoring the current value
-of \Fmxl in \Rmisa.
-
-A \emph{physical address} is directly usable on the system bus.
-
-A \emph{virtual address} is an address as a hart sees it. If there is address
-translation this may be different from the \emph{physical address}. If there is
-no translation then it will be the same.
-
-\subsection{Context}
+\section{Context}
 
 \begin{steps}{This document is written to work with:}
 \item The RISC-V Instruction Set Manual, Volume I: User-Level ISA, Document
--- a/introduction.tex
+++ b/introduction.tex
@@ -79,7 +79,7 @@ functionality.
     \item[XLEN]
         Width of a hart's integer register file.
     \item[virtual address]
-        An address as a hart sees it. If there is address translation this
+        An address as a hart sees it. If the hart is using address translation this
         may be different from the physical address. If there is no
         translation then it will be the same.
 \end{description}
--- a/introduction.tex
+++ b/introduction.tex
@@ -28,7 +28,7 @@ functionality.
         JTAG instruction that selects a single bit data register, also called BYPASS.
     \item[component]
         A RISC-V core, or other part of a hardware platform.
-        Typically all components will all be connected to a single system
+        Typically all components will be connected to a single system
         bus.
     \item[CSR]
         Control and Status Register.
--- a/introduction.tex
+++ b/introduction.tex
@@ -51,13 +51,14 @@ functionality.
     \item[IR]
         JTAG Instruction Register.
     \item[JTAG]
-        Joint Test Action Group, referring to IEEE 1149.
+        Refers to work done by IEEE's Joint Test Action Group, described in
+        IEEE 1149.1.
     \item[NAPOT]
         Naturally Aligned Powers-Of-Two.
     \item[NMI]
         Non-Maskable Interrupt.
     \item[hardware platform]
-        A single integrated circuit consisting of one or more \emph{components}.
+        A single system consisting of one or more \emph{components}.
     \item[physical address]
         An address that is directly usable on the system bus.
     \item[SBA]
--- a/introduction.tex
+++ b/introduction.tex
@@ -51,7 +51,7 @@ functionality.
     \item[IDCODE]
         32-bit Identification CODE, and a JTAG instruction that returns the IDCODE value.
     \item[IR]
-        Instruction Register.
+        JTAG Instruction Register.
     \item[JTAG]
         Joint Test Action Group.
     \item[MIE]
--- a/introduction.tex
+++ b/introduction.tex
@@ -23,8 +23,6 @@ functionality.
 \begin{description}[style=nextline]
     \item[AMO]
         Atomic Memory Operation.
-    \item[BIST]
-        Built-In Self Test.
     \item[BYPASS]
         JTAG instruction that selects a single bit data register, also called BYPASS.
     \item[component]
@@ -34,13 +32,13 @@ functionality.
     \item[CSR]
         Control and Status Register.
     \item[DM]
-        Debug Module.
+        Debug Module (see Section \ref{dm}).
     \item[DMI]
-        Debug Module Interface.
+        Debug Module Interface (see Section \ref{dmi}).
     \item[DR]
         JTAG Data Register.
     \item[DTM]
-        Debug Transport Module.
+        Debug Transport Module (see Section \ref{dtm}).
     \item[DXLEN]
         Debug XLEN, which is the widest XLEN a hart supports, ignoring the
         current value of \Fmxl in \Rmisa.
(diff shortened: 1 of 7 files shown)


introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+    \item[BIST]
+        Built-In Self Test.
+    \item[BYPASS]
+        JTAG instruction that selects a single bit data register, also called BYPASS.
+    \item[component]
+        A RISC-V core, or other part of a platform.
+        Typically all components will all be connected to a single system
+        bus.
+    \item[CSR]
+        Control Status Register.
```
mwachs5, line 35:
```suggestion
        Control/Status Register.
```

Isn't this an "and/or" type thing?

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+        Debug XLEN, which is the widest XLEN a hart supports, ignoring the
+        current value of \Fmxl in \Rmisa.
+    \item[GPR]
+        General Purpose Register.
+    \item[hart]
+        A hardware thread in a RISC-V core.
+    \item[IDCODE]
+        32-bit Identification CODE, and a JTAG instruction that returns the IDCODE value.
+    \item[IR]
+        Instruction Register.
```
mwachs5, line 54:
```suggestion
        JTAG Instruction Register.
```

?

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+        Typically all components will all be connected to a single system
+        bus.
+    \item[CSR]
+        Control Status Register.
+    \item[DM]
+        Debug Module.
+    \item[DMI]
+        Debug Module Interface.
+    \item[DR]
+        Data Register.
```
mwachs5, line 41:
```suggestion
        JTAG Data Register.
```
?

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+    \item[hart]
+        A hardware thread in a RISC-V core.
+    \item[IDCODE]
+        32-bit Identification CODE, and a JTAG instruction that returns the IDCODE value.
+    \item[IR]
+        Instruction Register.
+    \item[JTAG]
+        Joint Test Action Group.
+    \item[MIE]
+        Machine Interrupt Enable register; also a bit in mstatus.
```
mwachs5, line 58:
```suggestion
        Machine-Mode Interrupt Enable register; also a bit in mstatus.
```

When using `MIE` do you always mean the bit in `mstatus`? Would you use `mie` if you were talking about the actual register?

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+    \item[RV32]
+        RISC-V core with XLEN of 32.
+    \item[RV64]
+        RISC-V core with XLEN of 64.
+    \item[SBA]
+        System Bus Access.
+    \item[TAP]
+        Test Access Port.
+    \item[XLEN]
+        Width of the core's integer register file.
```
mwachs5, line 80:
```suggestion
        Width of a hart's integer register file.
```

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+    \item[virtual address]
+        An address as a hart sees it. If there is address translation this
```
mwachs5, line 82:
```suggestion
        An address as a hart sees it. If the hart is using address translation this
```

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
-\subsection{Context}
+\begin{description}[style=nextline]
+    \item[AMO]
+        Atomic Memory Operation.
+    \item[BIST]
+        Built-In Self Test.
+    \item[BYPASS]
+        JTAG instruction that selects a single bit data register, also called BYPASS.
+    \item[component]
+        A RISC-V core, or other part of a platform.
```
mwachs5, line 31:
```suggestion
        A RISC-V core, or other part of a platform.
```

should this be core or hart?

introduction.tex
pdonahue-ventana: I agree on the distinction between MIE and mie.  (I find it confusing but I think it's too late now.)

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+    \item[NAPOT]
+        Naturally Aligned Powers-Of-Two.
+    \item[NMI]
+        Non-Maskable Interrupt.
+    \item[OpenOCD]
+        Open On-Chip Debugger.
+    \item[PC]
+        Program Counter.
+    \item[platform]
+        A single integrated circuit consisting of one or more \emph{components}.
```
pdonahue-ventana, line 68: Is this a "platform" or a "hardware platform"?

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+    \item[BYPASS]
+        JTAG instruction that selects a single bit data register, also called BYPASS.
+    \item[component]
+        A RISC-V core, or other part of a platform.
+        Typically all components will all be connected to a single system
+        bus.
+    \item[CSR]
+        Control Status Register.
+    \item[DM]
+        Debug Module.
```
pdonahue-ventana, line 37: Can this include a link to the DM chapter?  Same question for DMI and DTM and SBA.

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
-A \emph{virtual address} is an address as a hart sees it. If there is address
-translation this may be different from the \emph{physical address}. If there is
-no translation then it will be the same.
-
-\subsection{Context}
+\begin{description}[style=nextline]
+    \item[AMO]
+        Atomic Memory Operation.
+    \item[BIST]
+        Built-In Self Test.
```
pdonahue-ventana, line 27: BIST is only used in one place in the document so it's probably easier to just say "BIST (built-in self test)" in that one location instead of having the reader refer to the glossary.

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+    \item[JTAG]
+        Joint Test Action Group.
+    \item[MIE]
+        Machine Interrupt Enable register; also a bit in mstatus.
+    \item[NAPOT]
+        Naturally Aligned Powers-Of-Two.
+    \item[NMI]
+        Non-Maskable Interrupt.
+    \item[OpenOCD]
+        Open On-Chip Debugger.
```
pdonahue-ventana, line 64: This is only used once in the document and given that context I don't this glossary entry is necessary.

introduction.tex
```diff
@@ -20,22 +20,71 @@ \chapter{Introduction}
...
+    \item[physical address]
+        An address that is directly usable on the system bus.
+    \item[RV32]
+        RISC-V core with XLEN of 32.
+    \item[RV64]
+        RISC-V core with XLEN of 64.
+    \item[SBA]
+        System Bus Access.
+    \item[TAP]
+        Test Access Port.
```
pdonahue-ventana, line 78: This doesn't seem self-explanatory.  Maybe "see IEEE 1149" (or whatever) on both this and the JTAG definition.

pdonahue-ventana (commented):
I think that it's useful to define things like DM, DXLEN, IDCODE, and TAP.  I would remove things like hart, PC, MIE, RV32/64, and XLEN which are all defined in the base architecture.  This spec must presume some familiarity with the priv/unpriv architecture or else it would have to define M-mode, mepc, ebreak, exception, trap, and a hundred other things.

introduction.tex
timsifive: The privspec defines this register as "Machine interrupt-enable register." so -mode seems out of place. In this spec, we always style mie lowercase, and never refer to the bit in mstatus.

introduction.tex
timsifive: I agree with @pdonahue-ventana

introduction.tex
timsifive: Looking it up, this is a "hardware platform." But do we use that term anywhere?

timsifive: Based on this, I think I need to look at each use of "platform" and probably change it to "hardware platform." Do you agree?

introduction.tex
timsifive: Done.

introduction.tex
timsifive: Done.

introduction.tex
timsifive: Done.

introduction.tex
```diff
@@ -20,22 +20,57 @@ \chapter{Introduction}
...
+    \item[IR]
+        JTAG Instruction Register.
+    \item[JTAG]
+        Joint Test Action Group, referring to IEEE 1149.
+    \item[NAPOT]
+        Naturally Aligned Powers-Of-Two.
+    \item[NMI]
+        Non-Maskable Interrupt.
+    \item[hardware platform]
+        A single integrated circuit consisting of one or more \emph{components}.
```
pdonahue-ventana, line 60: I didn't notice this the first time but it doesn't seem like it's necessarily a single integrated circuit.  Multi-chip modules, multi-socket systems, etc.  Fifteen years ago I worked on a multi-board system with seamless cache coherence between cores on the same chip, cores on different chips on the same board, and cores on chips on different boards.

introduction.tex
```diff
@@ -20,22 +20,57 @@ \chapter{Introduction}
...
+    \item[GPR]
+        General Purpose Register.
+    \item[hart]
+        A hardware thread in a RISC-V core.
+    \item[IDCODE]
+        32-bit Identification CODE, and a JTAG instruction that returns the IDCODE value.
+    \item[IR]
+        JTAG Instruction Register.
+    \item[JTAG]
+        Joint Test Action Group, referring to IEEE 1149.
```
ernie-sifive, line 54:
```suggestion
        Joint Test Action Group, referring to IEEE 1149.1.
```

ernie-sifive (approved):
One additional term that could be useful:
```
    \item[TM]
        Trigger Module (see Section \ref{tm}).
```

introduction.tex
timsifive: I consciously didn't include .1, thinking that .7 should also be included. But maybe .7 really is cJTAG only?

introduction.tex
timsifive: Sure. So what do I call it? A single system? A thing?

introduction.tex
timsifive: How about "Refers to work done by IEEE's Joint Test Action Group, described in IEEE 1149.1."

ernie-sifive: That sounds good.  The Debug Spec DTM section only describes an 1149.1 implementation, so it seems correct to refer to that as opposed to "1149".

introduction.tex
pdonahue-ventana: I'd go with system.

introduction.tex
pdonahue-ventana: I have always found it annoying and confusing that RISC-V uses the word "platform" both for a particular implementation and for standards such as the Unix platform.  But we're stuck with that so "hardware platform" seems appropriate.

Now that I'm looking, I notice that this spec uses both "platform" and "system" to describe the overall implementation:
- "ndmreset resets all the harts in the system"
- "Harts are nonexistent if they will never be part of this system"
- "In most platforms there will only be one DM"
- "There may be multiple DTMs in a single platform."

Is there a distinction?  Maybe "system" (like in the title of chapter 2) means the group of components involved in debug but not all of the components in the whole chip/board/whatever?  Or maybe there's really no distinction?  The base architecture also uses the word "system" so maybe I'm overthinking this.

timsifive: I don't think there's any intentional use of platform vs. system. They should probably all be "hardware platform."

timsifive:
Can somebody give this a final once-over with my latest changes?

introduction.tex
```diff
@@ -9,33 +9,72 @@ \chapter{Introduction}
...
+        JTAG Instruction Register.
+    \item[JTAG]
+        Refers to work done by IEEE's Joint Test Action Group, described in
+        IEEE 1149.1.
+    \item[NAPOT]
+        Naturally Aligned Powers-Of-Two.
+    \item[NMI]
+        Non-Maskable Interrupt.
+    \item[hardware platform]
+        A single system consisting of one or more \emph{components}.
```
ernie-sifive, line 62: The rest of this list is alphabetized.  When this item was changed from `platform` to `hardware platform`, it is now in the wrong place.

ernie-sifive (approved):

introduction.tex
```diff
@@ -9,33 +9,72 @@ \chapter{Introduction}
...
-\subsection{Context}
+\begin{description}[style=nextline]
+    \item[AMO]
+        Atomic Memory Operation.
+    \item[BYPASS]
+        JTAG instruction that selects a single bit data register, also called BYPASS.
+    \item[component]
+        A RISC-V core, or other part of a hardware platform.
+        Typically all components will all be connected to a single system
+        bus.
```
pdonahue-ventana, line 32: Not a big deal but:
```suggestion
        Typically all components will be connected to a single system
        bus.
```

pdonahue-ventana (approved):
Other than my non-critical wording suggestion and Ernie's alphabetical order comment, this looks good.

PR #611, opened 2020-12-29
Move dret completely into the appendix.

timsifive:
Suggested by Ken Dockser, since dret has no meaning outside specific
implementations.

### DIFF
--- a/implementations.tex
+++ b/implementations.tex
@@ -62,6 +62,9 @@ write to \RdmDataZero, take effect before the DM returns \FdmAbstractcsBusy
 to 0.
 
 To resume execution, the debug module sets a flag which causes the hart to execute a {\tt dret}.
+{\tt dret} is an instruction that only has meaning while in Debug Mode and
+not executing from the Program Buffer.
+Its recommended encoding is 0x7b200073.
 When {\tt dret} is executed, \Rpc is restored from \RcsrDpc and normal execution resumes at the
 privilege set by \FcsrDcsrPrv.
 
--- a/core_debug.tex
+++ b/core_debug.tex
@@ -33,8 +33,8 @@ How Debug Mode is implemented is not specified here.
     timers are stopped.
 \item The {\tt wfi} instruction acts as a {\tt nop}.
 \item Almost all instructions that change the privilege level have \unspecified\ 
-    behavior.  This includes {\tt ecall}, {\tt mret}, {\tt sret}, {\tt uret},
-    and {\tt dret}.  (To change the privilege level, the debugger can write
+    behavior.  This includes {\tt ecall}, {\tt mret}, {\tt sret}, and {\tt uret}.
+    (To change the privilege level, the debugger can write
     \FcsrDcsrPrv and \FcsrDcsrV in \RcsrDcsr). The only exception is {\tt ebreak}, which ends
     execution of the Program Buffer when executed.
 \item All control transfer instructions may act as illegal instructions if
@@ -148,17 +148,6 @@ executed.
         \FcsrMcontrolMprv in \Rmstatus is cleared.
     \item The hart is no longer in debug mode.
 \end{steps}
-
-\section{{\tt dret} Instruction} \label{dret}
-
-To assist with certain types of implementations (see \ref{execution_based}),
-a new instruction is defined: {\tt dret}. It has an encoding of 0x7b200073.
-
-Execution of this instruction in the program buffer causes \unspecified\ behavior.
-Executing {\tt dret} outside of Debug Mode causes an illegal instruction exception.
-
-It is defined in this specification only to reserve the opcode and
-allow for reusable Debug Module implementations.
 
 \section{XLEN}
 


implementations.tex
```diff
@@ -62,6 +62,9 @@ \section{Execution Based} \label{execution_based}
 to 0.
 
 To resume execution, the debug module sets a flag which causes the hart to execute a {\tt dret}.
+{\tt dret} is an instruction that only has meaning while in Debug Mode and
```
pdonahue-ventana, line 65: I think that "only has meaning" loses the illegal instruction exception behavior in non-debug-mode.  Can you copy this sentence down here: "Execution of this instruction outside of Debug Mode causes an illegal instruction exception."

I don't think that the PB execution sentence needs to be copied here because the point of that was to tell debuggers to avoid dret in the PB.  But this "only has meaning" outside the PB sentence basically says the same thing (and that instruction is otherwise illegal and sane debuggers shouldn't be executing illegal instructions anyway).

implementations.tex
```diff
@@ -62,6 +62,9 @@ \section{Execution Based} \label{execution_based}
...
+not executing from the Program Buffer.
+Its recommended encoding is 0x7b200073.
```
ernie-sifive, line 67: Are we really changing dret from a specified encoding to a recommended encoding?

timsifive:
Here's the email thread, condensed:

>>> You mentioned that an instruction encoding was reserved for debug implementations. Are you talking about EBREAK or something else?

>> No, I'm talking about dret. One of the implementations the debug suggests (which is a popular one) relies on this instruction. To prevent a collision, the debug spec reserves 0x7b200073 for this instruction.

> It seems like this is an optional instruction that only applies to debug and furthermore is not intended to exist in code as it is only executed as the result of the debug module setting a flag. I think we need to formalize it as such rather than treating it as a reserved opcode.

My takeaway is that dret is an implementation detail, so we cannot say anything about its behavior outside of Debug Mode, nor can we require an encoding.

The only argument to be more strict is that it would allow Debug ROM to be shared among multiple hart implementations, but the Debug ROM is already very dependent on the memory map that it's closely tied to an implementation regardless.

pdonahue-ventana:
Frankly, an implementation could use an MRET in debug mode to return to normal execution.  Just as there was only a [single ERET instruction](https://github.com/riscv/riscv-isa-manual/blob/master/src/machine.tex#L2426) in older versions of the architecture until it was broken into separate mode-specific exception returns, the MRET instruction could use dpc/dcsr.prv or mepc/mstatus.MPP depending on current mode (since there is no need to virtualize the debug ROM code and run it in some other mode).  So I agree that DRET is an implementation detail and even ROM-based implementations don't necessarily need a dedicated DRET.

On further consideration, there should be no requirement that DRET in non-debug-mode raises an illegal instruction exception.  The DRET encoding isn't normally required to trap (the base architecture says "The behavior upon decoding a reserved instruction is UNSPECIFIED") so I don't think that ROM-based debug implementations must absolutely trap it.  Of course, sane implementations probably want to trap (or at least make it a NOP) but the RISC-V philosophy is that this choice is up to the implementation.

ernie-sifive:
Ok, so this proposal is basically saying that a feature used in only some debug implementations does not rise to the level of reserving an opcode for the purpose, meaning this opcode is available for future extensions unrelated to debug.

If we apply this logic consistently, then the same would be true of some of the CSRs such as dscratch0 and dscratch1.  These are only useful to an execution-based debug implementation, so by the above logic, shouldn't these CSR indexes be available for other uses in other implementations or to be allocated by future extensions?

If one these architectural resources is used in a future extension, existing implementations that adopt that extension would have to be changed.  If DRET is being demoted, then will DRET be removed from standard GNU tools?  https://github.com/riscv/riscv-binutils-gdb/blob/riscv-binutils-2.35/include/opcode/riscv-opc.h#L248

allenjbaum:
Isn't dscratch0,1 in the Debug-mode_only set of CSRs, as defined by the priv spec?
IF so, that's already covered: outside of Dmode accesses should trap.
DRET, on the other hand, isn't defined in the priv spec. 
In theory, it could do anything someone wants it to do outside of Dmode as long as it behaves as specified in Dmode, and everything would be backwards compatible. 
It just shouldn't be some kind of backdoor that does a priv escalation....

timsifive:
> If we apply this logic consistently, then the same would be true of some of the CSRs such as dscratch0 and dscratch1. These are only useful to an execution-based debug implementation, so by the above logic, shouldn't these CSR indexes be available for other uses in other implementations or to be allocated by future extensions?

I would agree with you, but the language in the spec says dscratchN can be available to program buffer programs. In the definition of nscratch the spec says "Number of dscratch registers available for the debugger to use during program buffer execution, starting from dscratch0. The debugger can make no assumptions about the contents of these registers between commands." I'm pretty sure dscratch being public is vestigial from 0.11, and this language is added because the resources exist and then somebody can use them. I don't think we can change that and still claim the spec is stable. Something for 1.1?

> If one these architectural resources is used in a future extension, existing implementations that adopt that extension would have to be changed.

For dret, that is not the case. Whatever encoding is chosen for dret, we only recommend a given behavior in the Debug ROM. Everywhere else the instruction would have its new function.

As Allen points out, dscratch is already in a region reserved for debug CSRs, so we're in the clear there as well.

> If DRET is being demoted, then will DRET be removed from standard GNU tools? https://github.com/riscv/riscv-binutils-gdb/blob/riscv-binutils-2.35/include/opcode/riscv-opc.h#L248

Presumably, yes.

ernie-sifive (approved):

PR #612, opened 2021-01-22
hartsel and hasel are WARL

pdonahue-ventana:
hasel is WARL since it may be tied off\
hartsel is WARL because not all bits have to be implemented

(Found by @benscotstaveley)


ernie-sifive (approved):
Looks right to me

PR #613, opened 2021-01-27
Clarify aarpostincrement wrapping

pdonahue-ventana:
See https://lists.riscv.org/g/tech-debug/message/339


timsifive:
It would be nice to allow implementations to use a <16-bit adder, especially since most implementations won't have any valid registers > 0x1fff. Can we phrase this so that the next value must be either the current value plus one, or zero?

timsifive:
Actually, that's too permissive. We don't want to allow 0x1010 to wrap to 0.

The next value may be 0 when, on this implementation, there are no higher values referring to registers that exist, and the value is 0xffff, 0x7fff, 0x3fff, or 0x1fff.

pdonahue-ventana:
Is this field WARZ?  I think so but I could be wrong.  WARZ says that a debugger can write any value but doesn't say that it turns into a legal value like WARL does.  Without this behavior I wasn't sure if you can implement fewer than 16b.  If WARZ does sanitize the value like WARL then I agree with you, though I think there's no way for a debugger to discover where the wrap occurs so it's not really a useful feature anyway.

ernie-sifive:
> I think there's no way for a debugger to discover where the wrap occurs so it's not really a useful feature anyway.

I agree this is not a useful feature.  How about just saying that auto-increment behavior when regno=0xffff is undefined?

ernie-sifive:
@pdonahue-ventana That behavior sounds good to me.

timsifive:
The last wording Paul suggested sounds great.

As pointed out, wrapping isn't useful to the debugger. I think the original text was added because somebody wanted to know what the behavior is, and I picked what seemed easiest on the hardware. This is even easier for hardware to implement, and less ambiguous to boot.

pdonahue-ventana:
I'll update it sometime today.

timsifive (approved):

PR #614, opened 2021-01-29
Clarify MCONTROL/MCONTROL6 'timing' description

tmw-wdc:
In a pipelined core, executing an instruction takes multiple cycles. It is also possible for an instruction to be executed but not committed or executed out-of-order with respect to the preceding/succeeding instruction.  The 'timing' bit description as written appears to exclude pipelining, speculation, and out-of-order execution.

Changed 'executed' to 'committed' three times each in 'timing' description of MCONTROL and MCONTROL6 registers.


timsifive (approved):
Thank you.

PR #621, opened 2021-03-11
Clarify DM behavior for non-existent registers

pdonahue-ventana:
Clarify for https://lists.riscv.org/g/tech-debug/message/377


timsifive (approved):
Looks good. Thanks!

PR #624, opened 2021-03-20
Clarify dpc for non-mcontrol triggers

pdonahue-ventana:
I think that dpc is always the next instruction to execute (which on timing=0 also happens to be the instruction that hit the trigger).  But I wasn't sure if _load data before_ was a special case that violates that so I just did the minimal clarification.


timsifive:
Your comment talks about "load data before" but the actual change only clarifies that timing is only relevant for mcontrol and mcontrol6 triggers.

pdonahue-ventana:
I think that the change is correct.  My comment was questioning whether it could even be simplified further and I'm wondering if you have advice on that.

Is it true that if we enter debug mode due to the TM then dpc always points to the next instruction to be executed?  I don't think that the fact that normal "before" triggers have dpc point to the instruction that caused the trigger is a special case but more of a natural consequence of the way that "before" triggers work paired with the "dpc always points to the next instruction to be executed" rule.

Then I started to doubt that simplification with respect to "load data before" triggers.  I'm not sure that doubt was really justified.  It seems that any load that hit such a trigger is the next instruction to execute since it hasn't updated the destination register yet.  There are warnings about this type of trigger and I/O devices but that's a separate concern that doesn't really affect dpc (because it's the next instruction to execute with respect to the register file but it has already executed with respect to memory).

So I'm wondering if this description could be simplified to just:
> The address of the next instruction to be executed at the time that debug mode was entered.

Or maybe something like:
> The address of the next instruction to be executed at the time that debug mode was entered.  If the trigger is \RcsrMcontrol or \RcsrMcontrolSix and \FcsrMcontrolTiming is 0, this also corresponds to the address of the instruction which caused the trigger to fire.

I can make either change and update the PR if you agree.

pdonahue-ventana:
I also notice that this table doesn't list the dcsr.cause=5 (resethaltreq) and dcsr.cause=6 (halt group) cases.  Presumably, they're both the next instruction to execute (which in the resethaltreq case is the address of the reset vector).  I think that the rule in all these cases really is that dpc always points to the next instruction to execute.  The only exception is ebreak where it points to the ebreak itself.

(The ebreak case really isn't different since it's the next PC expected to be executed assuming the debugger changes the instruction before resuming.  And arguably the ebreak hasn't executed yet since I seem to recall that minstret isn't incremented on ebreak.)

ernie-sifive:
Hi Paul,

I agree that dpc is always the next instruction to execute (or the first instruction that _didn't_ execute) prior to entering debug mode.  EBREAK behavior is consistent with this - like other instructions that cause traps, the instruction does not retire.

pdonahue-ventana:
I just noticed that the ball was in my court on this.  @timsifive: Take a look at the latest wording.

timsifive (approved):

PR #625, opened 2021-03-20
clarify abstractauto

pdonahue-ventana:
Clarify abstractauto per https://lists.riscv.org/g/tech-debug/message/413


ernie-sifive (approved):
This change looks good to me.

PR #629, opened 2021-04-19
Update Run/Halt Debug State Machine.

timsifive:
Bring the diagram up-to-date with what the spec says. Added abstract
command error states when an abstract command is issued while running.
Removed individual abstract commands since it got to be too messy.

In response to thread on mailing list subject:"A query regarding the
State Machine".


timsifive:
To see the new graph, go to "Files Changed" on this PR. Then click the ... next to the .pdf file, and select "View file."

pdonahue-ventana:
1. "M/U/S mode" should probably also add VS/VU.  Or just something like "modes other than debug mode" that's more generic.
2. What about quick access?  Can you just add a note that quick access goes through the "halting"->"halted"->"halted abstract command" path or maybe mention quick access on the haltreq/resumereq arcs?
3. Should there be an arc from "running error wait" to a state that's similar to "halting" but with cmderr!=0?  I think when cmderr!=0 and you're running that you can still halt.  Similar with "halted error wait" and "resuming" since you can resume with cmderr!=0.  Just like there are two sets of "error busy" and "error wait" and "abstract command" states, do there also need to be two sets of the yellow and green states where cmderr!=0?

timsifive:
> 1 & 2

Done.

> Should there be an arc from "running error wait" to a state that's similar to "halting" but with cmderr!=0? I think when cmderr!=0 and you're running that you can still halt. Similar with "halted error wait" and "resuming" since you can resume with cmderr!=0. Just like there are two sets of "error busy" and "error wait" and "abstract command" states, do there also need to be two sets of the yellow and green states where cmderr!=0?

Technically, yes. And likewise from halted error wait to resuming. And I think also from the error busy states. I think adding all that would make the diagram too messy for states that generally won't occur. (You could write a debugger that issues abstract commands without checking for results, but I can't really think of a real-world use for that.) So I'm inclined to leave that as-is to keep the diagram readable.

pdonahue-ventana:
I agree that it's unlikely that a debugger would, for example, do something that causes cmderr!=0 and then resume without clearing the error.  At the risk of having too many notes, can you add a note that says that this diagram is not exhaustive?  Or maybe have the caption say that these are the "normal" subset of actual states?  This whole thing came up when someone noticed that the state machine was not completely accurate so we should have a disclaimer somewhere if it's not intended to be 100% accurate.

timsifive:
Done. Thanks.

pdonahue-ventana (approved):
Looks good.

Issue #631, opened 2021-05-07
Misssing arcs in State machine diagram


odedy77:
The State machine diagram in Figure 3.1 does not reflect the trigger_with_action_1_fires, halt_request/resume_request from resume groups or external input triggers, and resuming as a result of DRET instruction.

timsifive:
Please reopen if the change didn't address this issue adequately.

PR #632, opened 2021-05-11
Further diagram update.

timsifive:
Hopefully addresses #631 and make it more exhaustive.


pdonahue-ventana:
I like this concept and it seems to more fully cover everything.

I think that abstractcs.busy should be removed from the left state machine and that it should only be part of the right state machine.  You can be in running state (and possibly halting or resuming) with abstractcs.busy=1 if the implementation supports abstract commands without halting.  You can certainly be in halted state with abstractcs.busy=1.

I don't think that "software breakpoint" is the fully descriptive term but "software breakpoint when dcsr.ebreakX=1 and executing in mode X" is way too long.  Maybe "enabled software breakpoint" or something like that?  Or just the existing "software breakpoint" but with an asterisk that says "if enabled in dcsr."

I went through the list of halt causes.  Should resethaltreq also be added to the normal execution->halting arc?  It's arguable whether you were even in normal execution state to start with.  How about something that indicates that:
- normal execution is the initial state if resethaltreq=0
- halting is the initial state if resethaltreq=1
- abstract command idle is the initial state on the other state machine

pdonahue-ventana:
I didn't notice this before but "abstract command idle" is not necessarily non-debug after the two state machines were split apart.  So it probably should have a different color.

timsifive:
I think it is non-debug. It is the state that machine is in when there's no debugging going on (in that state machine).

What color/group do you think it belongs to?

pdonahue-ventana (approved):
I interpreted "non-debug" as meaning modes other than debug mode.  But I see what you mean.  This diagram does describe the states accurately and subtleties like what mode you're in are clear in the text but are not represented here.

pdonahue-ventana:
> What if I rename "Non-Debug" to "Normal Execution"?

I think it should stay the way it is.  If you label it as "normal execution" then you're reusing the name of one of the states.  Someone may misconstrue the meaning of that and think that "abstract command idle" having the "normal execution" _color_ might have something to do with the "normal execution" _state_ of the other state machine and that they're not independent state machines.  I think that "non-debug" is probably less confusing and I now don't think that my previous point was a big enough problem to worry about.

(It's hard to anticipate what readers may think so I'm trying to imagine various ways that someone seeing this for the first time might interpret things.)

> I'm not sure what subtleties you're referring to. The U/S/M mode that actually reflects the currently executing code on the hart?

Those modes vs. debug mode.  Maybe it was hard to parse but I meant "subtleties (like what mode you're in) are clear in the text but are not represented here."  Even though orange means debug mode and blue/gray means M/S/U in the left state machine, I don't think that this needs to be called out.  Like I said, it's clear in the text and this diagram isn't meant to replace the rest of the spec.

I don't remember paying much attention to the diagram before but now I think that it's clear, concise, and complete.

PR #633, opened 2021-05-11
Fix #630

### LINKED ISSUE #630: Single Step Clarification request
In 4.4.1  please change  
“An external debugger can cause a halted hart to execute a single instruction or trap and then re-enter Debug Mode by setting step before setting resumereq.” 

 to   something like:  

“An external debugger can cause a halted hart to execute a single instruction or trap and then re-enter Debug Mode by setting step before resuming.”  

and maybe make it clearer, add another sentence:
“If the step bit is set when a hart resumes then it will single step, regardless of the reason for resuming.”

pdonahue-ventana:
Used the language that @odedy77 suggested.

Also fixed a typo.


timsifive (approved):
Thank you!

PR #634, opened 2021-05-13
External debuggers should set dmode.

timsifive:


pdonahue-ventana (approved):
Looks good.  No hardware change but it gives external debuggers advice on how to play nice using existing hardware.

PR #635, opened 2021-05-14
Divide the spec into ISA and non-ISA.

timsifive:
There are separate definition-of-done requirements for ISA vs. non-ISA
specs. Our spec contains both, so clearly mark what is what.

### DIFF
--- a/core_debug.tex
+++ b/core_debug.tex
@@ -1,4 +1,4 @@
-\chapter{RISC-V Debug}
+\chapter{RISC-V Debug, ISA}
 \label{sec:core_debug}
 
 Modifications to the RISC-V core to support debug are kept to a minimum.  There
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -1,4 +1,4 @@
-\chapter{Debug Module (DM)} \label{dm}
+\chapter{Debug Module (DM), non-ISA} \label{dm}
 
 \begin{steps}{The Debug Module implements a translation interface between abstract debug
     operations and their specific implementation. It might support the following
--- a/dtm.tex
+++ b/dtm.tex
@@ -1,4 +1,4 @@
-\chapter{Debug Transport Module (DTM)} \label{dtm}
+\chapter{Debug Transport Module (DTM), non-ISA} \label{dtm}
 
 Debug Transport Modules provide access to the DM over one or more transports
 (e.g.\ JTAG or USB).
--- a/introduction.tex
+++ b/introduction.tex
@@ -178,10 +178,16 @@ incompatible, but unlikely to be noticeable:}
 \subsection{Structure}
 
 This document contains two parts. The main part of the document is the
-specification, which is given in the numbered sections. The second part
-of the document is a set of  appendices. The information
+specification, which is given in the numbered chapters. The second part
+of the document is a set of appendices. The information
 in the appendices is intended to clarify and provide examples, but is
 not part of the actual specification.
+
+\subsection{ISA vs. non-ISA}
+
+This specification consists of both ISA and non-ISA parts. Chapters whose
+contents are solely one or the other are labeled as such in their title.
+Chapters without such a label apply to both ISA and non-ISA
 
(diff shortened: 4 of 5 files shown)


timsifive:
I'm not quite happy with the language in the intro, but I'm unsure how to improve it.

pdonahue-ventana:
I think it would be good to say what ISA vs. non-ISA means.  Is there a definition of this term that keeps getting tossed around?  In poking around, I found this comment from you: https://docs.google.com/document/d/1bSJSnPfL17kDTL-CUB4xE32Zjp77RTAk1XOKXQF_3Qc/edit?disco=AAAAHD73F9Y

Intuitively, I think that "ISA is everything that affects encodings and behaviors of instructions that execute on harts" and "non-ISA is everything else."  There could be subtle things I'm missing.  Or maybe (at least in our case) "non-ISA is related to devices external to the hart" and "ISA is everything else."  But I think it would be helpful for there to be an official definition and then for the intro to contain that definition.

timsifive:
If there was a good definition I'd happily cut and paste it, but I haven't seen anything yet. Realistically this is just some internal RISC-V housekeeping and readers don't care at all.

The best definition I've seen is in a private email thread:
> This document as written applies to ISA specifications - namely specifications of what goes inside of a CPU (aka a RISC-V ‘hart’).  In contrast, non-ISA specifications are for features and components that reside elsewhere in the system.  Examples include: External Debug support, system-level RAS, interrupt controllers, IOMMUs, bus standards.  In a few cases (e.g. the RISC-V Debug architecture), the specification contains both ISA and non-ISA components.  The ISA components are subject to this Ratification Policy; the non-ISA components are subject to the following paragraph.

> Non-ISA specifications go through the same steps except the public review and final step of getting board approval. This is because they are ecosystem-oriented specifications and not ISA specifications.  RISC-V International hosts implementation documents and non-ISA specifications for RISC-V as a convenience but is chartered to ratify ISA specifications. It is up to the governing committee if they send the non-ISA spec for public review. Also note that there is a different definition of done checklist for non-ISA documents and specifications. If there are upstream review and approval requirements for non-ISA specs, then you must follow those rules and conventions

I'm not sure why that language did not end up in the ISA vs. non-ISA document.

pdonahue-ventana:
> Realistically this is just some internal RISC-V housekeeping and readers don't care at all.

I agree.  But the readers are exposed to it by calling out parts that fit in one category or the other.  So I think it's useful to either say what the categories mean or to say "these categories are internal housekeeping that you shouldn't care about" (which would tend to make readers question why they're being exposed to such details).

I pointed out your previous question to Mark.  He was pretty responsive and changed the policy to have the definitions.  It now says:
> ISA specifications only specify instructions, state and behavior and they are unique. Non-ISA specification may include other things like calling conventions, debug protocol message or field types, guidance on how to write architecture tests, etc, Non-ISA may have multiple ways or standards to do things (e.g. E-Trace and Nexus). Some existing specifications contain both ISA and NON-ISA specifications together. It will be up to the documentation SIG to determine the long term strategy of how to arrange ISA and Non-ISA specifications that depend on each other. The approval/ratification rules are different for ISA and Non-ISA specifications and can be found in the Ratification Policy.

I don't think that these definitions are 100% accurate but they're OK.  So what do you think about:
> This specification consists of both ISA and non-ISA parts.  ISA parts specify hart state and behavior.  Non-ISA parts specify non-hart state and behavior.

(Since dret was moved to the appendix, I don't think that any instructions are specified in our particular case.  I don't think that ISA in our case specifies all state and behavior so I changed Mark's definition to restrict that to hart state and behavior.  I was tempted to just say non-ISA is everything else but I think it's related to state and behavior of everything that's not in the ISA category.)

timsifive:
I like that suggestion.

timsifive:
Is this good to go?

pdonahue-ventana (approved):
Sorry for the delay.  This looks OK to me.

PR #637, opened 2021-06-02
Traps caused by action=0 triggers can be delegated via medeleg

pdonahue-ventana:
Rather than expanding "mepc" to be "mepc or sepc or vsepc (depending on delegation)" I just used the generic term.

This is based on https://lists.riscv.org/g/tech-debug/topic/83267261#482


timsifive (commented):
This looks good. I'm looking at the privileged spec to see if there's a better term than "exception PC" (which isn't mentioned anywhere). When talking about mepc it says "When a trap is taken into M-mode, mepc is written with the virtual address of the instruction
that was interrupted or that encountered the exception."

Does that mean that "trap PC" (also not mentioned anywhere) is a more accurate term? Maybe we should use "trap PC" and define it somewhere?

I'm just brainstorming here. Not sure what's the best solution. As usual it's hard to be both precise and concise.

I'll come back to this later, but maybe this shakes something loose in your head.

pdonahue-ventana:
The privileged spec uses the term "exception program counter" for all of those CSRs but I (perhaps improperly) shortened it to "exception PC" just to be more concise.  I think that "exception program counter" is fine or even something like "the appropriate exception program counter CSR" to indicate that it's whichever one makes sense given the delegation setup.  In the opposite direction, I considered simply saying "xepc" with the implication that you know what x is, but I didn't like that.  I'm OK having more words if it makes things clear and uses terms that people can search for in other specs.

timsifive (approved):
So it does (although only in the register name). That's probably as good as we're going to get.

PR #640, opened 2021-06-08
Labels must come after (or in) their caption.

### LINKED ISSUE #639: Fix reference to table number in section 5.3 (Priority)
I am using PDF from [linked](https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-stable.pdf) stable version. There is an incorrect reference to table in section 5.3 (Priority) of Chapter 5 (Trigger Module). In the first line it refers to table 5.3 (Trigger Registers). Instead, I think it should refer to table 5.2 (Synchronous exception priority in decreasing priority order).

timsifive:
Fixes #639.

### DIFF
--- a/trigger.tex
+++ b/trigger.tex
@@ -94,7 +94,6 @@ in the chain.
 
 \begin{table}[H]
 \centering
-\label{tab:priority}
 \begin{tabular}{|l|r|l|l|}
   \hline
   Priority      & Exception & Description & Trigger \\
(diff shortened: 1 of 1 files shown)


pdonahue-ventana (approved):
You win the LaTeX guru trophy.

timsifive:
Hah! Can I have a Google keywords trophy instead?

PR #641, opened 2021-06-08
Mention license in debug_defines.h.

### LINKED ISSUE #618: License of generated headers?
The `debug_defines.h` is currently spit out without any license attached, which is a Bad Thing. That should be explicitly marked as CC-BY-4.0 or released under a license more suitable for code (riscv-opcodes uses BSD-3, for example).

Not causing any issues now, but I've dealt with lawyers enough to want to nip this in the bud.

timsifive:
Fixes #618


pdonahue-ventana (commented):
Should it be "Attribution 4.0 International" or "Creative Commons Attribution 4.0 International"?

timsifive:
Hmm... maybe should be more verbose. In LICENSE it even says "Creative Commons Attribution 4.0 International Public License"

Issue #619, opened 2021-02-26
Design constrained halt group?


stanleyleee:
Hi all,
    In specification of 1.0.0 stable, it mentions that hardcoded halt group is explicitly allowed. However, I am wondering if the following scenario is also allowed.

- Hart 0,1 are pre-defined in halt group 1 and are required always in the same halt group

- Hart 2, 3 are pre-defined in halt group 2 and are required always in the same halt group

- Debugger writes dmcs2 to move hart 2 to halt group 1, but design also moves hart 3 to halt group 1 (which is not the intention of debugger).

Thanks

pdonahue-ventana:
Hard-coded values aren't writable.  If harts 2 and 3 are hard-coded into halt group 2 then the debugger cannot move hart 2 into halt group 1 (or any other halt group) and the design cannot move hart 3 into halt group 1 (or any other halt group).

stanleyleee:
Hi Donahue,
   
Thanks for your response that hardwire halt groups can not be modified.

It seems the case above should not be configured as hardwire halt groups.

If an implementation expect HARTs in the same CORE should enter halt state together because of shared resources,
Are there any suggested ways to provide the flexibility to configure COREs into same or different halt groups.

Many Thanks

pdonahue-ventana:
In your example implementation with two cores that each contain two harts, I think that you can certainly hardcode harts 0 and 1 to halt group 1 and harts 2 and 3 to halt group 2.  Or you could hardcode all four harts to group 1 but then you can't independently halt the independent cores.  Either way, things are hardcoded which eliminates the flexibility to decide whether you want to hart cores independently or together.

It seems like a reasonable use case to want the ability to dynamically put the two cores in separate groups or the same group.  All harts would start up in group 1 (since they can't be in group 0 and group 1 is the lowest group number they can be in).  You could put hart 0 in group 2 but then I think that in your implementation hart 1 ends up in group 2.  I don't think that the non-explicit change of the group for hart 1 is allowed by the spec unless there's something preventing you from selecting only hart 0 to start with.  @timsifive, do you have any ideas?

stanleyleee:
Hi,

Here is another question about halt, resume group and external trigger.

There is a description of dmcs2, "The DM external triggers available to add to halt groups may be the same as or distinct from the DM external triggers available to add to resume groups".

Which of the following cases satisfy the description?

1: An external trigger can be added to a halt group and a resume group at the same time.

[...]

PR #642, opened 2021-06-09
Unselected harts may change groups when hgwrite=1

timsifive:
Fixes the first issue brought up in #619.

Also remove some duplicate language.

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -672,8 +672,8 @@
             hart the DM will change its group to the value written to \FdmDmcsTwoGroup,
             if the hardware supports that group for that hart.
             Implementations may also change the group of a minimal set of
-            unselected harts in the same way, if that is necessary to work
-            around a hardware limitation.
+            unselected harts in the same way, if that is necessary due to
+            a hardware limitation.
 
             When 1 is written and \FdmDmcsTwoHgselect is 1, the DM will change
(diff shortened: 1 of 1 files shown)


xml/dm_registers.xml
```diff
@@ -668,12 +668,12 @@
...
         <field name="hgwrite" bits="1" access="W1" reset="-">
-            When \FdmDmcsTwoHgselect is 0, writing 1 changes the group of all
-            selected harts to the value written to \FdmDmcsTwoGroup.
-
             When 1 is written and \FdmDmcsTwoHgselect is 0, for every selected
             hart the DM will change its group to the value written to \FdmDmcsTwoGroup,
             if the hardware supports that group for that hart.
+            Implementations may also change the group of a minimal set of
+            unselected harts in the same way, if that is necessary to work
+            around a hardware limitation.
```
pdonahue-ventana, line 676: I wouldn't say that it's working around something since in those implementations it's an intentional decision.  Maybe:
```suggestion
            Implementations may also change the group of a minimal set of
            unselected harts in the same way, if that is necessary due to
            a hardware limitation.
```

pdonahue-ventana (approved):
Looks good.  And thanks for removing that first sentence.  Before reviewing the PR, I read the existing spec first.  I was scratching my head for a minute over the difference between the first and second sentences.

PR #643, opened 2021-06-09
hasel etc. are only ignored by abstract commands.

timsifive:
This fixes a contradiction pointed out in #619, in favor of the language
in dmcs2.


pdonahue-ventana (approved):
I went back and reread some of the comments in that issue and this seems like an appropriate change.

PR #649, opened 2021-06-14
more dots  riscv

odedy7:
more dots riscv


odedy7:
closing test fork.  please reject.

PR #654, opened 2021-06-25
xepc instead of exception PC

### LINKED ISSUE #652: Nomenclature suggestion: use xepc, xcause etc
To describe trap behavior when breakpoint exceptions have been delegated from Machine mode, the specification is using the imprecise term "exception PC" where it used to say `mepc`. The Fast Interrupt specification uses terminology `xepc`, `xcause` etc in this context, which I think is more precise and preserves the link with the CSR name better. I think it would improve things to adopt this naming convention.

pdonahue-ventana:
Fixes #652 

I don't think that we need to change mcause to xcause.  The only place I could find mcause was in the [definition of etrigger](https://github.com/riscv/riscv-debug-spec/blob/master/xml/hwbp_registers.xml#L928):
>  This trigger may fire on up to XLEN of the Exception Codes defined in \Rmcause (described in the Privileged Spec, with Interrupt=0).

It doesn't seem necessary to point to other cause registers since the codes in mcause are a superset of all xcause codes.


introduction.tex
```diff
@@ -72,6 +72,9 @@ \section{Terminology}
         An address as a hart sees it. If the hart is using address translation this
         may be different from the physical address. If there is no
         translation then it will be the same.
+    \item[\Rxepc]
+        The exception program counter CSR (e.g. mepc) that is appropriate for
```
timsifive, line 76: \Rmepc?

xml/hwbp_registers.xml
```diff
@@ -293,7 +293,7 @@
         <field name="timing" bits="18" access="WARL" reset="0">
             0: The action for this trigger will be taken just before the
             instruction that triggered it is committed, but after all preceding
-            instructions are committed. The exception PC or \RcsrDpc (depending
+            instructions are committed. The \Rxepc or \RcsrDpc (depending
```
timsifive, line 296: Just "\Rxepc", not "The \Rxepc".

Same comment 3 more times below.

introduction.tex
pdonahue-ventana: Changed.

xml/hwbp_registers.xml
pdonahue-ventana: Oops.  Changed.

timsifive (approved):
Thank you!

PR #655, opened 2021-06-28
Clarify postincrement/transfer behavior.

timsifive:
When transfer=0 but postincrement=1, it is undefined whether regno is
incremented or not.

See mailing list discussion subject:"aarpostincrement and transfer=0"


pdonahue-ventana (approved):
Seems reasonable to me.

PR #657, opened 2021-07-01
Attempt to build the document using github actions.

timsifive:
Travis stopped working, and it's always been a bit voodoo to me.


timsifive:
@pdonahue-ventana can you take a look at this? Travis being busted is causing every PR to fail its checks, and I'd like to have a replacement ready to go for when I turn it off.

pdonahue-ventana (approved):
Sorry, I was out.

I don't know enough about this to have a lot of feedback.  I assume that the `make` package is part of ubuntu-latest and doesn't need to be specified.  I also notice that graphviz and texlive-base used to be specified but no longer are.  I assume that texlive-base is autoinstalled when installing the other texlive package but I don't know about graphviz.

As long as this works, it's fine with me.  And we can fix bugs via other PRs as necessary.

Issue #653, opened 2021-06-23
Self-hosted debug and triggers inside handlers


pdonahue-ventana:
Copied from https://lists.riscv.org/g/tech-debug/topic/self_hosted_debug_and/83591174 so it doesn't get forgotten.

> The debug spec already contemplates that bad things happen when you set an action=0 trigger to fire in an M mode handler.  It suggests that this is a corner case but has two possible solutions: mask action=0 triggers in M mode when MIE=0 or implement tcontrol.mte/mpte.
> 
> I think that this problem is much larger than that.  In particular, if breakpoint exceptions are delegated to S mode then the same problem exists with triggers firing in an S mode handler (or an M mode handler).  Same with VS mode.  Also, if breakpoint exceptions are delegated to S mode and exception X is also delegated to S mode then if etrigger[X]=1 and you hit exception X (from S or U) then the breakpoint exception will happen right after exception X and will overwrite scause/sepc.  Same with VS mode.  Perhaps most importantly, tmexttrigger is completely asynchronous and may happen at inconvenient times while in any of these modes.
> 
> I suggest that if any of these things can occur (i.e. if tmexttrigger is supported or if breakpoint delegation is supported) then it be made mandatory to do one of these:
> 1. The hardware must prevent triggers with action=0 from firing while in M-mode and while
MIE in mstatus is 0. If medeleg [3]=1 then it must prevent triggers with action=0 from firing
while in S-mode and while SIE in sstatus is 0. If medeleg [3]=1 and hedeleg [3]=1 then it
must prevent triggers with action=0 from firing while in VS-mode and while SIE in vsstatus
is 0.
> 2. mte and mpte in tcontrol must be implemented. medeleg [3] must be hard-wired to 0.
> 
> Because tcontrol is not accessible to S-mode, the second option cannot be extended to
> accommodate delegation without adding additional S-mode and VS-mode CSRs (which I don't think is worth doing).
> 
> Thoughts?
> 
> 
> Thanks,
> 
> -Paul

pdonahue-ventana:
This was addressed by #660.

PR #660, opened 2021-07-23
Native triggers and reentrancy

pdonahue-ventana:
My proposal to address #653.

### DIFF
--- a/trigger.tex
+++ b/trigger.tex
@@ -142,33 +142,81 @@ be set, giving a debugger an opportunity to handle this case. What happens with
 trace actions when triggers with different actions are also firing is left to
 the trace specification.
 
-\section{Native M-Mode Triggers}
-\label{sec:mmtrigger}
+\section{Self-hosted Triggers}
+\label{sec:shtrigger}
 
-Triggers can be used for native debugging. On a fully featured hart, triggers
-will be set using \FcsrMcontrolU or \FcsrMcontrolS, and when firing they can cause a breakpoint exception
-to trap to a more privileged mode. It is possible to set triggers natively to
-fire in M-mode as well. In that case there is no higher privilege mode to trap
-to. When such a trigger causes a breakpoint exception while already in a trap
-handler, this might leave the hart unable to resume normal execution.
+Triggers can be used for self-hosted debugging when \FcsrMcontrolSixAction=0.
+If supported by the hart and desired by the debugger, triggers will often be
+programmed to have \FcsrMcontrolSixM=0 so that when they fire they cause a
+breakpoint exception to trap to a more privileged mode. That breakpoint
+exception can either be taken in M-mode or it can be delegated to a less
+privileged mode. However, it is possible for triggers to fire in the same
+mode that the resulting exception will be handled in.
 
-On full-featured harts this is a remote corner case that can probably be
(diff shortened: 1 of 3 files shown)


timsifive:
This reads like a good clarification, although it now requires one of the two solutions be implemented. Was that your intention? It's not strictly backwards compatible if we require one of the two solutions. Did I miss something and are the proposed solutions actually different?

pdonahue-ventana:
It was my intention to add that as a requirement if it's possible for any of the enumerated situations to exist.  The 0.13 spec said that this was a remote corner case but I think that's not really true, especially when delegation comes into play.  Without this, tmexttrigger with action=0 is useless and etrigger/itrigger with action=0 are definitely hampered.  We can, of course, debate these things.  If it remains optional in this spec (with just a "debuggers need to take care not to cause infinite loops of traps" warning) then I think that platform specs should make it mandatory.

I don't quite understand your last question.  The two solutions are definitely different from each other since the first doesn't require tcontrol.  The first solution is the same as 0.13 but with the addition of masking at lower privilege modes when appropriate.  The second solution is the same as 0.13 but with the addition of hard-wiring medeleg to prevent having the problem at lower privilege modes (which happens naturally for M or M+U implementations but could also work for M+S+U).

timsifive:
The change is OK by me, but it is definitely not a backwards-compatible change. We've been saying that the spec is stable and won't see functional changes between now and ratification.

So... We could change it anyway. We'd have to invent some extra process to discourage us from doing that in the future. We could start a 1.1 fork and put this change in there.

What do you think?

ernie-sifive:
I would argue that once the spec is "Stable" no further functional changes can be made.  Implementations of v1.0 already exist.  I think this change should be put into v1.1.

pdonahue-ventana:
I knew that would be controversial and I certainly understand why.

How about this?
- "Harts that support triggers with \FcsrMcontrolSixAction=0 **should** implement one of the following two solutions"
- change "should" to "must" in 1.1

I think that it's important to talk about the problem and at least provide guidance.  That's what the 0.13 spec did, though I think that it didn't go far enough since it didn't contemplate delegation (which is what this change addresses).

trigger.tex
```diff
@@ -142,33 +142,84 @@ \section{Priority}
...
+
+In these cases such a trigger may cause a breakpoint exception while already
+in a trap handler. This might leave the hart unable to resume normal execution
+because state such as \Rmcause and \Rmepc would be overwritten.
+\end{commentary}
+
+\begin{steps}{Harts that support triggers with \FcsrMcontrolSixAction=0
+should implement one of the following two solutions to solve the problem of
+reentrancy:}
+\item The hardware must prevent triggers with \FcsrMcontrolSixAction=0 from
```
timsifive, line 191: Must in these two items is confusing. I'd leave the line above (which now says should) and then describe the behavior without must/should. E.g. "The hardware prevents triggers ..."

timsifive (changes requested):
Overall this looks good. Just a language comment below.

trigger.tex
pdonahue-ventana: I obviously didn't reread the whole thing when making that must->should change.  I'll reread the whole thing and make changes along the lines of your suggestion.

timsifive (approved):

PR #665, opened 2021-08-13
Fix contradiction in icount description.

### LINKED ISSUE #662: Clearing for icount.m/s/u/vs/vu
Hello,

I see the following description in chapter 5.5.13 Instruction Count (debug-spec v1.0.0 stable). The following description is conflict.

icount description
`
When count is decremented from 1 to 0 then pending becomes set. When pending is set, the trigger fires just before any further instructions are executed in a mode where the trigger is enabled.
`

icount.count description
`When count is decremented to 0, the trigger fires. Instead of changing count from 1 to 0, it is also acceptable for hardware to clear m, s, u, vs, and vu. This allows count to be hard-wired to 1 if this register just exists for single step.
`

icount is for single-step and icount.count hardwired to 1. icount.m/s/u/vs/vu bits are cleared if icount.count is from 1 to 0. The pending bit is set because of changing count from 1 to 0. icount cannot fire in any mode because icount.m/s/u/vs/vu bits are cleared. I think that icount.m/s/u/vs/vu bits should be cleared when icount fires.

timsifive:
Specifically, after we added the pending bit, the description of a
possible 1-bit implementation of this register was not updated. As a
result the count description said that decrementing resulted in a
trigger, while in the introductory part we said that decrementing
resulted in pending getting set.

Rewrote it a bit, so it's hopefully easier to read.

Fixes #662.


xml/hwbp_registers.xml
```diff
@@ -801,6 +808,11 @@
         useful both for external debuggers and for software monitor programs.
         For that case it is not necessary to support \FcsrIcountCount greater
         than 1.
+
+        If \RcsrIcount is only implemented for single step, then it can be done
+        using a single bit that ties \FcsrIcountM, \FcsrIcountS, \FcsrIcountU,
+        \FcsrIcountVs, and \FcsrIcountVu together while \FcsrIcountCount is
+        hard-wired to 1.
```
pdonahue-ventana, line 815: Plus a pending bit?

Consider this:
- mstatus.MIE=0 and we're in M mode
- write icount action=0, m=s=u=vs=vu=1 (i.e. set that single bit), count=1 (which is hard-wired), pending=0
- execute a bunch of instructions.
- read icount.  I would expect to see pending=1.
- write mstatus.MIE=1
- I would expect the trap to happen before the instruction following the MIE write because pending is already set.  If pending is hard-wired (under the assumption that it will always fire immediately) then I think it would trap one instruction later.

Perhaps "a mode where the trigger is enabled" (up above) needs to be clarified with respect to this MIE behavior.  I'm assuming that icount is updated regardless of MIE and that only the trigger is suppressed.  But I think others could interpret it differently.

pdonahue-ventana:
Consider this scenario:
- mstatus.MIE=0 and mstatus.MPIE=1 and mstatus.MPP=3 and we're in M mode
- write icount action=0, m=1, count=1, pending=0
- mret (which sets MIE=1 and stays in M mode)
- execute an instruction

It's pretty clear that if this instruction is an "add" then we trap immediately after the instruction (since the act of executing the instruction decrements count which sets pending).  Now consider:
- the instruction is a CSR write which writes hit=0.  My assumption is that this shouldn't change anything (i.e. the CSR write happens before the trigger fires so the end result is hit=1).
- the instruction is a CSR write which writes pending=0.  My assumption is also that nothing changes and that the CSR write happens before the pending update.  But if it's like minstret then the CSR write actually causes the automatic/implicit CSR update to not happen.
- the instruction is a CSR write which writes count=2.  My assumption is that pending should not get set (i.e. the count value is read only after the CSR write happens).

I don't really think that the behavior matters in these situations so I suggest that we say that the value of pending is UNSPECIFIED if the instruction being stepped writes to icount (no matter what field).  

Another scenario:
- mstatus.MIE=1 and we're in M mode
- write icount action=0, m=1, pending=1

My assumption is that we must trap immediately after the CSR write instruction because:
> When pending is set, the trigger fires just before any further instructions are executed in a mode where the trigger is enabled.

My conclusion is that action=0, m=1 is kind of a recipe for disaster.

core_debug.tex
```diff
@@ -130,6 +130,23 @@ \subsection{Icount Trigger}
...
+It also allows single step on tiny system which only supports M-Mode, but the
+code becomes tricky to get right. To single step, the debug stub would execute
+something like:
+\begin{verbatim}
+    li    t0, \FcsrIcountCount=4, \FcsrIcountAction=0, \FcsrIcountM=1
+    csrw  tdata1, t0    /* Write the trigger. */
+    lw    t0, 8(sp)     /* Restore t0, count decrements to 3 */
+    lw    sp, 0(sp)     /* Restore sp, count decrements to 2 */
+    ret                 /* Return to program being debugged. count decrements to 1 */
+\end{verbatim}
```
pdonahue-ventana, line 148: I think it should be mret instead of ret.  You got into the single-step handler via a trap which put the return address in mepc instead of x1/ra.  And interrupts are disabled in the handler (which is a good thing because an interrupt coming between the csrw and the return would cause problems) and mret will reenable interrupts.  But otherwise this code seems right.

But I think it's best to not give code examples that are approximately correct and tricky to get right.  Should the spec say "for example, the debug stub can execute this" (giving one possibility for something that definitively works) or should it not provide code at all and just say something more vague about the M-mode code needing to set icount.count to 1 plus the number of instructions after the csr write but before the instruction being stepped?  It's useful to readers to be definitive but more maintainable and less error-prone to be vague.

This is a valuable addition to point out that some implementations may need to consider supporting count>1 (which is not what the spec previously said).

xml/hwbp_registers.xml
```diff
@@ -777,18 +777,30 @@
...
-        When
-        \FcsrIcountCount is decremented from 1 to 0 then \FcsrIcountPending
-        becomes set. When \FcsrIcountPending is set, the trigger fires just
+
+        When \FcsrIcountCount is 1, and an instruction is completed or trap
+        taken from a mode where the trigger is enabled, then \FcsrIcountPending
+        becomes set. In addition \FcsrIcountCount may become 0 as well.
+
+        The only exception to the above is when the instruction executed is a
+        write to the icount trigger. In that case \FcsrIcountPending may or may
```
pdonahue-ventana, line 789: I think that we were generally going with "might or might not" or simply "might not" in order to avoid the ambiguous "may not" (which is really not ambiguous in this particular context with "may or" but I think it's best to use consistent phrases).

xml/hwbp_registers.xml
```diff
@@ -777,18 +777,30 @@
...
+        not become set if \FcsrIcountCount was 1. Afterwards \FcsrIcountCount
+        contains the newly written value.
+
+        When \FcsrIcountCount is 0 it stays at 0.
```
pdonahue-ventana, line 793: until explicitly written by a CSR instruction.

xml/hwbp_registers.xml
```diff
@@ -777,18 +777,30 @@
...
+
+        When \FcsrIcountPending is set, the trigger fires just
         before any further instructions are executed in a mode where the
         trigger is enabled. As the trigger fires, \FcsrIcountPending is
+        cleared. In addition, if \FcsrIcountCount is 1 then \FcsrIcountM,
+        \FcsrIcountS, \FcsrIcountU, \FcsrIcountVs, and \FcsrIcountVu are all
```
pdonahue-ventana, line 799: This isn't currently required, though the count field says that this is allowed.  For instance, if icount.m=s=0 and icount.u=1 then there's no need to clear icount.u when trapping to M or S.  And there's no requirement to clear any of the mode filters on an action=1 trigger.  So this should change to be an allowed but not required behavior.

xml/hwbp_registers.xml
```diff
@@ -825,19 +842,17 @@
...
         </field>
 
         <field name="m" bits="9" access="WARL" reset="0">
             When set, enable this trigger in M-mode.
         </field>
         <field name="pending" bits="8" access="R/W" reset="0">
             This bit becomes set when \FcsrIcountCount is decremented from 1
-            to 0. It is cleared when the trigger fires.
+            to 0. It is cleared when the trigger fires, which will happen just
+            before the next instruction in one of the enabled modes is executed.
```
pdonahue-ventana, line 855: This is just a little easier for me to parse:
```suggestion
            before executing the next instruction in one of the enabled modes.
```

core_debug.tex
timsifive: I mostly wrote this code to force myself to work through the problem. How about I move it into the appendix?

My main conclusion is that this is probably doable, but getting the code right might be very tricky.

xml/hwbp_registers.xml
timsifive: "until explicitly written." Could happen through abstract write command.

xml/hwbp_registers.xml
timsifive: I think this is right. If icount.count=1 when pending is set, that means count was hardwired to 1. Otherwise it would be 0. But I'll clarify that this only applies when count is hardwired to 1.

pdonahue-ventana: > If icount.count=1 when pending is set, that means count was hardwired to 1. Otherwise it would be 0.

Or it means that software wrote pending=1, count=0.  Useful?  No.  Possible?  Yes.

xml/hwbp_registers.xml
pdonahue-ventana: Good point.  Between the two of us (and any onlookers who are also welcome to contribute), we'll get it right.

timsifive:
I think I've addressed all the feedback.

xml/hwbp_registers.xml
timsifive: This section is already wrong. You can't tie all the mode bits together and have a single step useful to native debuggers. I'm redoing it.

timsifive: In fact, I'm mostly getting rid of it. The single-bit implementation can work, but only if you tie all mode bits together and have max count of 1. If all mode bits are enabled together, there is no case where the existence of a pending bit can make a different. But this is only useful for an external debugger, and we already say that for external debuggers dcsr.step is the preferred solution.

For native debugging, the least hardware I can think of is 2 bits, because there are effectively 3 states: disabled, pending, and enabled. This assumes the mode bits are tied together in such a way that there is only one sensible option. I'm not sure if that's even worth mentioning anywhere. It would be a long explanation with little benefit.

xml/hwbp_registers.xml
```diff
@@ -795,21 +795,34 @@
     <register name="Instruction Count" short="icount" address="0x7a1">
         This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 3.
 
-        If \FcsrIcountCount is not 0, then every instruction completed or
+        When \FcsrIcountCount is greater than 1, every instruction completed or
         trap taken from a mode where the trigger is enabled decrements
```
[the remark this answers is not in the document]
pdonahue-ventana: > I think it just means the privilege mode. That seems simpler, and I'm not sure why you'd want something more complex.

I agree.  Unless you think this is obvious, can you add a parenthetical phrase or a separate sentence or some sort of clarification that interrupt enables don't matter for purposes of this sentence?  I know that this was preexisting text and so it's maybe a bit more than you intended to do, but it helps to get us incrementally closer to perfection (which is more of an asymptote than anything actually achievable).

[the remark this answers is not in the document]
timsifive: I've added a line.

timsifive:
@pdonahue-ventana Can you approve this, so we can refer to just a single icount implementation in the email thread?

xml/hwbp_registers.xml
```diff
@@ -846,19 +860,17 @@
             and writing it has no effect.
         </field>
         <field name="count" bits="23:10" access="WARL" reset="1">
-            When count is decremented to 0, the trigger fires. Instead of
-            changing \FcsrIcountCount from 1 to 0, it is also acceptable for hardware to
-            clear \FcsrIcountM, \FcsrIcountS, \FcsrIcountU, \FcsrIcountVs, and
-            \FcsrIcountVu. This allows \FcsrIcountCount to be hard-wired
-            to 1 if this register just exists for single step.
+            The trigger will generally fire after \FcsrIcountCount instructions
+            in enabled modes have been executed.
```
pdonahue-ventana, line 864: How about adding "See above for the precise behavior."  Otherwise the "generally" sounds pretty imprecise and the reader might not understand that it's all specified properly.

pdonahue-ventana (commented):
Sorry, just one more little thing.

xml/hwbp_registers.xml
timsifive: Good idea. Done.

pdonahue-ventana (approved):
Make it so.

PR #668, opened 2021-08-26
Don't call read/write CSRs read-only.

### LINKED ISSUE #645: tinfo writability
The privileged spec says:
> Attempts to access a CSR without appropriate privilege level or to write a read-only register also raise illegal instruction exceptions.

and:
> By convention, the upper 4 bits of the CSR address (csr\[11:8\]) are used to encode the read and write accessibility of the CSRs according to privilege level as shown in Table 2.1. The top two bits (csr\[11:10\]) indicate whether the register is read/write (00, 01, or 10) or read-only (11).

That's just a convention but, for instance, Spike only decodes the top two bits to determine whether to trap on writes.  Every CSR seems to follow that convention except for tinfo.

The tinfo register says "This entire register is read-only."  Its encoding is 0x7a4 which does not have 11 in the top bits.

I suggest simply declaring that this is a read/write register with no writable bits.

I was going to just propose this via a PR but it's a little more complicated than just changing that read-only sentence.  That sentence is added by registers.py and I'm not familiar with that whole flow (and I can figure it out but I'd prefer to have this done by someone who already knows it).  Of course, that sentence still has value in the DM and DTM contexts but there needs to be different treatment for either tinfo or the whole hwbp_registers.xml (since tinfo is the only hwbp register that has no writable fields).

timsifive:
Read-only CSRs cause an exception when you write to them. What we mean
is that there are no writable bits, so writes are simply ignored.

Fixes #645

### DIFF
--- a/registers.py
+++ b/registers.py
@@ -398,7 +398,14 @@ def print_latex_custom( registers ):
         print()
 
         if r.fields:
-            if all(f.access in ('R', '0') for f in r.fields):
+            if registers.prefix == "CSR_":
+                if int(r.address, 0) >= 0xc00:
+                    print("This CSR is read-only.")
+                elif all(f.access in ('R', '0') for f in r.fields):
+                    print("Writing this read/write CSR has no effect.")
+                else:
+                    print("This CSR is read/write.")
+            elif all(f.access in ('R', '0') for f in r.fields):
                 print("This entire register is read-only.")
 
             print("\\begin{center}")


pdonahue-ventana (approved):
Looks good.  I checked out the branch, ran make, and eyeballed some CSRs (including tinfo) and some DM registers to make sure that everything looked OK.

PR #669, opened 2021-09-07
Label Ztrigger and Zdebug chapters.

timsifive:
This is the latest thought on how to divide the spec in order to get it
ratified.


pdonahue-ventana:
My understanding of the Zdebug/Ztrigger was that there should conceptually be these two named extensions.  But in practice I don't think that the names Zdebug and Ztrigger comply with the unpriv spec chapter 27:
> The first letter following the “Z” conventionally indicates the most closely related alphabetical extension category, IMAFDQLCBKJTPV.

I had asked previously in tech-debug about extension names and I don't think that anything was really resolved except that it didn't matter.  Now that it matters, I think that we need to do something that complies with the naming convention.  Like:
- Smdebug and Smtrigger (which doesn't comply with the current chapter 27 but does comply with [this PR](https://github.com/riscv/riscv-isa-manual/pull/688) which reflects names that people are already using)
- Change the naming convention (possibly in that same PR) to reserve Sd for debug and then have Sdbase for the base debug stuff and Sdtrigger for the TM
- Something else?

timsifive:
I like Sdbase and Sdtrigger, but whatever people will accept is fine by me. I think you're more in on this kind of stuff than I am.

pdonahue-ventana:
The architecture review committee discussed this.  I'm told that there are broader naming convention issues that won't be resolved until 2022.  In the meantime, the suggestion from the committee was to have tentative names and expect them to change.  The proposed tentative name for the TM was Sdtrig1p0 (which by my reading is v1.0 of the extension called Sdtrig).  There was no proposal for the other extension but we can presumably come up with some similar placeholder within this "Sd" namespace that's presumably reserved for debug.  I had been thinking Sdbase (the base extension) but you can support Sdtrig without supporting external debug (if only action=0 is supported) so maybe it's not really the base extension at all.  Debug mode exists iff external debug is supported so now I'm leaning toward Sdext.  It'll probably change anyway so what do you think of Sdtrig and Sdext? (a.k.a. Sdtrig1p0 and Sdext1p0 in this version of the spec, with Sdtrig0p13 and Sdext0p13 for the 0.13 version which they also seem to be concerned with but which is beyond the scope of this PR)

pdonahue-ventana (changes requested):
@timsifive: I think that this should `s/Ztrigger/Sdtrig/g` and `s/Zdebug/Sdext/g` per my previous comment.

pdonahue-ventana (approved):
Works for me.  There's a non-zero probability that this will change but hopefully these names are good enough that it's closer to 0% than 100%.

PR #673, opened 2021-09-14
SBA write memory example: configure size

### LINKED ISSUE #671: B8.1 (Writing Memory) Using System Bus Acces
The first example shows writing a word to memory, Is this always a word transfer independent of the sbaccess value in sbcs or needs sbcs to be set according to the wanted transfer width?

timsifive:
Fixes #671


pdonahue-ventana (approved):
The two examples are kind of the same except that there's only one write on the first one, which makes autoincrement optional.  But it's fine to have the simple example first and then build on that for the second example.

PR #681, opened 2021-10-15
Fix obvious error in dcsr.cause=2 description.

pdonahue-ventana:
action=1 enters debug mode, not action=0.


timsifive (approved):
Good catch.

PR #683, opened 2021-10-19
fixed wrong ref in Terminology section and rebuilt PDF

allexoll:
There was a ?? missing ref in the PDF prior to this in section 1.1 Terminology


timsifive:
Thanks.

Can you force push a change that doesn't rebuild the PDF? If we rebuild the PDF all the time then there will be a ton of conflicts, and generally git isn't great at handling lots of binary changes. I manually rebuild the PDF once a month, and push that.

PR #688, opened 2021-11-02
Hide registers in Latex but generate them to C/Scala code

JanMatCodasip:
This is useful for e.g. DM_PROGBUF## and DM_DATA## registers which are omitted
in the documentation but we still wish to use the corresponding constants
in program code (e.g. in OpenOCD).

### DIFF
--- a/registers.py
+++ b/registers.py
@@ -34,13 +34,14 @@ def sympy_compare_lowBit( a, b ):
     return 0
 
 class Register( object ):
-    def __init__( self, name, short, description, address, sdesc, define ):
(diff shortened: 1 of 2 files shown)


JanMatCodasip:
This change is also needed for OpenOCD to compile, without the need to replace all occurrences of `DM_DATA1`+ or `DM_PROGBUF1`+.

JanMatCodasip:
I am taking this change back. I was looking at an older revision by mistake, where the `DM_PROGBUF##` and `DM_DATA##` registers were commented out in the XML file. This is no longer true.

PR #692, opened 2021-12-10
Resume ack is set for any resume cause.

timsifive:
With input from Oded Yishay.

This is not technically backwards compatible, but if I recall the meeting discussion correctly SiFive already implements it this way, and everybody agreed it makes more sense.


pdonahue-ventana (approved):
Looks good.

Issue #691, opened 2021-12-08
Clarify priority of dcsr.step with respect to other debug exceptions


JamesKenneyImperas:
I think that the description of debug exception priorities should be tightened up to clarify how debug step exceptions (`dcsr.step=1`) fit when other debug exceptions are also potentially pending. I'm also wondering if the currently specified priority for `ebreak` is anomalous.

Section 4.8.1 lists these priorities, in ascending order:
priority 0: debug step
priority 1: `haltreq`
priority 2: `resethaltreq`
priority 3: `ebreak`
priority 4: trigger module action
priority 5: halt group

This implies that debug step is always lowest priority, which would mean that if `haltreq` is asserted at the point at which a debug step exception would otherwise be taken, then `haltreq` should be reported (cause 3).

However, section 4.4.1 (Step Bit in Dcsr) says this:
_If the instruction that is executed causes the PC to change to an address where an instruction fetch causes an exception, that exception does not occur until the next time the hart is resumed. **Similarly, a trigger at the new address does not fire until the hart actually attempts to execute that instruction**._

This implies that a trigger module debug exception (priority 4) on the _next_ instruction does not take priority over a debug step exception (priority 0) on completion of the _current_ instruction, which I think makes sense. This would in turn imply that if `haltreq` is asserted at the point at which a debug step exception would otherwise be taken, then `haltreq` should _not_ be reported, because if a trigger module debug exception was also asserted that would have higher priority, which contradicts the statement above.

I think it might help to redefine exception priorities to include _current_ instruction and _next_ instruction in the priority ordering, by adding the explicit rule that all debug exceptions occurring _after the current instruction_ have priority over anything that will occur _before the next instruction_. This could give this priority order:

**after current instruction**
1: trigger module action with timing After (priority 4)
2: debug step (priority 0)
**before next instruction**
3: halt group (priority 5)
4: trigger module action with timing Before (priority 4)
5: `resethaltreq` (priority 2)
6: `haltreq` (priority 1)

I have omitted `ebreak` from this list, because if that instruction causes entry into Debug mode it is going to have the effect of disabling any other attempt to enter Debug mode. This means that I think it should really sit above step 1 in the above list, and its indicated priority of 3 is really meaningless.

I think the only significant case where this really makes a difference is to clarify what happens when `haltreq` is asserted _during execution_ of an instruction when `dcsr.step=1`, which currently isn't clear.

Thanks.

pdonahue-ventana:
[...]

PR #693, opened 2021-12-10
Change dcsr.cause priorities

pdonahue-ventana:
Here's a draft proposal to address #691 and https://lists.riscv.org/g/tech-debug/topic/86584183 which has been on my to-do list for months.  I welcome any suggestions for additional wordsmithing and/or substantive changes.

I took advantage of the fact that table 5.2 doesn't have any triggers that are lower priority than "Environment break" that could possibly fire on an ebreak instruction (since load/store triggers can't occur on an ebreak).  That led to my observation that the few lower priority things like timing=after triggers are actually treated as high priority exceptions (or halts in this case) on the next instruction.

@JamesKenneyImperas


JamesKenneyImperas:
Hello Paul,

I think this is certainly an improvement. Moving the priority of trigger below haltreq removes the non-determinism implied by the pipeline effect that I mentioned in issue #691, so that's a good thing, and distinguishing Before and After triggers is also good. I still think it could be an improvement to be even more explicit about this and include the rule that all triggers on a preceding instruction are higher priority than any on the next instruction, like this:

pdonahue-ventana:
@JamesKenneyImperas: You suggest:
> after current instruction
1: ebreak
2: trigger module action with timing After
3: debug step
before next instruction
4: resethaltreq
5: halt group
6: haltreq
7: trigger module action with timing Before

That seems equivalent to:
> before current instruction
1: resethaltreq
2: halt group
3: haltreq
4: trigger module action with timing Before
> after current instruction
5: ebreak
6: trigger module action with timing After
7: debug step

And that seems equivalent to what I said except for item 6 (trigger with timing=after).  I proposed timing=after on instruction N to be at the same priority as timing=before on instruction N+1.  So the question is whether triggers with timing=after should be at this priority with respect to debug step, resethaltreq, halt group, haltreq (the four things in between this instruction's item 6 and next instruction's item 4).  I don't think that the priority relative to resethaltreq matters here since that can't really happen concurrently with a timing=after trigger.  The assertion of halt group and haltreq are race conditions so my thought is that they don't really matter here (while I could come up with a weird case in which you could reliably observe the priority of timing=before triggers vs. haltreq, I couldn't do the same with timing=after triggers).  That leaves debug step (where the priority is clearly going to be visible).

In my proposal, the debug step priority is below the priorities of anything in table 5.2 (which is equivalent to either being the lowest priority on this instruction or the highest priority on the next).  In your proposal, debug step's priority is between the "mcontrol/mcontrol6 after" priority and the "mcontrol/mcontrol6 execute address before" priority.  That seems unnatural to me but I'm curious to hear your thoughts.

JamesKenneyImperas:
Hello Paul,

I was imagining the case when we are single-stepping through a sequence of instructions _ABC_ where there is a trigger with timing _Before_ on instruction _C_. In your proposal, assuming a return from Debug mode to continue execution each time, there would be a step event after instruction _A_ then a _trigger_ event after instruction _B_, then a step event after instruction _C_ (I think). So the trigger event effectively suppresses the step event.

Now imagine the same scenario except that an interrupt occurs just before execution of instruction _C_ so that it is not executed. In this case, I think you would have a step event after instruction _A_, then a _step_ event after _B_, and then possibly a step event after the first instruction of the interrupt handler (but no trigger on _C_ before this, because of the restrictions in section 4.4.1). The effect is that an interrupt _after_ the stepped instruction affects the reported cause after execution of that instruction and before any attempt to execute the next instruction.

Lowering the priority of the _Before_ trigger event below the step event would in the first case produce a step event after instruction _A_, a step event after instruction _B_, a trigger event before instruction _C_ and a step event after instruction _C_. So no step events are lost, but there is an extra debug event reported. In the second case, the events would be a step event after instruction _A_, a step event after instruction _B_, and then possibly a step event after the first instruction of the interrupt handler, as before.

A disclaimer: I am not in any way an expert on debuggers or debug hardware, so feel free to ignore anything said here. However, I do know about writing processor models, any one key bit of those is the priority ordering of debug and interrupt events between instructions. The code for this in the Imperas model is a bit of a dog's breakfast at the moment because of the oddities in the specified priorities that we've both identified, which require that _Before_ triggers are speculatively probed. Your changes would significantly clean that code up, and lowering the priority of the _Before_ trigger event below the step event would further simplify it.

Thanks.

JamesKenneyImperas:
Hello Paul,

Another point - having _Before_ trigger higher priority than debug step also could lead to non-determinism if presentation of the trigger event is delayed by pipeline bubbles as the designers I discussed this with suggested - you might get either the step event _or_ the trigger event reported.

I think what I'm trying to do is to push reporting of _Before_ triggers back to the point when we know that the instruction they refer to is _definitely_ just about to be executed. I think that should be deterministic.

Thanks.

pdonahue-ventana:
> I was imagining the case when we are single-stepping through a sequence of instructions ABC where there is a trigger with timing Before on instruction C. In your proposal, assuming a return from Debug mode to continue execution each time, there would be a step event after instruction A then a trigger event after instruction B, then a step event after instruction C (I think). So the trigger event effectively suppresses the step event.

I disagree.  I'd expect my proposal to produce:
- exit debug mode with dpc=A
- execute A
- enter debug mode due to the step with dpc=B
- exit debug mode with dpc=B
- execute B
- enter debug mode due to the step with dpc=C *
- exit debug mode with dpc=C
- timing=before trigger fires which prevents C from retiring (the whole point of timing=before)
- enter debug mode due to the trigger with dpc=C (since C has not executed)

You were thinking that the * item would indicate a trigger instead of a step.  But you have to complete execution of B before you execute C so you don't even know about the timing=before trigger on C yet.  The lowest priority cause on B (the step) is higher than the highest priority cause on C (the trigger).

> Now imagine the same scenario except that an interrupt occurs just before execution of instruction C so that it is not executed. In this case, I think you would have a step event after instruction A, then a step event after B, and then possibly a step event after the first instruction of the interrupt handler (but no trigger on C before this, because of the restrictions in section 4.4.1). The effect is that an interrupt after the stepped instruction affects the reported cause after execution of that instruction and before any attempt to execute the next instruction.

This is presumably assuming dcsr.stepie=1 which is generally not what debuggers want but it's legal and it can be beneficial to explore these weird cases.  In this case, I'd expect:
- exit debug mode with dpc=A
- execute A
- enter debug mode due to the step with dpc=B
- exit debug mode with dpc=B
- execute B
- enter debug mode due to the step with dpc=C (everything is the same up to this point)
- an interrupt becomes pending but is not taken because we're in debug mode (I put this at this point in time to avoid any race conditions between the interrupt coming and executing C)
- exit debug mode with dpc=C and stepie=1
- the interrupt traps, with mepc/sepc/vsepc pointing to C
- enter debug mode due to the step (of the interrupt trap) with dpc pointing to mtvec/stvec/vstvec (which is the first unexecuted instruction)

That seems reasonable to me.  You can continue to step and you'll see the whole interrupt handler, the mret which goes to C (because mepc=C), C itself (at which time any timing=before triggers on C may fire), etc.

> A disclaimer: I am not in any way an expert on debuggers or debug hardware, so feel free to ignore anything said here.

You can also ignore anything I say here.  Most people do. 😄  You're pointing out some good points and I don't purport to have all the answers.  I think that we can work things out in a way that hopefully makes it easier to write debuggers, to write RTL, and to write software models.

> I think what I'm trying to do is to push reporting of Before triggers back to the point when we know that the instruction they refer to is definitely just about to be executed. I think that should be deterministic.

It should already be true that timing=before triggers are reported only if the instruction is about to be executed.  There are certainly implementations that do speculative execution but they must appear to execute things in-order (at least with respect to this hart - ignoring RVWMO and effects on other harts).  For instance, in this code:
```
A:   ld x1,(x2)
B:   j D
C:   ld x3,(x4)
D:   ld x5,(x6)
```
you can't take a timing=before trigger on C.  Only if C is definitely going to execute (which it isn't in this case) can you take that trigger on C.  Just like (ignoring debug altogether) you can't take an instruction page fault on B if you took a (relatively low-priority) load access fault on A.

pdonahue-ventana:
Maybe the real problem is that we have two different priority lists.  There's the dcsr description and there's table 5.2.  The table is the priority for just triggers.  The dcsr description is the priority for more things including triggers, but it doesn't distinguish all the flavors of triggers which have different priorities amongst themselves.

I was going to propose unifying everything into table 5.2 (add resethaltreq, halt group, haltreq at the top and add dcsr.step at the bottom).  This would make it clear that "mcontrol/mcontrol6 store data before" with action=1 is lower priority than "mcontrol/mcontrol6 execute data before" with action=0.  But having one table would be mixing external debug into the trigger module (which can be implemented independently, e.g. with action hardwired to 0).

Maybe the dcsr section should just say "The priorities are as in table 5.2 with the following modifications: resethaltreq is at the very top, followed by halt group, followed by haltreq, followed by the table, followed by dcsr.step."  Some wordsmithing is needed but I hope you get the idea.  I'm assuming that it's clear that "environment break" in the table either traps or halts depending on dcsr.ebreakm/ebreaks/etc.

JamesKenneyImperas:
Hello Paul,

It looks like we are actually saying pretty much the same thing here, which is that all events caused by instructions _that have been executed_ are higher priority than anything reported on an instruction _that has not been started_. That makes the apparently-lower-priority step reason have priority over the apparently-higher-priority address trigger on the next instruction. This is excellent, and resolves my dog's breakfast scenario (when you also pull the priority of `haltreq` above a timing=before trigger, as you already proposed).

In simple terms, I think we are proposing that "between" any instructions _A_ and _B_, debug exceptions should be considered for dispatch in the following priority order:
1. Anything resulting from direct execution of instruction _A_ (`ebreak`, then timing=after trigger, then `step`);
2. Anything asynchronous (`resethaltreq`, then halt group, then `haltreq`; `resethaltreq` can be considered to fit here because when it is active there is no instruction _A_, so case 1 can't have done anything, but if you prefer you can hoist it above case 1).
3. Any timing=before trigger on instruction _B_, if that is definitely going to be executed (so not if an asynchronous debug event or other interrupt gets in first, for example).

This looks like it eliminates all ambiguities and any non-deterministic behaviors for pipeline effects.

Thanks.

timsifive:
> Maybe the dcsr section should just say "The priorities are as in table 5.2 with the following modifications: resethaltreq is at the very top, followed by halt group, followed by haltreq, followed by the table, followed by dcsr.step." Some wordsmithing is needed but I hope you get the idea. I'm assuming that it's clear that "environment break" in the table either traps or halts depending on dcsr.ebreakm/ebreaks/etc.

I think you can achieve this nicely by updating the table you added, and where it says "triggers" expand to say that this means including Table 5.2 into this one, in its existing order.

timsifive (commented):
It does sound like everybody is wants the same thing, but the language isn't quite clear to James.

I think this change is a nice improvement, tightening up a little bit of sloppiness in the spec. It feels like we're really close to something that is clear for everybody.

timsifive (approved):
This looks good to me.

pdonahue-ventana:
My recollection is that there are still some clarifications necessary.  This is on my todo list (but clearly has not been near the top).

pdonahue-ventana:
> I think you can achieve this nicely by updating the table you added, and where it says "triggers" expand to say that this means including Table 5.2 into this one, in its existing order.

After lots of paging in from long-term storage, I finally did the above.  I didn't use those exact words but see what you think.

timsifive (changes requested):
This changes the priorities as follows:

```
| old             new
| ------------- | -------
| group         | resethaltreq
| trigger       | group
| ebreak        | haltreq
| resethaltreq  | trigger
| haltreq       | ebreak
| step          | step
```

The new priorities seem fine, but it is a lot of change. Is this backwards compatible? Certainly the name of the PR should be to change dcsr.cause priorities.

pdonahue-ventana:
> The new priorities seem fine, but it is a lot of change. Is this backwards compatible?

What changed is that resethaltreq and haltreq moved up.

For resethaltreq, I don't think that there's any incompatibility since dcsr and triggers are reset to 0.  Therefore, the first instruction out of reset cannot get trigger or ebreak.  (If the first instruction is an ebreak then it will trap rather than halt.)  There could be a halt group when you come out of reset.  However, I don't think that resethaltreq can be lower priority than group without also elevating haltreq above resethaltreq, since haltreq and group must be adjacent to each other (since group can map to haltreq so they presumably have to be evaluated at a comparable time).

The only noticeable change is moving haltreq above ebreak and trigger to be near asynchronous interrupts and adjacent to group.

> Certainly the name of the PR should be to change dcsr.cause priorities.

I just did that.

timsifive:
If it fair to say this change reorganized the priorities with an eye on simplifying implementation?

I agree the new priorities are better. (It makes no sense to require you to check if the first instruction out of reset is ebreak so you can set priority to halt instead of resethaltreq if that was requested.) What do we do about existing implementations that implement the old order? Should we say this order is recommended but not required for existing implementations?

pdonahue-ventana:
I think that you can say that it simplifies implementation.  I doubt that anybody actually implemented some of the old things like halt groups not being treated similar to normal haltreq.  We don't have an exhaustive set of tests that simultaneously cause both events and check the priority so I guess we don't know.

As for making the order a recommendation, how about tightening it up a bit more like this?
- Require the relative priority of group > trigger > ebreak > step.  Since only resethaltreq and haltreq changed, this would allow the old way and the new way but not some other way.
- Require resethaltreq to be higher priority than haltreq.  This was true in both the old way and new way.
- Recommend that resethaltreq and haltreq have the new priorities relative to the other 4.

xml/core_registers.xml
```diff
@@ -2,9 +2,52 @@
...
     <register name="Debug Control and Status" short="dcsr" address="0x7b0">
+        Upon entry into Debug Mode, \FcsrDcsrV and \FcsrDcsrPrv are updated
+        with the privilege level the hart was previously in, and \FcsrDcsrCause
+        is updated with the reason for Debug Mode entry.  Other than these
+        fields and \FcsrDcsrNmip, the other fields of \RcsrDcsr are only
+        writable by the external debugger.
+
+        \begin{steps}{Table~\ref{tab:dcsrcausepriority} shows the priorities of
+        reasons for entering Debug Mode.  Implementations should implement
+        priorities as shown in the table.  resethaltreq and haltreq are
```
timsifive, line 13:
```suggestion
        priorities as shown in the table. For compatibility with old versions of this spec, resethaltreq and haltreq are
```

timsifive (changes requested):

timsifive (approved):
Looks good!

PR #694, opened 2021-12-11
Specify accessibility of mscontext

### LINKED ISSUE #690: Default accessibility of mscontext
Can you please clarify the accessibility restrictions for _mscontext_ CSR? Given it is encoded like a machine CSR, will an access from supervisor mode result in an illegal trap?

pdonahue-ventana:
Fixes #690.

This doesn't conform with what @timsifive said but I think it's correct.  I also tried to clearly discourage use of this and explain why.

@valtrix
@JamesKenneyImperas


timsifive (approved):
This looks good, and is better than what I commented on #690.

PR #695, opened 2021-12-16
Define mtval behavior on action=0 triggers

### LINKED ISSUE #636: What value should be written to xtval when triggers fire?
Could you please clarify the values expected to be written to `xtval` registers when triggers fire and cause breakpoint exceptions? There is almost no information about this in either the Privileged or Debug specifications.

In the Privileged Specification, there is only one reference for each `xtval`, of this form:

> When a **hardware breakpoint is triggered**, or an address-misaligned, access-fault, or page-fault exception occurs on an instruction fetch, load, or store, **mtval is written with the faulting virtual address**.

The term _faulting virtual address_ is inapplicable here, because there is no fault, so this needs clarification. Some triggers (e.g. icount) have no associated address - is zero written in this case, or something else (e.g. the current PC)? Chained triggers can have multiple addresses (e.g. write to address A while executing at address B) - which if any address is written in such cases?

Ideally, the required behavior should be clearly stated for all trigger possibilities. If this is deemed to be implementation dependent, then that should be clearly stated, with possible implementations listed.

Thanks.

pdonahue-ventana:
Note that you're always allowed to write 0 to mtval (per the priv spec) so this doesn't, for instance, prevent writing mtval=0 on any action=0 trigger.

Also made it clear that "all CSRs" that are updated on etrigger/itrigger is referring to the CSRs that are updated due to the exception or interrupt,separately from anything the trigger itself does.

Fixes #636.


Sdtrig.tex
```diff
@@ -131,7 +131,9 @@ \section{Priority}
 \end{table}
 
 When multiple triggers in the same priority fire at once, \FcsrMcontrolHit (if
-implemented) is set for all of them. If one of these triggers has
+implemented) is set for all of them.  If more than one of these triggers has
+\FcsrMcontrolSixAction=0 then {\tt tval} is updated in accordance with one of
+them (which one is \unspecified).  If one of these triggers has
```
timsifive, line 136:
```suggestion
them, but which one is \unspecified.  If one of these triggers has
```

timsifive: I feel parens are for explanation etc, while this is an important part of the spec.

timsifive (changes requested):

Sdtrig.tex
pdonahue-ventana: Yes, that makes sense.

timsifive (approved):
Thanks.

Can you remove the pdf from this PR?

pdonahue-ventana:
Oops.  Careless use of "git commit -a" strikes again.  I removed the PDF.

PR #698, opened 2021-12-27
"privilege level" -> "privilege mode" per architecture review feedback

pdonahue-ventana:


timsifive (approved):
Thanks for doing this!

PR #699, opened 2021-12-27
Stoptime clarification per architecture review

pdonahue-ventana:
I was going to say that it should stop the time CSR plus any custom CSRs that the custom extension might want to stop.  But custom extensions can always define their behavior to depend on standard CSR bits like stoptime so I decided that this wouldn't add anything.

### DIFF
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -32,8 +32,9 @@ How Debug Mode is implemented is not specified here.
 \item No action is taken if a trigger matches.
 \item If \FcsrDcsrStopcount is 0 then counters continue. If it is 1 then
     counters are stopped.
-\item If \FcsrDcsrStoptime is 0 then timers continue. If it is 1 then
-    timers are stopped.
+\item If \FcsrDcsrStoptime is 0 then the time CSR continues to update. If
+    it is 1 then the time CSR will not update. It will resynchronize with
+    mtime after leaving Debug Mode.
 \item The {\tt wfi} instruction acts as a {\tt nop}.
 \item Almost all instructions that change the privilege level have \unspecified\ 
     behavior.  This includes {\tt ecall}, {\tt mret}, {\tt sret}, and {\tt uret}.
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -32,9 +32,9 @@ How Debug Mode is implemented is not specified here.
 \item No action is taken if a trigger matches.
 \item If \FcsrDcsrStopcount is 0 then counters continue. If it is 1 then
     counters are stopped.
-\item If \FcsrDcsrStoptime is 0 then the time CSR continues to update. If
-    it is 1 then the time CSR will not update. It will resynchronize with
-    mtime after leaving Debug Mode.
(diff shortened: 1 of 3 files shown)


timsifive (commented):
Sounds OK to me.

Should we allow implementations to stop mtime while all harts are in debug mode? That would help with backwards compatibility, and might be a nice feature on single-hart systems.

pdonahue-ventana:
OK.  I now allow but don't require mtime to stop when stoptime=1.  (Let me know if you think it should be required or if you think there should be some mention of single-hart systems.)

Sdext.tex
```diff
@@ -32,8 +32,9 @@ \section{Debug Mode} \label{debugmode}
 \item No action is taken if a trigger matches.
 \item If \FcsrDcsrStopcount is 0 then counters continue. If it is 1 then
     counters are stopped.
-\item If \FcsrDcsrStoptime is 0 then timers continue. If it is 1 then
-    timers are stopped.
+\item If \FcsrDcsrStoptime is 0 then the time CSR continues to update. If
+    it is 1 then the time CSR will not update. It will resynchronize with
+    mtime after leaving Debug Mode.
```
timsifive, line 37:
```suggestion
\item If \FcsrDcsrStoptime is 0 then \Rtime continues to update. If
    it is 1 then \Rtime will not update. It will resynchronize with
    \Rmtime after leaving Debug Mode.
```

riscv-debug-spec.tex
```diff
@@ -22,6 +22,7 @@
 \deffieldname{\FcsrSstatusSie}{SIE}
 \deffieldname{\Fmxl}{MXL}
 \defregname{\Rmisa}{misa}
+\defregname{\Rmtime}{mtime}
```
timsifive, line 25:
```suggestion
\defregname{\Rmtime}{mtime}
\defregname{\Rtime}{time}
```

xml/core_registers.xml
```diff
@@ -78,9 +78,11 @@
             An implementation may hardwire this bit to 0 or 1.
         </field>
         <field name="stoptime" bits="9" access="WARL" reset="Preset">
-            0: Increment timers as usual.
+            0: Increment the time CSR as usual.
```
timsifive, line 81:
```suggestion
            0: Increment \Rtime as usual.
```

xml/core_registers.xml
```diff
@@ -78,9 +78,11 @@
...
 
-            1: Don't increment any hart-local timers while in Debug Mode.
+            1: Don't increment the time CSR while in Debug Mode.  If all harts
+            have \FcsrDcsrStoptime=1 and are in Debug Mode then the \Rmtime
+            register is also allowed to stop incrementing.
```
timsifive, line 85:
```suggestion
            1: Don't increment \Rtime while in Debug Mode.  If all harts
            have \FcsrDcsrStoptime=1 and are in Debug Mode then \Rmtime
            is also allowed to stop incrementing.
```

timsifive (approved):
Looks good. Just making things a bit less verbose. If people are confused what time/mtime refer to we should mention them somewhere and expand the macro to link there.

PR #701, opened 2022-01-07
Clarify that tdata1 is WARL

timsifive:


pdonahue-ventana (approved):
OK.  I noticed another unrelated problem when reading the text immediately below the steps but I'll make a separate PR for that.

Sdtrig.tex
```diff
@@ -250,7 +250,8 @@ \section{Trigger Registers}
 
 \begin{steps}{As a result, a debugger can write any supported trigger as
 follows:}
-\item Write 0 to \RcsrTdataOne.
+\item Write 0 to \RcsrTdataOne. (This will result in \RcsrTdataOne containing a
+    non-zero value, since the register is WARL.)
```
pdonahue-ventana, line 254: Actually, should this be `\warl`?

timsifive: Yes. That macro is defined so we should use it, even though it's not used anywhere else in the document yet.

PR #702, opened 2022-01-07
Clarifications/bug fixes: mscontext and tselect

pdonahue-ventana:
mscontext is accessible in S mode.
tselect is WARL.  This was in the XML but not in the final PDF.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -22,8 +22,10 @@
         This register is \warl.
         Writes of values greater than or equal to the number of supported
         triggers may result in a different value in this register than what was
-        written. To verify that what they wrote is a valid index, debuggers can
-        read back the value and check that \RcsrTselect holds what they wrote.
+        written or may point to a trigger where \FcsrTdataOneType=0.
+        To verify that what they wrote is a valid index, debuggers can
+        read back the value and check that \RcsrTselect holds what they wrote
+        and read \RcsrTdataOne to see that \FcsrTdataOneType is non-zero.
 
         Since triggers can be used both by Debug Mode and M-mode, the external
         debugger must restore this register if it modifies it.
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -19,6 +19,7 @@
         trigger registers. It is optional if no triggers are implemented.  The
         set of accessible triggers must start at 0, and be contiguous.
 
+        This register is \warl.
```
timsifive, line 22: I don't think that's true. It even says a few lines down:
> Writes of values greater than or equal to the number of supported triggers may result in a different
> value in this register than what was written. To verify that what they wrote is a valid index,
> debuggers can read back the value and check that tselect holds what they wrote.

The rationale was that this way somebody could implement tselect on a system which has e.g. 3 triggers, and not have to add any complexity by making the register WARL since the debugger can already detect that another way. That sounds a bit silly to me now, but here we are.

Maybe it depends on what you call "legal". Is a a legal value for tselect a value that points at a trigger which correctly says it's not implemented?

pdonahue-ventana: Yes, I think that it's legal for tselect to point to a nonexistent trigger as long as tdata1.type=0 (as in the situation you mention).  It's also legal for tselect to not be able to point to a nonexistent trigger (like probably if the number of triggers is a power of two).  And, of course, a combination can be true where there are 3 triggers, tselect=3 is possible (type=0 on that one) and tselect>3 is not possible (not that the debugger would actually need to attempt writing tselect>3).  That's my interpretation of "Writes of values greater than or equal to the number of supported triggers **may** result in a different value in this register than what was written" along with the fact that type=0 is legal.

I think it's OK for debuggers to loop through triggers and find the first one that _either_ tselect isn't able to point to or that has type=0.

If you agree then this should probably be more explicit.  If you disagree then we obviously need to converge on an answer.

I didn't think that this particular line actually changed anything since the XML already has:
```
 <field name="index" bits="XLEN-1:0" access="WARL" reset="0" />
```

timsifive: I agree. So I'd suggest saying "This register is \warl." like you do, and then add something like "Note that a value that is legal for \RcsrTselect might still point to a trigger where \FcsrTdataOneType=0."

pdonahue-ventana: I tried to weave that proposal into the existing text.  Take a look.

timsifive (approved):

PR #707, opened 2022-03-10
Clarify tmexttrigger with action=0.

pdonahue-ventana:
The privileged spec says that medeleg only delegates synchronous exceptions so clarify that in the case of tmexttrigger medeleg is delegating an asynchronous exception.

Also clarify what happens to tmexttrigger events that happen while the trigger is prevented from firing.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -1026,7 +1026,15 @@
         sequentially).  Unsupported inputs are hardwired to be inactive.
 
         If the trigger fires with action=0 then zero is written to the
-        {\tt tval} CSR on the breakpoint trap.
+        {\tt tval} CSR on the breakpoint trap.  This trigger fires
+        asynchronously but it is subject to delegation by medeleg[3] like
+        the other triggers.
+
+        The external trigger input can signal when the trigger is prevented
+        from firing due to one of the mechanisms in section ~\ref{sec:nativetrigger}.
+        An implementation may either ignore the signal altogether when it cannot
+        fire (dropping the trigger event) or it may hold the action as pending
+        and fire the trigger once it is legal to do so.
         <field name="type" bits="XLEN-1:XLEN-4" access="R" reset="4" />
         <field name="dmode" bits="XLEN-5" access="WARL" reset="0" />
         <field name="hit" bits="XLEN-6" access="WARL" reset="0">


xml/hwbp_registers.xml
```diff
@@ -1026,7 +1026,15 @@
...
-        {\tt tval} CSR on the breakpoint trap.
+        {\tt tval} CSR on the breakpoint trap.  This trigger fires
+        asynchronously but it is subject to delegation by medeleg[3] like
+        the other triggers.
+
+        The external trigger input can signal when the trigger is prevented
+        from firing due to one of the mechanisms in section ~\ref{sec:nativetrigger}.
+        An implementation may either ignore the signal altogether when it cannot
+        fire (dropping the trigger event) or it may hold the action as pending
+        and fire the trigger once it is legal to do so.
```
timsifive, line 1037: I'd imagined these triggers as edge-sensitive. If they don't happen, then they don't happen. It would be confusing if minutes ago some external trigger fired, and then when it becomes enabled you immediately take it. Is there some background discussion as to why we'd want it to be pending?

pdonahue-ventana: I tried to say that you can have either implementation.  The implementation that drops it would obviously be simpler.  The implementation that keeps it pending would not lose events.  By making implementers aware of the two options, they can make whatever tradeoffs they want (particularly since the source of these events is up to the implementation so it's hard for us to prejudge the optimal trigger behavior after the occurrence of some event that we can't envision).

The spec states elsewhere that "it is better to implement triggers imprecisely than to not implement them at all."  That seems to also apply here.  It is better (at least in some situations) to have this trigger fire much after an event than to not have it fire at all (depending on why it's firing).

Of course, this only affects action=0.  If you want some event to enable/disable trace or enter debug mode then you're guaranteed not to lose any of those events and this sentence is N/A since they are always eligible to fire.

pdonahue-ventana: You might do the pending behavior if this asynchronous trigger uses existing asynchronous interrupt mechanisms.  Since all RISC-V interrupts are level-sensitive (at the point of mip, not necessarily in the interrupt controller), the interrupt state machine wouldn't otherwise be designed to work with pulses.  It might need to be held high for some number of cycles.  In that case, the interrupt logic would already be capturing the pulse and holding it pending for normal cases and the MIE=0 case would just happen naturally.

Of course, I recognize that many/most implementations will find it easier to just drop the event like you suggest.  That's why I explicitly said that's OK and even listed it first.

ernie-sifive: FWIW, SiFive's implementation of external triggers fed by HPM overflows works this way -- trigger logic will detect the single-cycle overflow event and then hold the break request until it is satisfied.  HPM counters could increment while the core is in WFI or otherwise not executing anything, so overflows may not be associated with instruction flow at all.  Seems like you would have to use the asynchronous interrupt mechanism to handle this type of trigger.

timsifive (approved):

PR #709, opened 2022-03-15
Move nmi from etrigger to itrigger.

timsifive:
Per mailing list discussion subject:"etrigger.nmi".

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -921,7 +921,11 @@
           This bit is hard-wired to 0 if the hart does not support
           virtualization mode.
         </field>
-        <field name="0" bits="10" access="R" reset="0" />
+        <field name="nmi" bits="10" access="WARL" reset="0">
+            When set, non-maskable interrupts cause this
+            trigger to fire, regardless of the values of \FcsrItriggerM,
+            \FcsrItriggerS, \FcsrItriggerU, \FcsrItriggerVs, and \FcsrItriggerVu.
+        </field>
         <field name="m" bits="9" access="WARL" reset="0">
             When set, enable this trigger for interrupts that are taken from M
             mode.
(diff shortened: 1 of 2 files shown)


xml/hwbp_registers.xml
```diff
@@ -923,8 +923,7 @@
         </field>
         <field name="nmi" bits="10" access="WARL" reset="0">
             When set, non-maskable interrupts cause this
-            trigger to fire, regardless of the values of \FcsrItriggerM,
-            \FcsrItriggerS, \FcsrItriggerU, \FcsrItriggerVs, and \FcsrItriggerVu.
+            trigger to fire if the trigger is enabled for the current mode.
```
pdonahue-ventana, line 926: And tdata3?  Or don't say "if the trigger is enabled for the current mode" and then it's implied that normal filtering applies.  (Or is your goal to highlight that there was a change from the draft version?)

Also, something I mentioned in my email was that there can be multiple flavors of NMI and that tdata2 filtering can be useful (an argument that @gfavor pointed out to me).  If tdata2 is applied then I think that this should be explicitly called out because it might not be obvious that NMI support means that tdata2[0] must be implemented.  Because mie[0] is architecturally hardwired to 0, supporting nonzero in that tdata2 bit would otherwise not be necessary and people may miss this additional requirement when supporting NMI.

Maybe something like this (which is similar to how etrigger talks about mcause):
> When set, non-maskable interrupts cause this trigger to fire [if the trigger is enabled in the current mode and tdata3].  The trigger may fire on up to XLEN of the Exception Code values in mcause (described in the Privileged Spec section on Non-Maskable Interrupts).  Those causes are configured by writing the corresponding bit in tdata2.  (E.g. to trap on NMI with mcause Exception Code of 0, the debugger sets bit 0 in tdata2.)

timsifive: I'm struggling with this. I don't want to simply remove the clause, because "When set, non-maskable interrupts cause this trigger to fire" can be read to mean the mode bits don't matter at all. If I mention tdata3, then it should really be mentioned for every trigger, which it currently is not. Does that need to be added? It feels obvious to me that if you implement textra, then it applies, and if you don't implement it then it's irrelevant.

Which leaves me with exactly the language that's in the current PR.

What exactly did you intend with the square brackets in your example?

pdonahue-ventana:
I'm OK with that.

I agree that mentioning tdata3/textra in more places would improve the spec.  It is somewhat in the background today so it's easy to overlook that it plays a role in a lot of things.

pdonahue-ventana (approved):

PR #711, opened 2022-03-21
Mention textra in the trigger types that it affects.

timsifive:


xml/hwbp_registers.xml
```diff
@@ -260,6 +260,10 @@
         \unspecified\ whether a load trigger with \FcsrMcontrolSelect=1 will
         match.  Whether store triggers with \FcsrMcontrolSelect=1 match on
         AMOs is \unspecified.
+
+        If \RcsrTextraThirtytwo or \RcsrTextraSixtyfour are implemented, this
```
pdonahue-ventana, line 264: How about "are implemented for this trigger" on all of these?

timsifive: Good idea. Done.

pdonahue-ventana (approved):

Issue #713, opened 2022-03-22
Reference to configuration string unclear


christian-herber-nxp:
Among others, Section 3.14.9 Configuration String Pointer 0 refers to
`The configuration string itself is described in the Privileged Spec`
However, I cannot locate this, so I assumed this moved in the meanwhile in the spec. Maybe this needs to be adjusted

pdonahue-ventana:
The RISC-V Configuration Structure Task Group was formed 2 years ago to come up with the thing that this register will point to.
- https://lists.riscv.org/g/tech-config
- https://github.com/riscv/configuration-structure/

christian-herber-nxp:
thanks for the info. But indeed, I think the reference in the debug spec should be updated. With respect to the privileged spec, mconfigptr should be referenced, which is technically speaking also not a config string but a pointer to a config data structure.
Adding the reference to the config structure spec also makes sense

pdonahue-ventana:
This was addressed in #714.

PR #714, opened 2022-03-23
constrptr points to the configuration structure, not string.

pdonahue-ventana:
Also describe where this structure is mentioned in the priv spec.


pdonahue-ventana:
Fixes #713 

@christian-herber-nxp

timsifive (approved):
Beautiful. I especially like how confstrptr is still a reasonable short name, and that we won't need to update our spec when the configuration structure stuff is ratified.

PR #716, opened 2022-03-24
If masking is supported then tdata2 must be able to hold all XLEN bits.

### LINKED ISSUE #710: Issues with mcontrol and mcontrol6
I've noticed some issues with the `mcontrol` documentation, also present in the `mcontrol6` documentation. The first has a simple fix, but the latter two should be discussed.

1. In the case of an implementation supporting a lower bitmask comparison (`mcontrol.match` == 4 or `mcontrol.match` == 12), `tdata2` must be able to hold the full `XLEN` bits, otherwise it's not necessarially possible store all desired values to be ANDed with the compare value in (`XLEN`-1):(`XLEN`/2) of `tdata2`.
2. The specification does not indicate that including both Execute Address and Load Address triggers within a single chain is permitted or prohibited. To the best of my knowledge, current hardware implementations based on RocketCore and BOOM will fail to handle this chaining. (RocketCore does not currently give any ISA-visible indication of its failure here, except that the breakpoint exception never happens.) It is trivial to add the capability, and I've done so in a RocketCore prototype. The specification should require implementations not supporting this type of chaining indicate lack of support, either by zeroing the chain bit, or by zeroing the relevant non-conforming execute/store/load bits.
3. If chain that includes both Execute Address and Load Address triggers, the specification is ambiguous about what value should be present in `tval` when the exception occurs. In most implementations, the Load Address would be known later, and therefore be "the address which caused the" entire chain of "triggers to fire", but it's not clear that this is desired, and it's certainly not required by the specification. Perhaps the specification should indicate that if both an Execute Address and a Load Address triggered the chain, that the Load Address will be present in `tval`, especially since the Execute Address will end up in `epc` anyway. Alternatively, either the first or last trigger in the chain could be specified as contributing its address to `tval`, or maybe the specification should indicate this is entirely up to the implementation.

(I realize that some of this discussion should probably be on the mailing list, but I'm waiting on my university to get a membership first.)

pdonahue-ventana:
Fixes #710

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -228,7 +228,8 @@
         address triggers.
 
         This trigger type may be limited to address comparisons (\FcsrMcontrolSelect is
-        always 0) only. If that is the case, then \RcsrTdataTwo must be able to
+        always 0) only. If that is the case and masking is not supported (match
+        values 4, 5, 12, 13), then \RcsrTdataTwo must be able to
         hold all valid virtual addresses but it need not be capable of holding
         other values.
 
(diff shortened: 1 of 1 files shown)


timsifive (approved):
Thank you.

Issue #718, opened 2022-04-01
Clarify semantics for writing supported triggers


JamesKenneyImperas:
I've not been monitoring this specification for a while, so apologies in advance if this question has been already answered in another thread - if so, please point me at that.

On January 10th this year, text about writing supported triggers (section 5.5) was updated like this:

> As a result, a debugger can write any supported trigger as follows:
> 1. Write 0 to tdata1. _(This will result in tdata1 containing a non-zero value, since the register is WARL.)_
> 2. Write desired values to tdata2 and tdata3.
> 3. Write desired value to tdata1.

I'm not sure I understand how this should work in practice. Does this mean that each trigger now has a hidden "valid" bit that is not observable by software, and that when the trigger is in an "invalid" state the values in `tdata2` and `tdata3` can be inconsistent with the apparent value of `tdata1`? This feels a bit odd if so: it means that an apparently-programmed trigger might not actually be valid if for some reason step 3 has not been performed and it happens to be the case that the non-zero WARL value seen in `tdata1` is the same one that should have been written in step 3. This means that any software using the trigger module for its own purposes can't actually know whether a trigger is enabled or not, because the observed CSR state of an enabled and disabled trigger can be identical.

I think it would be better to either see the value 0 in `tdata1` when the trigger is in the intermediate programmable state or to introduce some other explicit value to indicate a programmable but disabled trigger (assuming value 0 is needed to indicate an absent trigger).

Thanks.

timsifive:
There shouldn't be a need for an internal valid bit.

The intent is that writing 0 to tdata1 results in a value in tdata1 that
allows any value that might be supported by any of the trigger types to be
written to tdata2 and tdata3.

If one of the supported trigger types is mcontrol/mcontrol6, then likely
tdata2 can hold any value, so writing 0 to tdata1 can simply result in type
2/6 being set in the trigger.
But you're right that if a trigger only supports itrigger and etrigger,
then this mechanism might not work, depending on which
interrupts/exceptions are actually supported.

Does this need to be updated to say that you have write 0 to tdata[1,2,3]
and then write tdata[1,2,3] in order?

Tim

JamesKenneyImperas:
Hi Tim,

[...]

PR #721, opened 2022-04-21
Use trigger type 15 to indicate disabled.

timsifive:
Also clarify tdata2/tdata3 behavior when a trigger is disabled.
Add a paragraph about context restoring triggers.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -68,7 +68,9 @@
             15: This trigger is disabled. In this state, \RcsrTdataTwo and
             \RcsrTdataThree can be written with any value that is supported for
             any of the types this trigger implements. The remaining bits in this
-            register are ignored. Other values are reserved for future use.
+            register are ignored.
+            
+            Other values are reserved for future use.
 
         </field>
         <field name="dmode" bits="XLEN-5" access="WARL" reset="0">
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -33,6 +33,10 @@
     <register name="Trigger Data 1" short="tdata1" address="0x7a1">
         This register is optional if no triggers are implemented.
 
+        Writing 0 to this register must result in a trigger that is disabled.
+        If this trigger supports multiple types, then this should be done by
+        changing \FcsrTdataOneType to 8.
+
         <field name="type" bits="XLEN-1:XLEN-4" access="WARL" reset="Preset">
             0: There is no trigger at this \RcsrTselect.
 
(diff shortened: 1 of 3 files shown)


timsifive:
Hopefully clarifies #718.

pdonahue-ventana:
Did you figure out what type=15 is for?

timsifive:
I just dug through my email. I invented type 15 at the very beginning of
tdata1.type (June 15, 2016). I really don't remember why.

Maybe it's to represent that the trigger is currently owned by the external
debugger, when the target tries to read the register? But the dmode bit
already shows that.

Tim

xml/hwbp_registers.xml
```diff
@@ -33,6 +33,10 @@
     <register name="Trigger Data 1" short="tdata1" address="0x7a1">
         This register is optional if no triggers are implemented.
 
+        Writing 0 to this register must result in a trigger that is disabled.
+        If this trigger supports multiple types, then this should be done by
+        changing \FcsrTdataOneType to 8.
```
pdonahue-ventana, line 38: This should make clear that the change should be made by hardware as part of the WARL process, not by the debugger.

xml/hwbp_registers.xml
```diff
@@ -59,6 +63,11 @@
             7: The trigger is a trigger source external to the TM.  The
             remaining bits in this register act as described in \RcsrTmexttrigger.
 
+            8: This trigger was disabled. In this state, \RcsrTdataTwo and
```
pdonahue-ventana, line 66: I think that we should avoid allocating a new type if we don't have to.  Can 15 serve this purpose?  15 could mean the trigger exists and is either disabled or not currently available.

xml/hwbp_registers.xml
timsifive: Done.

xml/hwbp_registers.xml
timsifive: Done.

xml/hwbp_registers.xml
```diff
@@ -61,10 +65,11 @@
 
             12--14: These trigger types are available for non-standard use.
 
-            15: This trigger exists (so enumeration shouldn't terminate), but
-            is not currently available.
+            15: This trigger is disabled. In this state, \RcsrTdataTwo and
+            \RcsrTdataThree can be written with any value that is supported for
+            any of the types this trigger implements. The remaining bits in this
+            register are ignored. Other values are reserved for future use.
```
pdonahue-ventana, line 71: I think that the last sentence should remain a separate paragraph since it's not related to value 15.
```suggestion
            register are ignored.
            
            Other values are reserved for future use.
```

pdonahue-ventana (changes requested):
This looks good overall.  Just a formatting issue.

pdonahue-ventana (approved):

PR #722, opened 2022-04-26
Tighten up language around icount matching.

timsifive:
Specifically, now icount triggers on M-Mode only systems are more
useful, because the count might not decrement while in an interrupt
handler.

See 2022 email discussion subject:"trigger questions"


pdonahue-ventana:
I think that this looks like the right approach.  It's technically not backward-compatible with the earlier stable spec.  Should that be mentioned in section 1.2.1.5 (incompatible changes during 1.0 stable)?

pdonahue-ventana:
I think that this looks like the right approach.  It's technically not backward-compatible with the earlier stable spec.  Should that be mentioned in section 1.2.1.5 (incompatible changes during 1.0 stable)?

pdonahue-ventana (approved):
Cool.

PR #723, opened 2022-04-26
mte/mpte apply only to breakpoint traps

timsifive:
This makes it possible to use triggers to debug some of the exception
handling code as well.

See 2022 email subject:"trigger questions"

### DIFF
--- a/introduction.tex
+++ b/introduction.tex
@@ -122,6 +122,7 @@ order to implement 1.0:}
     \item When a selected trigger is disabled, \RcsrTdataTwo and \RcsrTdataThree
     can be written with any value supported by any of the types this trigger
     supports. \PR{721}
+    \item \RcsrTcontrol fields only apply to breakpoint traps, not any trap. \PR{723}
 \end{steps}
 
 \subsubsection{Minor Changes from 0.13 to 1.0}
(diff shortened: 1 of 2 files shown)


pdonahue-ventana:
This seems like it's not backwards compatible.  I think that a note should be added to 1.2.1.5.

timsifive:
Huh. I'd written that but not pushed it up yet. Here it is.

pdonahue-ventana (approved):
Sounds good.

timsifive:
The e-mail thread about this really starts at https://lists.riscv.org/g/tech-debug/topic/89848254#691.

This change solves some problems when stepping into exception handlers when native debugging the highest privilege mode. (This is really the ideal use case for external debug, but we try to accommodate native debugging here as well.) If every trap into M-Mode clears mte, then you can't use icount triggers to step into an exception handler. By limiting this clearing to breakpoint exceptions only, more code can be debugged.

timsifive:
AR feedback is that this change results in the mechanism not serving its purpose when non-breakpoint traps occur.

That sounds right to me. Consider:
1. icount is set so that it'll trigger on the first instruction of an exception handler.
2. Code takes an exception, saving mepc etc.
3. Now icount fires causing a breakpoint exception, clobbering mepc.

@pdonahue-ventana can you look this over? You have a gift for corner cases.

PR #724, opened 2022-04-30
Add arguments to C macros that need them.

timsifive:
Previously no macros took arguments, even those that depended on e.g.
XLEN.

Old:
 #define CSR_TDATA1_TYPE_OFFSET              (XLEN-4)
 #define CSR_TDATA1_TYPE_LENGTH              4
 #define CSR_TDATA1_TYPE                     (0xfULL << CSR_TDATA1_TYPE_OFFSET)

New:
 #define CSR_TDATA1_TYPE_OFFSET(XLEN)        (XLEN + -4)
 #define CSR_TDATA1_TYPE_LENGTH              4
 #define CSR_TDATA1_TYPE(XLEN)               (0xf * (1ULL<<(XLEN + -4)))

The new syntax can be a little awkward, but it's correct and works in
OpenOCD at least. The most awkward new version is this one:
 #define DTM_DMI_ADDRESS_OFFSET              0x22
 #define DTM_DMI_ADDRESS_LENGTH(abits)       abits
 #define DTM_DMI_ADDRESS(abits)              ((0x400000000ULL * (1ULL<<abits)) + -17179869184)

Some of the awkwardness stems from the fact that sympy doesn't support a
<< operator, so I use ** internally and then print it out strangely.
It's correct, just not idiomatic.

### DIFF
--- a/registers.py
+++ b/registers.py
@@ -211,6 +211,46 @@ def write_definitions( fd, registers ):
                         toLatexIdentifier( registers.prefix, regid, f.name ),
                         f.name ) )
 
+class Macro:
+    def __init__(self, name, expressionText):
+        self.name = name
+        self.expression = sympy.simplify(expressionText)
+        self.atoms = self.expression.atoms(sympy.Symbol)
+
+    def prototype(self):
+        if self.atoms:
+            return "%s(%s)" % (self.name, ", ".join(str(a) for a in self.atoms))
+        else:
+            return self.name
+
+    def body(self):
+        return self.expression
+
+def sympy_to_c(expression):
+    """Implement our own string function, so we can replace 2** with 1<<."""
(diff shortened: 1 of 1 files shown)


pdonahue-ventana (approved):
The negative decimal number in that example is weird but it sounds unavoidable.

It seems like you don't need to break the <8 case into two pieces (0-7 and <0) since `"%d" % expression` would include the minus sign as appropriate without you having to explicitly add it in a separate case.  But that's not a big deal.

timsifive:
Good point. I think earlier I experimented with "-0x%x" and then I did need it.

pdonahue-ventana (approved):
I didn't previously notice the lack of U and ULL on the negative side.  This looks good.

PR #725, opened 2022-04-30
Select and issue a command can be simultaneous

timsifive:
It is not necessary to first select and then issue the command in
dmcontrol. The old language could have been interpreted that way.


pdonahue-ventana (approved):
This is better for the reason you intended and also because it previously implied that you couldn't select multiple harts.

PR #727, opened 2022-05-11
Add field values to XML register description.

timsifive:
Generate constants in the C header for the values. E.g.
```
 #define CSR_TDATA1_TYPE_NONE                0
 #define CSR_TDATA1_TYPE_LEGACY              1
 #define CSR_TDATA1_TYPE_MCONTROL            2
 #define CSR_TDATA1_TYPE_ICOUNT              3
 ...
```
Make formatting slightly more consistent in the PDF.

### DIFF
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -135,7 +135,7 @@
         </field>
         <field name="write" bits="16">
             When \FacAccessregisterTransfer is set:
-            <value v="0" name="arg">
+            <value v="0" name="arg0">
             Copy data from the specified register into {\tt arg0} portion
                of {\tt data}.
             </value>
@@ -263,7 +263,7 @@
         </field>
         <field name="0" bits="18:17" />
         <field name="write" bits="16">
-            <value v="0" name="arg">
+            <value v="0" name="arg0">
             Copy data from the memory location specified in {\tt arg1} into
             the low bits of {\tt arg0}. The value of the remaining bits of
             {\tt arg0} are \unspecified.
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -75,15 +75,15 @@
         </field>
         <field name="0" bits="23" />
         <field name="aarsize" bits="22:20">
-            <value v="2" name="32">
+            <value v="2" name="32bit">
             Access the lowest 32 bits of the register.
             </value>
 
-            <value v="3" name="64">
+            <value v="3" name="64bit">
             Access the lowest 64 bits of the register.
             </value>
 
-            <value v="4" name="128">
+            <value v="4" name="128bit">
             Access the lowest 128 bits of the register.
             </value>
 
@@ -233,23 +233,23 @@
             that same abstract command with \FacAccessmemoryAamvirtual cleared.
         </field>
         <field name="aamsize" bits="22:20">
-            <value v="0" name="8">
+            <value v="0" name="8bit">
             Access the lowest 8 bits of the memory location.
             </value>
 
-            <value v="1" name="16">
+            <value v="1" name="16bit">
             Access the lowest 16 bits of the memory location.
             </value>
 
-            <value v="2" name="32">
+            <value v="2" name="32bit">
             Access the lowest 32 bits of the memory location.
             </value>
 
-            <value v="3" name="64">
+            <value v="3" name="64bit">
             Access the lowest 64 bits of the memory location.
             </value>
 
-            <value v="4" name="128">
+            <value v="4" name="128bit">
             Access the lowest 128 bits of the memory location.
             </value>
         </field>
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -265,8 +265,8 @@
         <field name="write" bits="16">
             <value v="0" name="arg">
             Copy data from the memory location specified in {\tt arg1} into
-            the low bits of {\tt arg0}. Any remaining bits of {\tt arg0} now
-            have an undefined value.
+            the low bits of {\tt arg0}. The value of the remaining bits of
+            {\tt arg0} are \unspecified.
             </value>
 
             <value v="1" name="memory">
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -75,11 +75,17 @@
         </field>
         <field name="0" bits="23" />
         <field name="aarsize" bits="22:20">
-            2: Access the lowest 32 bits of the register.
+            <value v="2" name="32">
+            Access the lowest 32 bits of the register.
+            </value>
 
-            3: Access the lowest 64 bits of the register.
+            <value v="3" name="64">
+            Access the lowest 64 bits of the register.
+            </value>
 
-            4: Access the lowest 128 bits of the register.
+            <value v="4" name="128">
+            Access the lowest 128 bits of the register.
+            </value>
 
             If \FacAccessregisterAarsize specifies a size larger than the register's actual size,
             then the access must fail. If a register is accessible, then reads of \FacAccessregisterAarsize
@@ -91,37 +97,53 @@
             Table~\ref{tab:datareg}.
         </field>
         <field name="aarpostincrement" bits="19">
-            0: No effect. This variant must be supported.
+            <value v="0" name="disabled">
+            No effect. This variant must be supported.
+            </value>
 
-            1: After a successful register access, \FacAccessregisterRegno is
+            <value v="1" name="enabled">
+            After a successful register access, \FacAccessregisterRegno is
             incremented. Incrementing past the highest supported value
             causes \FacAccessregisterRegno to become \unspecified.  Supporting
             this variant is optional. It is undefined whether the increment
             happens when \FacAccessregisterTransfer is 0.
+            </value>
         </field>
         <field name="postexec" bits="18">
-            0: No effect. This variant must be supported, and is the only
+            <value v="0" name="disabled">
+            No effect. This variant must be supported, and is the only
             supported one if \FdmAbstractcsProgbufsize is 0.
+            </value>
 
-            1: Execute the program in the Program Buffer exactly once after
+            <value v="1" name="enabled">
+            Execute the program in the Program Buffer exactly once after
             performing the transfer, if any. Supporting this variant is
             optional.
+            </value>
         </field>
         <field name="transfer" bits = "17">
-            0: Don't do the operation specified by \FacAccessregisterWrite.
+            <value v="0" name="disabled">
+            Don't do the operation specified by \FacAccessregisterWrite.
+            </value>
 
-            1: Do the operation specified by \FacAccessregisterWrite.
+            <value v="1" name="enabled">
+            Do the operation specified by \FacAccessregisterWrite.
+            </value>
 
             This bit can be used to just execute the Program Buffer without
             having to worry about placing valid values into \FacAccessregisterAarsize or \FacAccessregisterRegno.
         </field>
         <field name="write" bits="16">
             When \FacAccessregisterTransfer is set:
-            0: Copy data from the specified register into {\tt arg0} portion
+            <value v="0" name="arg">
+            Copy data from the specified register into {\tt arg0} portion
                of {\tt data}.
+            </value>
 
-            1: Copy data from {\tt arg0} portion of {\tt data} into the
+            <value v="1" name="register">
+            Copy data from {\tt arg0} portion of {\tt data} into the
                specified register.
+            </value>
         </field>
         <field name="regno" bits="15:0">
           Number of the register to access, as described in
@@ -197,25 +219,39 @@
             physical accesses, but it must fail accesses that it doesn't
             support.
 
-            0: Addresses are physical (to the hart they are performed on).
+            <value v="0" name="physical">
+            Addresses are physical (to the hart they are performed on).
+            </value>
 
-            1: Addresses are virtual, and translated the way they would be from
+            <value v="1" name="virtual">
+            Addresses are virtual, and translated the way they would be from
             M-mode, with \FcsrMstatusMprv set.
+            </value>
 
             Debug Modules on systems without address translation (i.e. virtual addresses equal physical)
             may optionally allow \FacAccessmemoryAamvirtual set to 1, which would produce the same result as
             that same abstract command with \FacAccessmemoryAamvirtual cleared.
         </field>
         <field name="aamsize" bits="22:20">
-            0: Access the lowest 8 bits of the memory location.
+            <value v="0" name="8">
+            Access the lowest 8 bits of the memory location.
+            </value>
 
-            1: Access the lowest 16 bits of the memory location.
+            <value v="1" name="16">
+            Access the lowest 16 bits of the memory location.
+            </value>
 
-            2: Access the lowest 32 bits of the memory location.
+            <value v="2" name="32">
+            Access the lowest 32 bits of the memory location.
+            </value>
 
-            3: Access the lowest 64 bits of the memory location.
+            <value v="3" name="64">
+            Access the lowest 64 bits of the memory location.
+            </value>
 
-            4: Access the lowest 128 bits of the memory location.
+            <value v="4" name="128">
+            Access the lowest 128 bits of the memory location.
+            </value>
         </field>
         <field name="aampostincrement" bits="19">
             After a memory access has completed, if this bit is 1, increment
@@ -227,12 +263,16 @@
         </field>
         <field name="0" bits="18:17" />
         <field name="write" bits="16">
-            0: Copy data from the memory location specified in {\tt arg1} into
+            <value v="0" name="arg">
+            Copy data from the memory location specified in {\tt arg1} into
             the low bits of {\tt arg0}. Any remaining bits of {\tt arg0} now
             have an undefined value.
+            </value>
 
-            1: Copy data from the low bits of {\tt arg0} into the memory
+            <value v="1" name="memory">
+            Copy data from the low bits of {\tt arg0} into the memory
             location specified in {\tt arg1}.
+            </value>
         </field>
         <field name="target-specific" bits="15:14">
             These bits are reserved for target-specific uses.
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -148,7 +148,7 @@
             The hart single stepped because \FcsrDcsrStep was set. (priority 0, lowest)
             </value>
 
-            <value v="5" name="reset">
+            <value v="5" name="resethaltreq">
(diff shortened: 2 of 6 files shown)


timsifive:
@pdonahue-ventana can you take a look at this?

xml/abstract_commands.xml
```diff
@@ -91,37 +97,53 @@
...
+            Do the operation specified by \FacAccessregisterWrite.
+            </value>
 
             This bit can be used to just execute the Program Buffer without
             having to worry about placing valid values into \FacAccessregisterAarsize or \FacAccessregisterRegno.
         </field>
         <field name="write" bits="16">
             When \FacAccessregisterTransfer is set:
-            0: Copy data from the specified register into {\tt arg0} portion
+            <value v="0" name="arg">
```
pdonahue-ventana, line 138: Should the value names on 0 and 1 be read and write instead of arg and register?

xml/abstract_commands.xml
```diff
@@ -75,11 +75,17 @@
         </field>
         <field name="0" bits="23" />
         <field name="aarsize" bits="22:20">
-            2: Access the lowest 32 bits of the register.
+            <value v="2" name="32">
```
pdonahue-ventana, line 78: That shows up as:
> 2 (32): Access the lowest 32 bits of the register.

It seems at least slightly awkward for value 2 to have a name of 32 which is just another number.  Then you end up with two numbers before the colon which is possibly confusing.  Maybe "32b" or something like that?

xml/abstract_commands.xml
```diff
@@ -197,25 +219,39 @@
...
             M-mode, with \FcsrMstatusMprv set.
+            </value>
 
             Debug Modules on systems without address translation (i.e. virtual addresses equal physical)
             may optionally allow \FacAccessmemoryAamvirtual set to 1, which would produce the same result as
             that same abstract command with \FacAccessmemoryAamvirtual cleared.
         </field>
         <field name="aamsize" bits="22:20">
-            0: Access the lowest 8 bits of the memory location.
+            <value v="0" name="8">
```
pdonahue-ventana, line 236: Similar concern about the name being just a number.

xml/abstract_commands.xml
```diff
@@ -227,12 +263,16 @@
         </field>
         <field name="0" bits="18:17" />
         <field name="write" bits="16">
-            0: Copy data from the memory location specified in {\tt arg1} into
+            <value v="0" name="arg">
```
pdonahue-ventana, line 266: Should these also be "read" and "write"?

xml/abstract_commands.xml
```diff
@@ -227,12 +263,16 @@
...
+            Copy data from the memory location specified in {\tt arg1} into
             the low bits of {\tt arg0}. Any remaining bits of {\tt arg0} now
             have an undefined value.
```
pdonahue-ventana, line 269: I just noticed but should this be `\unspecified` instead of undefined?

xml/core_registers.xml
```diff
@@ -67,22 +97,30 @@
...
             These counters include the {\tt instret} CSR. On single-hart cores
             {\tt cycle} should be stopped, but on multi-hart cores it must keep
             incrementing.
+            </value>
 
             An implementation may hardwire this bit to 0 or 1.
         </field>
         <field name="stoptime" bits="9" access="WARL" reset="Preset">
-            0: Increment \Rtime as usual.
+            <value v="0" name="ignore">
```
pdonahue-ventana, line 115: "ignore" seems strange.  Should it be "increment" like stopcount above?  If "increment" is not ideal then maybe there's a common term that is appropriate for both stopcount and stoptime (like "normal" or something that implies that it behaves as specified in the non-debug specs.

xml/core_registers.xml
```diff
@@ -93,20 +131,32 @@
...
             (priority 1)
+            </value>
 
-            4: The hart single stepped because \FcsrDcsrStep was set. (priority 0, lowest)
+            <value v="4" name="step">
+            The hart single stepped because \FcsrDcsrStep was set. (priority 0, lowest)
+            </value>
 
-            5: The hart halted directly out of reset due to \Fresethaltreq. It
+            <value v="5" name="reset">
```
pdonahue-ventana, line 151: resethaltreq instead of reset?

xml/dm_registers.xml
```diff
@@ -78,19 +86,27 @@
 
         <!-- Fields that apply to the entire DM. -->
         <field name="authenticated" bits="7" access="R" reset="Preset">
-            0: Authentication is required before using the DM.
+            <value v="0" name="required">
```
pdonahue-ventana, line 89: This seems more like authenticated/unauthenticated or authenticated=true/false than required/passed.

xml/dm_registers.xml
```diff
@@ -101,26 +117,40 @@
...
 
-            1: \RdmConfstrptrZero--\RdmConfstrptrThree hold the address of the
+            <value v="1" name="valid">
+            \RdmConfstrptrZero--\RdmConfstrptrThree hold the address of the
             configuration structure.
+            </value>
         </field>
         <field name="version" bits="3:0" reset="3" access="R">
-            0: There is no Debug Module present.
+            <value v="0" name="not present">
```
pdonahue-ventana, line 131: "none" is a little more concise (but I'm OK with what you have, too).

xml/dm_registers.xml
```diff
@@ -101,26 +117,40 @@
...
+            </value>
 
-            3: There is a Debug Module and it conforms to version 1.0 of this
+            <value v="3" name="1.0">
+            There is a Debug Module and it conforms to version 1.0 of this
             specification.
+            </value>
 
-            15: There is a Debug Module but it does not conform to any
+            <value v="15" name="non-compliant">
```
pdonahue-ventana, line 150: "custom" seems less scary.  Unless the goal is to be scary so people are compliant.

xml/dm_registers.xml
```diff
@@ -403,7 +453,9 @@
         </field>
         <field name="0" bits="23:13" access="R" reset="0" />
         <field name="busy" bits="12" access="R" reset="0">
-            1: An abstract command is currently being executed.
+            <value v="1" name="busy">
```
pdonahue-ventana, line 456: Should there also be a "ready" value like with authbusy?

xml/dm_registers.xml
```diff
@@ -424,30 +476,46 @@
 
             This field only contains a valid value if \FdmAbstractcsBusy is 0.
 
-            0 (none): No error.
+            <value v="0 " name="none">
```
pdonahue-ventana, line 479: All of the values in this field have spaces in the `v=` string.

xml/dm_registers.xml
```diff
@@ -761,10 +837,14 @@
 
     <register name="System Bus Access Control and Status" short="sbcs" address="0x38">
         <field name="sbversion" bits="31:29" access="R" reset="1">
-            0: The System Bus interface conforms to mainline drafts of this
+            <value v="0" name="old">
```
pdonahue-ventana, line 840: "legacy"?

xml/hwbp_registers.xml
```diff
@@ -362,31 +398,51 @@
...
 
-            0: The trigger will attempt to match against an access of any size.
+            <value v="0" name="any">
+            The trigger will attempt to match against an access of any size.
             The behavior is only well-defined if $|select|=0$, or if the access
             size is XLEN.
+            </value>
 
-            1: The trigger will only match against 8-bit memory accesses.
+            <value v="1" name="8">
```
pdonahue-ventana, line 407: Similar name=number concern.

xml/hwbp_registers.xml
```diff
@@ -670,31 +758,51 @@
...
         <field name="size" bits="19:16" access="WARL" reset="0">
-            0: The trigger will attempt to match against an access of any size.
+            <value v="0" name="any">
+            The trigger will attempt to match against an access of any size.
             The behavior is only well-defined if $|select|=0$, or if the access
             size is XLEN.
+            </value>
 
-            1: The trigger will only match against 8-bit memory accesses.
+            <value v="1" name="8">
```
pdonahue-ventana, line 767: name=number

xml/hwbp_registers.xml
```diff
@@ -1135,10 +1267,14 @@
...
+            Ignore \FcsrTextraThirtytwoMhvalue.
+            </value>
 
-            4: This trigger will only match if the low bits of
+            <value v="4" name="equal">
+            This trigger will only match if the low bits of
             \RcsrMcontext/\RcsrHcontext equal \FcsrTextraThirtytwoMhvalue.
+            </value>
 
             1, 5: This trigger will only match if the low bits of
```
pdonahue-ventana, line 1279: 1,5,2,6 should have values.  That may require splitting them into separate entries.
- 1 matches when m/hcontext[0]=0 and m/hcontext[n:1]=mhvalue
- 5 matches when m/hcontext[0]=1 and m/hcontext[n:1]=mhvalue
- etc.

And maybe the names should be something like odd/even + context/vmid.

Tangent: In looking at this now, I don't exactly remember why we have value 4 plus 1 and 5.  Obviously 1/5 have an extra bit of range but why was that important?  In the mcontext CSR, it is recommended that the size be limited to 6 on RV32 and 13 on RV64 which means that we don't need the extra range.

xml/hwbp_registers.xml
```diff
@@ -1166,12 +1302,17 @@
             This field should be tied to 0 when S-mode is not supported.
         </field>
         <field name="sselect" bits="1:0" access="WARL" reset="0">
-            0: Ignore \FcsrTextraThirtytwoSvalue.
+            <value v="0" name="ignore">
+            Ignore \FcsrTextraThirtytwoSvalue.
+            </value>
 
-            1: This trigger will only match if the low bits of
+            <value v="1" name="equal">
```
pdonahue-ventana, line 1309: I would say that the names are 1=scontext and 2=asid.

xml/hwbp_registers.xml
```diff
@@ -1135,10 +1267,14 @@
             Data used together with \FcsrTextraThirtytwoMhselect.
         </field>
         <field name="mhselect" bits="25:23" access="WARL" reset="0">
-            0: Ignore \FcsrTextraThirtytwoMhvalue.
+            <value v="0" name="ignore">
+            Ignore \FcsrTextraThirtytwoMhvalue.
+            </value>
 
-            4: This trigger will only match if the low bits of
+            <value v="4" name="equal">
```
pdonahue-ventana, line 1274: All modes have some sort of equality comparison.  It's a matter of what values are being compared.  I would say name=mcontext on this one.

xml/jtag_registers.xml
```diff
@@ -63,11 +63,17 @@
...
+            Version described in spec version 0.11.
+            </value>
 
-            1: Version described in spec versions 0.13 and 1.0.
+            <value v="1" name="1.0">
+            Version described in spec versions 0.13 and 1.0.
+            </value>
 
-            15: Version not described in any available version of this spec.
+            <value v="15" name="undescribed">
```
pdonahue-ventana, line 74: custom?

pdonahue-ventana (commented):
I got out my fine-tooth comb.  I'm excited that this may help with my "does timing=0 mean before or after" problem.

xml/dm_registers.xml
pdonahue-ventana: Might want to use names that are similar to the version numbers in the DTM section.

xml/abstract_commands.xml
timsifive: I chose this because the field is already called write. So now you're setting write=arg or write=register, which makes more sense to me than write=write or write=read.

If I were renaming the field as well, maybe direction=into_arg and direction=into_register?

xml/abstract_commands.xml
timsifive: Hmm... that is awkward. I explicitly did not choose 32b because in a C macro all letters are upper case, and then is b bits or bytes? I could call it 32bit? That feels unnecessarily verbose, but that's better than confusing.

xml/abstract_commands.xml
timsifive: Similar response.

xml/abstract_commands.xml
pdonahue-ventana: 32bit works for me.

xml/abstract_commands.xml
timsifive: Done.

xml/core_registers.xml
timsifive: increment feels weird. Clocks don't increment. How about run/freeze? Does run/freeze make sense for stopcount?

xml/core_registers.xml
timsifive: Done.

xml/dm_registers.xml
timsifive: Changed to false/true.

xml/dm_registers.xml
timsifive: Changed to none. I'm not sure what version numbers in the DTM section you're referring to.

xml/dm_registers.xml
timsifive: Done.

xml/dm_registers.xml
timsifive: I've added it. In my change I decided to only modify existing cases of `<number>:` which is why here I only did 1.

xml/dm_registers.xml
timsifive: But you should've seen the macro I created to automatically make the XML tags and put the value in the right place!

xml/core_registers.xml
pdonahue-ventana: Yes, increment seems weird.  Counters don't even necessarily increment if they're counting something that never happens.  normal/freeze?

xml/dm_registers.xml
timsifive: geriatric? (j/k, done)

xml/hwbp_registers.xml
timsifive: Hah. I dug up the original PR because I was puzzling about this as well. What happened is that this used to be a 1-bit field, and then you turned it into a 3-bit field with the chosen encodings to retain backwards compatibility.

Can I punt on this case for this change? I'll make up names so they'll show up in the PDF consistently.

xml/abstract_commands.xml
pdonahue-ventana: Oh, I see.  I have always thought of this as read/write of the hart register which does a write/read of data0 (plus maybe data1 or more) as a necessary part of that operation.  But it's pretty clear from the text here and elsewhere that reading a hart register will write "arg0" which is some number of data registers.  If you call it "arg0" instead of "arg" then I think it makes sense and is consistent with everything else.

xml/abstract_commands.xml
pdonahue-ventana: "arg0" seems better.

xml/dm_registers.xml
pdonahue-ventana: > I'm not sure what version numbers in the DTM section you're referring to.

dtmcs.version: https://github.com/riscv/riscv-debug-spec/pull/727/files#diff-7e8d764a69e8e81a55604e4793a25d1077853b7e5f267607e1ec58366ed7f6afL65

But there's no equivalent to "not present" so as long as value 15 has a consistent name on both then I'm happy.

xml/hwbp_registers.xml
pdonahue-ventana: That sounds vaguely familiar.  I think that Greg came up with that idea.  Maybe it made sense to have an extra bit at the time, especially since there was going to be an extra bit for the VMID.  That doesn't completely make sense to me now but I don't think we should be changing that now.  Anyway...

It's OK to punt for now.  Get the 99% solution done and then incremental improvements can happen over time.

xml/hwbp_registers.xml
timsifive: Done.

xml/hwbp_registers.xml
timsifive: We really need names for all 3 of the options, and it's a bit of a mis-mash. I'm going with mcontext, mcontext_select, and vmid_select for now.

xml/jtag_registers.xml
timsifive: Done.

xml/abstract_commands.xml
timsifive: Done.

xml/core_registers.xml
timsifive: Done.

xml/abstract_commands.xml
timsifive: Done.

xml/dm_registers.xml
timsifive: I think they do match now that I've incorporated your suggestion to name 15 "custom."

xml/hwbp_registers.xml
timsifive: It happened at https://github.com/riscv/riscv-debug-spec/pull/549/files#diff-979319471706782a119ea34ac26af26137b1beedfbeec7603a5a314f841810ebL828 if you're curious.

timsifive:
I think I've addressed all your feedback. Thanks for the fine-toothed comb.

xml/hwbp_registers.xml
pdonahue-ventana: More archaeology: Looks like it was Ernie's idea in the "Debug support for Hypervisor extension" email thread from "Sun, 2 Aug 2020 16:27:23 -0600" and I guess I wrote it up.

xml/abstract_commands.xml
pdonahue-ventana: I don't see an update here.  I thought you were going with 32bit.

xml/abstract_commands.xml
pdonahue-ventana: Same here.

xml/hwbp_registers.xml
pdonahue-ventana: This isn't perfect but perfect is the enemy of good.

pdonahue-ventana (changes requested):
It might have been lost in all the back-and-forth but I think there are still 2 outstanding items.

xml/abstract_commands.xml
timsifive: Huh. For some reason `ag` didn't search that file, but did search other XML files. Fixed.

xml/abstract_commands.xml
timsifive: Fixed.

pdonahue-ventana (approved):

PR #729, opened 2022-05-12
Describe debug_defines.h.

timsifive:
To make that fit in a reasonable section, combine the separate
external/native debug implementation chapters into a single debugger
implementation chapter.

### DIFF
--- a/debugger_implementation.tex
+++ b/debugger_implementation.tex
@@ -1,4 +1,13 @@
-\chapter{External Debugger Implementation}
+\chapter{Debugger Implementation}
+
+\section{debug\_defines.h}
+
+The source at \href{https://github.com/riscv/riscv-debug-spec}
+{https://github.com/riscv/riscv-debug-spec} can generate a C header file that
+defines macros for every field in every register/abstract command mentioned in
+this document. To generate this, run {\tt make debug\_defines.h}.
+
+\section{External Debugger Implementation}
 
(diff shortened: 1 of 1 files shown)


debugger_implementation.tex
```diff
@@ -1,4 +1,13 @@
-\chapter{External Debugger Implementation}
+\chapter{Debugger Implementation}
+
+\section{debug\_defines.h}
+
+The source at \href{https://github.com/riscv/riscv-debug-spec}
+{https://github.com/riscv/riscv-debug-spec} can generate a C header file that
+defines macros for every field in every register/abstract command mentioned in
+this document. To generate this, run {\tt make debug\_defines.h}.
```
pdonahue-ventana, line 8: I'm concerned that this info could get stale (especially if people have old spec versions).  The [README](https://github.com/riscv/riscv-debug-spec/blob/master/README.md) would presumably have the latest command that's appropriate for the Makefile/XML that they actually download.  Maybe something like:
```
\section{C header file}

\href{https://github.com/riscv/riscv-debug-spec/blob/master/README.md}
{https://github.com/riscv/riscv-debug-spec/blob/master/README.md}
has instructions for generating a C header file that
defines macros for every field in every register/abstract command mentioned in
this document.
```

timsifive: Done. Thanks.

pdonahue-ventana (approved):

Issue #717, opened 2022-03-30
DebugModule.dataX preservation 


Dolu1990:
Hi,

I'm currently implementing the spec in hardware and there is one behaviour the spec isn't explicit about (or I missed it / implicit).

Let's say you run one abstract command which return/write data0. Can this new value be usable as a argument of the next abstract command ?

Because if it isn't, it allows more flexibility in the hardware implementation.

Location in the spec : 
https://github.com/riscv/riscv-debug-spec/blob/eeef4d580971d70b86e3fb498a0536654ae90608/xml/dm_registers.xml#L570

Thanks ^^

timsifive:
> Let's say you run one abstract command which return/write data0. Can this new value be usable as a argument of the next abstract command ?

That is certainly what I had imagined, but I cannot think of a good use case for it. (Copy value from one register to another, I suppose, but that's hardly common.)

I cannot imagine what optimization you can make if this isn't the case. How does it help you be require the debugger to always write data0 for an abstract command which will use it?

Dolu1990:
First, sorry for the long post XD

@pdonahue-ventana 

> Imagine that the answer is no and that data0 is allowed to spontaneously change. 

There may be a missunderstanding, i didn't mean that it could spontaneously change, i mean : 
Should the returned value of a abstract command in dataX be reusable for the next abstract command.

In practice, is the following sequance ok or not : 
1. abstract read command that writes data0
2. abstract write command that reads data0 (that was written by the abstract command 1)

@pdonahue-ventana + @timsifive 

Mostly, from a more abstract perspective, I think it is about "Do we consider dataX to be registers/memory for abstract commands" or "Do we consider dataX as data channels for abstract commands"

If we consider dataX as data channel, it means that they are used by the debuger as a way to write abstract command arguements, and read abstract commands returned values.

> How does it help you be require the debugger to always write data0 for an abstract command which will use it?

I didn't mean that data0 is lost after each command which used it, i mean that the data0 returned by a abstract command can't be used as arguement of the nexts abstract commands.

> I cannot imagine what optimization you can make if this isn't the case.

I think a drawing is better than 1000 words ^^

![image](https://user-images.githubusercontent.com/4832082/161047684-a0f36395-5ff8-4460-8656-de8961aa4e20.png)

To give some context, i'm targeting a FPGA softcore (https://github.com/SpinalHDL/NaxRiscv)
So the tradeoff compaired to a ASIC can be quite different. Here logic can be "very expensive".

[...]

PR #731, opened 2022-05-17
The program buffer may not be readable.

timsifive:
This allows a more efficient implementation, especially on FPGAs.
See #717.


pdonahue-ventana (approved):
There was an argument that the data registers written by one abstract command could possibly be used as an input to the next abstract command in order to copy data (but we decided that wasn't a compelling argument).  However, there's really no reasonable argument that the debugger needs to be able to read the program buffer.

PR #732, opened 2022-05-19
Add trigger action values, for debug_defines.h.

timsifive:
Leave the descriptions empty, and change registers.py to not emit
anything for values whose descriptions are empty.

I'm not a huge fan of the duplication of those values in the XML file,
but I don't want to invent a new mechanism to avoid it right now. (Ooh,
we could use M4 to generate the XML and use an existing mechanism...)

### DIFF
--- a/registers.py
+++ b/registers.py
@@ -85,12 +85,14 @@ class Value( object ):
         self.range = element.get( 'range' )
         if self.range:
             self.low, self.high = self.range.split( ":" )
-        self.text = element.text.strip()
+        self.text = (element.text or "").strip()
         self.tail = element.tail.strip()
         self.name = element.get( "name" )
         self.duplicate = element.get( "duplicate" )
 
     def to_latex( self ):
+        if not self.text:
+            return ""
         result = []
         if self.range:
             result.append("%s--%s (%s): %s\n" % ( self.low, self.high, self.name, self.text ))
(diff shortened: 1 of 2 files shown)


pdonahue-ventana (commented):
One thing I notice is that actions 2-5 are defined as "Reserved for use by the trace specification" in table 5.1.  But this defines them as on/off/notify.  Practically speaking, that's fine since that's how the trace spec defines them but I wonder if this change breaks the firewall where the definition is entirely external.

Should this PR use more generic names like trace2, trace3, trace4 (and trace5)?  Should we recognize that the trace spec is ratified so we can actually specify on/off/notify in table 5.1 (and leave encoding 5 as reserved for trace)?  Or am I just overreacting?

timsifive:
I dug those definitions out of the trace spec, so that the generated header file includes useful values. Technically trace[2-5] is more correct, but it's not nearly as useful. I am certain the trace spec isn't going to redefine those values so I think we're in the clear.

pdonahue-ventana:
I agree that it won't change.  My question is whether it's time for table 5.1 to reflect what the trace spec says.

timsifive:
I'll make a separate change to update Table 5.1.

pdonahue-ventana (approved):
Fair enough.

PR #738, opened 2022-06-13
Fix order of TAP states to be consistent.

pdonahue-ventana:
It says:
> In Update-DR the operation will start, and in Capture-DR its results will be captured into data.

So I think that more time might be necessary between Update-DR and Capture-DR rather than the other order.  (Arguably, it's between those states regardless of the order in which they're listed but specifying them in temporal order is less confusing.)


timsifive (approved):
Good catch!

PR #740, opened 2022-06-22
Remove clarifying information that is false.

timsifive:
Specifically, it wasn't true for abstractauto.


pdonahue-ventana (approved):
OK.  This sentence was not true for abstractauto.  (Discovery of that feature is possible by writing ones and then doing a read.)  All the information that a debugger needs for the discovery process is present throughout the spec and this sentence didn't really add anything.

PR #741, opened 2022-06-23
Elaborate on version detection.

timsifive:
The side effects only really occur in an extreme corner case.

I think. I don't remember anything about why this section was written,
and I wonder if I'm forgetting about something.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -600,14 +600,20 @@ should not implement that register.
 \begin{steps}{To detect the version of the Debug Module with a minimum of side
     effects, use the following procedure:}
     \item Read \RdmDmcontrol.
-    \item Write \RdmDmcontrol, preserving \FdmDmcontrolHartreset, \FdmDmcontrolHasel, \FdmDmcontrolHartsello, and
-        \FdmDmcontrolHartselhi from the value that was read, setting \FdmDmcontrolDmactive, and
-        clearing all the other bits.
-    \item Read \RdmDmcontrol until \FdmDmcontrolDmactive is high.
+    \item
+        If \FdmDmcontrolDmactive is 0 or \FdmDmcontrolNdmreset is 1:
+        \begin{enumerate}[nolistsep]
+            \item Write \RdmDmcontrol, preserving \FdmDmcontrolHartreset,
+            \FdmDmcontrolHasel, \FdmDmcontrolHartsello, and
+            \FdmDmcontrolHartselhi from the value that was read, setting
+            \FdmDmcontrolDmactive, and clearing all the other bits.
+            \item Read \RdmDmcontrol until \FdmDmcontrolDmactive is high.
+        \end{enumerate}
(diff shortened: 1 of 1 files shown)


pdonahue-ventana:
I agree that the write doesn't need to happen if the initial read showed that dmactive=1.  But does the write need to happen if dmactive=ndmreset=1?  I question "or \FdmDmcontrolNdmreset is 1" since I didn't think that ndmreset caused dmstatus to become inaccessible.

And can the side-effects actually happen?  If dmactive=1 then the new recipe says there's no write so there's no problem.  If dmactive=0 then ndmreset=0 (the reset value), so a write of dmactive=1 and ndmreset=0 doesn't change that.  Same thing with haltreq.  And resumereq is W1 so writing 0 has no effect and this was never a problem.  The only real side-effect is that the DM becomes active, though the spec correctly says that this isn't really visible outside the DM (until more writes are performed).

timsifive:
> I agree that the write doesn't need to happen if the initial read showed that dmactive=1. But does the write need to happen if dmactive=ndmreset=1? I question "or \FdmDmcontrolNdmreset is 1" since I didn't think that ndmreset caused dmstatus to become inaccessible.

I included the bit about ndmreset, because the description of dmcontrol includes:
> For forward compatibility, \FdmDmstatusVersion will always be readable when bit 1 (\FdmDmcontrolNdmreset) is 0 and bit 0 (\FdmDmcontrolDmactive) is 1.

> And can the side-effects actually happen?

I'm not convinced that they can. But this was in the spec presumably for a reason. It must have come from something. Is there an older version of the spec where they can happen, just not in 1.0?

pdonahue-ventana:
I see now.  If dmactive=1 and ndmreset=1 then you'll do the write and not preserve ndmreset or haltreq or resumereq.  That will definitely have side-effects for ndmreset and haltreq.  It won't have an adverse effect on resumereq in 1.0 but it will in 0.13 (when resumereq was R/W instead of W1).  And if dmactive=0 then you'll do a write which sets dmactive=1.  So the recipe "might have" all 4 side-effects, though not all at the same time.

If the recipe preserved the value of ndmreset, haltreq, and resumereq in addition to hasel, hartsello, etc. then the first 3 could probably be avoided.  Thoughts?  (I don't have a strong preference since this would be a very unusual case.)

timsifive:
We can't preserve ndmreset, because:
> For forward compatibility, \FdmDmstatusVersion will always be readable when bit 1 (\FdmDmcontrolNdmreset) is 0 and bit 0 (\FdmDmcontrolDmactive) is 1.
which implies that version might not be readable when ndmreset is 1.

We can't preserve haltreq because it's WARZ.

We can't preserve resumereq because it's W1.

I still think this PR is good as written.

pdonahue-ventana (approved):
Yes, this is better than what's already there.  I thought that there might be ways to make it even a little better but I now see that it's as good as it can be.

PR #743, opened 2022-07-14
Remove license from debug_defines.h.

timsifive:
CC BY 4.0 is incompatible with the GPL, meaning the file could not be
used in GPL projects like OpenOCD. RISC-V has plans to address this, but
the lawyers are working on other issues right now. Until that is
properly resolved, we're reverting to the way things were before #641
and removing the license line.


cetola (approved):
LGTM

The spec is licensed with CC BY 4.0, but using that text in comments for code is a perfectly acceptable use of the spec text.

pdonahue-ventana (approved):
OK with me.  I'm a little confused about why this was a problem for GPL projects but I'm not a lawyer.

Issue #746, opened 2022-08-16
Why should hardware clear the action field whenever the action field is 1?


YenHaoChen:
Hello,

I cannot understand the following compound sentence with five clauses. Can you share some words on that?
![image](https://user-images.githubusercontent.com/39526191/184765305-1fc3c453-de2e-4b76-a1ca-7de2a1225422.png)

Thank you
Best regards,
YenHaoChen

pdonahue-ventana:
It's illegal to have action=1 and dmode=0 (as specified in the previous sentence) so the suggestion is that WARL sanitization should clear action rather than setting dmode.

YenHaoChen:
Thank you for clarification.

PR #748, opened 2022-08-16
Clarify writing 1 to action on dmode=0

YenHaoChen:
**Clarify/Simplify the sentence:**
 - Since the tdata registers are WARL, hardware should clear the action
field whenever the action field is 1, dmode is cleared, and the new
value of the action field would also be 1.

**Change into:**
 - If dmode is 0, writing 1 to the action clears the action field.

Reference: https://github.com/riscv/riscv-debug-spec/issues/746

### DIFF
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -66,8 +66,8 @@ Value & Description \\
 
     This action is only legal when the trigger's \FcsrTdataOneDmode is 1.
     Since the {\tt tdata} registers are WARL, hardware should clear the action
-    field whenever the action field is 1, \FcsrTdataOneDmode is cleared, and the
-    new value of the action field would also be 1. \\
+    field whenever the action field is 1, the new value of \FcsrTdataOneDmode would be 0, and the
+    new value of the action field would be 1. \\
 \hline
 2 & Trace on, described in the trace specification. \\
 \hline


ernie-sifive:
The new statement is not accurate.  For example, action should not be cleared if dmode=0 and action=2 and you write a new value with dmode=1 and action=2.

pdonahue-ventana:
In addition to Ernie's concern, this PR also restates a suggestion ("should") as a requirement.

YenHaoChen:
> The new statement is not accurate. For example, action should not be cleared if dmode=0 and action=2 and you write a new value with dmode=1 and action=2.

The example:     If dmode is 0, writing 1 to dmode should not clear the action field.
New statement: If dmode is 0, writing 1 to the action clears the action field.

The new statement seems accurate in the example to me. 

From the reference https://github.com/riscv/riscv-debug-spec/issues/746, the purpose of this sentence is to suggest a behavior of writing illegal action=1 when dmode=0 (illustrated in the previous sentence). I believe the new statement serves the purpose better.

pdonahue-ventana:
I don't think that captures the case being discussed here.  The spec effectively says something like this:
```
action_t action;
bit_t dmode;

void write(action_t new_action, bit_t new_dmode) {
  if (action == 1 && new_dmode == 0 && new_action == 1) {
    new_action = 0;
  }
  action = new_action;
  dmode = new_dmode;
}
```
(and I'm assuming that this write function is called by the set and clear functions)

Your new proposal says something like this:
```
void write(action_t new_action, bit_t new_dmode) {
  if (dmode == 0 && new_action == 1) {
    new_action = 0;
  }
  action = new_action;
  dmode = new_dmode;
}
```
That doesn't handle the legal case where you're writing dmode=1 and action=1 in the same operation.

YenHaoChen:
> I don't think that captures the case being discussed here. The spec effectively says something like this:
> 
> ```
> action_t action;
> bit_t dmode;
> 
> void write(action_t new_action, bit_t new_dmode) {
>   if (action == 1 && new_dmode == 0 && new_action == 1) {
>     new_action = 0;
>   }
>   action = new_action;
>   dmode = new_dmode;
> }
> ```
> 
> (and I'm assuming that this write function is called by the set and clear functions)
> 
> Your new proposal says something like this:
> 
> ```
> void write(action_t new_action, bit_t new_dmode) {
>   if (dmode == 0 && new_action == 1) {
>     new_action = 0;
>   }
>   action = new_action;
>   dmode = new_dmode;
> }
> ```
> 
> That doesn't handle the legal case where you're writing dmode=1 and action=1 in the same operation.

From the code, I think you meant to be 
```
action_t action;
bit_t dmode;

void write(action_t new_action, bit_t new_dmode) {
  if (new_dmode == 0 && new_action == 1) {
    new_action = 0;
  }
  action = new_action;
  dmode = new_dmode;
} 
```
The sentence should be "If writing dmode to 0 and the action to 1, hardware should clear the action field."

pdonahue-ventana:
The [spec](https://github.com/riscv/riscv-debug-spec/blob/master/Sdtrig.tex#L68) says:
> Since the tdata registers are WARL, hardware should clear the action
    field whenever the action field is 1, dmode is cleared, and the
    new value of the action field would also be 1.

There are three conditions in the above:
- the action field is 1
- dmode is cleared
- the new value of the action field would also be 1

I believe that my code correctly captures all three conditions.

YenHaoChen:
>

YenHaoChen:
> The [spec](https://github.com/riscv/riscv-debug-spec/blob/master/Sdtrig.tex#L68) says:
> 
> > Since the tdata registers are WARL, hardware should clear the action
> > field whenever the action field is 1, dmode is cleared, and the
> > new value of the action field would also be 1.
> 
> There are three conditions in the above:
> 
> * the action field is 1
> * dmode is cleared
> * the new value of the action field would also be 1
> 
> I believe that my code correctly captures all three conditions.

There is a difference between the code and the sentence in the second condition. Without the code, the spec fails to present the concept we are discussing.
* dmode is cleared => the new value of the dmode is cleared

Additionally, the first condition does not consist of the concept of the previous sentence. If the first condition stands, one can write the illegal value to the register (when the action is 2).

pdonahue-ventana:
I was taking shortcuts with my code since I only wanted to highlight the one case.  Here's something more complete:
```
action_t action;
bit_t dmode;

void write(action_t new_action, bit_t new_dmode) {
#ifdef USE_SUGGESTED_WARL_BEHAVIOR
  if (action == 1 && new_dmode == 0 && new_action == 1) {
    new_action = 0;  // make this legal in one particular way
  }
#endif
  if ((new_action == 1 && new_dmode == 0) || !is_legal_encoding(new_action)) {
    apply_warl_behavior(*new_action, *new_dmode);  // make them legal in an implementation-specific way
  }
  action = new_action;
  dmode = new_dmode;
}
```
Of course, there are other fields which need WARL sanitization.  This only deals with the two fields.

I think that your proposal would be simpler.  It's just not quite what the existing stable spec says.  But that's just a suggestion so maybe that's OK?

YenHaoChen:
> I was taking shortcuts with my code since I only wanted to highlight the one case. Here's something more complete:
> 
> ```
> action_t action;
> bit_t dmode;
> 
> void write(action_t new_action, bit_t new_dmode) {
> #ifdef USE_SUGGESTED_WARL_BEHAVIOR
>   if (action == 1 && new_dmode == 0 && new_action == 1) {
>     new_action = 0;  // make this legal in one particular way
>   }
> #endif
>   if ((new_action == 1 && new_dmode == 0) || !is_legal_encoding(new_action)) {
>     apply_warl_behavior(*new_action, *new_dmode);  // make them legal in an implementation-specific way
>   }
>   action = new_action;
>   dmode = new_dmode;
> }
> ```
> 
> Of course, there are other fields which need WARL sanitization. This only deals with the two fields.
> 
> I think that your proposal would be simpler. It's just not quite what the existing stable spec says. But that's just a suggestion so maybe that's OK?

Yes, I agree the spec is OK.

The concerns are:
1. The spec has a difference from your code, i.e., (dmode is cleared => the new value of dmode is cleared). 
2. Without the clarification from your code, I thought spec states 
```if (action == 1 && dmode == 0 && new_action == 1) new_action = 0```,
which is incorrect and confusing.
3. The spec should clearly and elegantly present the intended concept even without the code in the Github discussion.
4. The original sentence is wordy and does not cover all cases for avoiding illegal dmode=0, action=1.

With the above concerns, this PR provides a suggestion for improvement. Although it may not be the same as the original sentence, it serves the purpose of illustrating the concept better.
I think "may" would be better than "should" for a suggestion.
``` Writing dmode to 0 and action to 1 may clear the action field. ```

**Comment by pdonahue-ventana** [comment]
I think that this is all stemming from an assumption that "dmode is cleared" means "dmode is clear" which is not the intent.  However, "dmode is cleared" means "dmode is written with 0" rather than "dmode contains 0."  I think that ["cleared"](https://www.merriam-webster.com/dictionary/cleared) is a verb (which deals with the act of clearing) and "clear" is the adjective (and would be the word to use to describe the state of something).  However, that may not be universally understood.  How about just getting rid of the confusing "cleared" term without modifying the other conditions:
```
    Since the {\tt tdata} registers are WARL, hardware should clear the action
    field whenever the action field is 1, the new value of \FcsrTdataOneDmode would be 0, and the
    new value of the action field would be 1.
```

YenHaoChen:
That makes lots more sense to me.

pdonahue-ventana (approved):
This looks OK to me.  We'll see what Tim thinks.

timsifive (approved):

YenHaoChen:
Is there any missing from my side?

timsifive:
Nope. Thanks for the reminder.

PR #752, opened 2022-09-08
Clarify dcsr.cause when halt groups are used.

timsifive:


pdonahue-ventana (commented):
I know that it says "should" which is not "must" but it seems like it might be helpful to less-savvy readers to highlight that it won't necessarily be 6 (as explained in the dcsr.cause description).  Maybe "should be set to 6, but may be set to 3" or something like that.

pdonahue-ventana (approved):

PR #758, opened 2022-09-28
Clarify that PMP shouldn't prevent debug.

### LINKED ISSUE #755: PMP interaction with debug spec
The  debug spec section A.2 says:
     Note that for debug to be possible, the PMP must not disallow fetches, loads, or stores
      in the address range associated with the Debug Module when the hart is in Debug Mode.

This is ambiguous. 
This could mean that there is a hidden  PMP entry that allows 
all accesses to the debug range when the hart is in debug mode.
Or, 
it could mean that Mm-de SW must never configure a (highest priority) locked entry that 
covers the debug module address range with anything other than RWX permissions.

Which, if any, of these interpretations is correct? 
Or are both correct; it is an implementation choice?

The converse of this is also unclear: can Mmode SW access the debug mode address range without restrictions?
 
I am assuming that, in debug mode, the PMP ), and that

timsifive:
Fixes #755.

### DIFF
--- a/implementations.tex
+++ b/implementations.tex
@@ -77,8 +77,9 @@ For additional flexibility, \RdmProgbufZero, etc. are mapped into regular memory
 immediately preceding \RdmDataZero, in order to form a contiguous region of memory which
 can be used for either program execution or data transfer.
 
-Note that for debug to be possible, the PMP must not disallow fetches, loads, or stores
-in the address range associated with the Debug Module when the hart is in Debug Mode.
+Note that for debug to always be possible, the PMP must not disallow fetches,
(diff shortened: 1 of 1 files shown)


pdonahue-ventana (approved):
Good.  My proposed addition of "hardware" wasn't considering things like Spike.  Your wording sufficiently describes that it's the responsibility of the implementation and not of M mode software.

ernie-sifive:
You might want to add this item to the list of non-backward-compatible changes in v1.0.  This is adding a new requirement on hardware.

timsifive:
This is an appendix, which merely suggests possible implementations. It's not part of the spec itself.

Issue #756, opened 2022-09-28
Trigger chain with different types


YenHaoChen:
Hello,

The spec states, "Debuggers should not terminate a chain with a trigger with a different type." I think that means the spec allows a trigger chain with different types.
![image](https://user-images.githubusercontent.com/39526191/192690209-0d327db9-c692-47de-b3e1-56af0bc3c0c8.png)

Consider the configuration below, where the trigger chain consists of mcontrol and mcontrol6. The trigger index=3 has a different type from others. If we allow such a chain, this trigger chain can fire on memory access that matches all six triggers.
![圖片1](https://user-images.githubusercontent.com/39526191/192693031-dd995803-bd72-4e5d-b179-f481ed1e20ec.png)

Consider another configuration below, where the trigger chain consists of mcontrol and icount. A difference between mcontrol6 and icount is that the icount does not have the chain field. What should be the firing condition in this configuration? Is this still a chain that fires when matching all six triggers? Or, will this configuration have two distinct firing conditions, i.e., an icount (index 3) and a chain (index 4 and 5)?
![圖片2](https://user-images.githubusercontent.com/39526191/192693070-f74ac47d-7eff-4ef6-aae2-bc920bbab4c4.png)

pdonahue-ventana:
> Consider the configuration below, where the trigger chain consists of mcontrol and mcontrol6. The trigger index=3 has a different type from others. If we allow such a chain, this trigger chain can fire on memory access that matches all six triggers.

Implementations are not expected to support both mcontrol and mcontrol6.  The top of the mcontrol6 section says that it "replaces mcontrol in newer implementations".  Section 1.2.1.4 says that one new feature of v1.0 is that mcontrol is deprecated.  Maybe the mcontrol section should have a more prominent statement that it is deprecated and only described in the spec for backward compatibility.

> Consider another configuration below, where the trigger chain consists of mcontrol and icount. A difference between mcontrol6 and icount is that the icount does not have the chain field. What should be the firing condition in this configuration? Is this still a chain that fires when matching all six triggers? Or, will this configuration have two distinct firing conditions, i.e., an icount (index 3) and a chain (index 4 and 5)?

[...]

PR #759, opened 2022-09-30
More prominently say that mcontrol is deprecated.

pdonahue-ventana:
Clears up confusion from #756 about chaining mcontrol and mcontrol6.


timsifive (approved):

pdonahue-ventana:
@timsifive: I think that this can be merged.  I have always let you do that since I'm not sure if we want a merge commit vs. squash and merge, what sort of commit comment we generally use, etc.  A couple of other PRs have also been approved but not merged yet.

Issue #760, opened 2022-10-11
Data bitfield inside dmi register


ToTamire:
What JTAG should read after successfull write transaction on debug module interface? I mean when JTAG do next scan chain to read dmi register and it's op bitfield returns 0 (previous operation completed successfully). Data bitfield in dmi register should contain: written data, register value before write or register value after write?

timsifive:
Hmm... the spec is quite unclear about this. I don't think it matters what dmi.address and dmi.data contain after a write. I made #762 to try to clarify, and get feedback on what we really want to say.

pdonahue-ventana:
#762 is merged.

PR #762, opened 2022-10-12
Clarify dmi.{data,address} behavior.

timsifive:
It's not clear, as reported in #760.

### DIFF
--- a/xml/jtag_registers.xml
+++ b/xml/jtag_registers.xml
@@ -117,14 +117,24 @@
             This operation should never result in a busy or error response.
             The address and data reported in the following Capture-DR
             are undefined.
+
+            This operation leaves the values in \FdtmDmiAddress and \FdtmDmiData
+            \unspecified.
             </value>
 
             <value v="1" name="read">
-            Read from \FdmSbaddressZeroAddress.
+            Read from \FdtmDmiAddress.
+
+            When this operation succeeds, \FdtmDmiAddress contains the address
+            that was read from, and \FdtmDmiData contains the value that was
+            read.
             </value>
 
             <value v="2" name="write">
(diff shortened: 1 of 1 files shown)


xml/jtag_registers.xml
```diff
@@ -117,14 +117,24 @@
             This operation should never result in a busy or error response.
             The address and data reported in the following Capture-DR
             are undefined.
+
+            When this operation leaves the values in \FdtmDmiAddress and
+            \FdtmDmiData unchanged.
```
pdonahue-ventana, line 122: I think that there's a missing word (or extra word) here.  Also, it could be ambiguous whether this is unchanged from the value written by the nop command or unchanged from a previous value (where the command is parsed to determine if it's a nop and only if it's not a nop will the Update-DR update those other fields).  Since those are R/W fields and they would not be updated by hardware on a nop, my assumption has been that the required behavior is the former: the value seen on a read is the value written by the nop command.

timsifive: Yes, no grammar.

In spike, scanning NOP results in both data and address being updated.
In a SiFive FPGA image, scanning NOP results in data and address being zero'd out.

Since it doesn't matter for anything that I can think of, I think we should just say it's undefined.

timsifive: Unspecified, I mean.

pdonahue-ventana (approved):

xml/jtag_registers.xml
timsifive: Having DmiData reflect a possible value would not be a backwards-compatible change, and we're trying not to functionally change the spec anymore. Really what you're asking for is a separate dmi.op that performs a read followed by a write. That's something we could discuss, but not in this PR.

PR #764, opened 2022-10-17
halt/resume groups don't apply to unavailable harts.

timsifive:
Clarify this by explicitly saying the behavior only applies to running/halted harts respectively.

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -250,9 +250,10 @@ halt and resume groups.
     that's a member of the halt group fires:}
 \item That hart halts normally, with \FcsrDcsrCause reflecting the original
     cause of the halt.
-\item All the other harts in the halt group will quickly halt, even if they are
-    currently in the process of resuming. \FcsrDcsrCause for those harts should
-    be set to 6, but may be set to 3.
+\item All the other harts in the halt group that are running will quickly halt.
+    \FcsrDcsrCause for those harts should be set to 6, but may be set to 3.
+    Other harts in the halt group that are halted but have started the process
+    of resuming must also quickly become halted, even if they do resume briefly.
 \item Any external triggers in that group are notified.
 \end{steps}
 Adding a hart to a halt group does not automatically halt that hart, even if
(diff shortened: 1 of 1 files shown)


debug_module.tex
```diff
@@ -250,20 +250,22 @@ \section{Halt Groups, Resume Groups, and External Triggers} \label{hrgroups}
...
 \begin{steps}{When any hart in a resume group resumes, or an external trigger
     that's a member of the resume group fires:}
-\item All the other harts in that group will quickly resume as soon as any
-    currently executing abstract commands have completed, except for the harts
-    that are in the process of halting.
+\item All the other harts in that group that are halted will quickly resume as
+    soon as any currently executing abstract commands have completed.
     Each hart in the group sets its resume ack bit\index{resume ack bit} as soon as it has resumed.
+    Harts that are in the process of halting must complete that process and stay
+    halted.
```
pdonahue-ventana, line 268: This says that if it is in the process of halting then it must ultimately be halted but if it had gotten far enough into the halting process to be officially halted then it must resume and ultimately be running.  That's another race condition and I think that the requirements should be somewhat loose (though it can be useful to have suggestions about what to prioritize if two things happen in the same cycle).

For instance, what should the final running/halted state of hart A be if you have:
- hart A is in halt group H and resume group R
- debugger resumes another hart in resume group R and within +/- one cycle another hart in H hits a trigger with action=1

It all probably depends on propagation delays of various signals.  If the debugger (or any other observer) can't determine which event happened first then it doesn't really matter which one is prioritized.  The priority truly matters only if the debugger was able to observe that a hart is in some particular state and then it knows that some event happened after that observation.

timsifive: I don't think I changed anything here, just rephrased.

I believe the original rationale is that it's better to leave a hart halted than running if it's not clear what should happen, because then the debugger has control.

In your example, hart A:
* starts out halted
* then starts resuming because another hart in R resumes
* then learns that it needs to halt because a hart in H halted
* must quickly become halted, even though it has started the resume process

pdonahue-ventana: I think that the question is what does "in the process of halting" mean?  There's some point where you're not halted and some point at which you are halted.  There's some gray area where you've committed to halting but are not officially halted yet.

Consider this:
- resume groups are transmitted directly between harts, not via a central DM (i.e. there are wires)
- hart A hit a trigger and is in debug mode
- hart A has not yet told the DM that it's halted (due to propagation delays of the "I'm halted" wire, it hasn't executed the park loop store yet, etc.)
- hart B is told by DM to resume
- hart B resumes and directly tells hart A to resume

This seems like hart A is "in the process of halting" since the fact that it has halted is not visible to the external debugger.  I don't want to prohibit hart A from resuming in such an implementation.  (This probably wouldn't be a problem in an implementation where all halt/resume activity goes through a central clearinghouse in DM.  But I'm trying to consider alternate implementations and I think it's better to only provide guidance rather than mandates in these race condition cases.)

I propose changing "must" to "should" on line 267.

timsifive: Agreed. Done.

pdonahue-ventana: Looks good.

pdonahue-ventana (approved):

Issue #766, opened 2022-10-18
stoptime description needs clarification


aswaterman:
As currently written, the definition of the `stoptime` field implies that `time` and `mtime` are handled differently: that `stoptime` affects the `time` CSR always, but only affects `mtime` when all harts have `stoptime`=1.  But `time` and `mtime` are not independently controllable.  As the priv spec says, "The time CSR is a read-only shadow of the memory-mapped mtime register".

This might just be a matter of imprecise wording.  I suggest clarifying it by dropping all reference to `time`.  Here's a possible rewording that retains all of the semantic information that isn't contradictory:

"If `stoptime`=0, then `mtime` increments as usual.  If all harts have `stoptime`=1, then `mtime` is allowed to stop incrementing."

Of course, this does mean that `stoptime` might have no effect even when all harts have `stoptime`=1, which is odd.  But that is what the current spec effectively says, and I'm only trying to clarify, not make any normative changes.

Looks like @pdonahue-ventana most recently edited the text in 6ea272fc07b6ec7409cf46b3ab9beac44bf84e2b

cc @jhauser-us, who pointed this out.

aswaterman:
My point is that they aren’t independently controllable—that would contradict the priv spec, which says one is a shadow of the other. The debug spec can’t override that.

aswaterman:
The questions you raise are not within the bailiwick of the debug specification.  Perhaps we should resolve them elsewhere and revisit this concern later.  Since you're the one who brought it up to begin with, I'll leave it to you to decide whether to close this ticket for the time being.

jhauser-us:
Sorry for muddying this issue.  My concerns about ambiguity in the ISA specs need to be addressed elsewhere.  It came up here because I'd intended to back you up, Andrew, by quoting the ISA specs, then realized the specs didn't actually let me do that.

Getting back to the point, the intention of the specs has long been that a RISC-V system with multiple harts has a single, global `mtime`, and that the read-only `time` CSR in every hart is a copy of this global `mtime`.  (The value read from `time` may include a small time offset due to physical communication delays, but the system needs to try to equalize this offset across all harts.)  Given this intention, it's impossible to stop incrementing `mtime` for one hart separately from other harts, or to stop incrementing any hart's `time` CSR independently of `mtime`.

pdonahue-ventana:
> My point is that they aren’t independently controllable—that would contradict the priv spec, which says one is a shadow of the other. The debug spec can’t override that.

[...]

PR #769, opened 2022-10-26
Clarify stoptime behavior.

timsifive:
More accurate state the behavior specified in Section 4.1, so people don't have to read that in order to really understand this part.

See #766.

### DIFF
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -161,7 +161,7 @@
 
             <value v="1" name="freeze">
             \Rtime is frozen at the time that Debug Mode was entered. When
-            leaving Debug Mode, \Rtime will snap forward to reflect the latest
+            leaving Debug Mode, \Rtime will reflect the latest
             value of \Rmtime again.
 
             While all harts have \FcsrDcsrStoptime=1 and are in Debug Mode,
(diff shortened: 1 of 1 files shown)


xml/core_registers.xml
```diff
@@ -156,13 +156,16 @@
...
+            \Rtime continues to reflect \Rmtime.
             </value>
 
             <value v="1" name="freeze">
-            Don't increment \Rtime while in Debug Mode.  If all harts
-            have \FcsrDcsrStoptime=1 and are in Debug Mode then \Rmtime
-            is also allowed to stop incrementing.
+            \Rtime is frozen at the time that Debug Mode was entered. When
+            leaving Debug Mode, \Rtime will snap forward to reflect the latest
+            value of \Rmtime again.
```
pdonahue-ventana, line 165: I don't like the term "snap forward" because it seems like slang.  Thinking some more, it might not even be technically true.  If mtime stops counting (due to the sentence below) then it actually won't snap forward at all.  It'll just continue to increment when mtime starts incrementing again.

```suggestion
            \Rtime is frozen at the time that Debug Mode was entered. When
            leaving Debug Mode, \Rtime will reflect the latest
            value of \Rmtime again.
```

pdonahue-ventana (changes requested):
Also, https://github.com/riscv/riscv-debug-spec/issues/766#issuecomment-1291287732 suggested:
> We also suggest adding a note that we expect the bit to be hardwired to 0 in many multi-hart implementations.

pdonahue-ventana (approved):
This looks good to me.

Issue #772, opened 2022-10-28
Debug CSR DCSR/DPC/DSCRATCH access using CSR instructions


jrahmeh:
Section 4.8 of the debug spec (version 1.0.0) states that:

> They are CSRs, accessible using the RISC-V csr opcodes and optionally also using abstract debug commands.

However, the RISCV privileged spec states that:

> Machine-mode standard read-write CSRs 0x7A0–0x7BF are reserved for use by the debug system. Of these CSRs, 0x7A0–0x7AF are accessible to machine mode, whereas 0x7B0–0x7BF are only visible to debug mode. Implementations should raise illegal instruction exceptions on machine-mode access to the latter set of registers.

If a hart is executing a csr opcode it must be either running (not in debug mode) or executing from the program buffer.  In either case, the privilege mode is M or lower (section 4.1 of debug spec says program buffer executes in M privilege).  So is DCSR  accessible using a csr opcode?

pdonahue-ventana:
> If a hart is executing a csr opcode it must be either running (not in debug mode) or executing from the program buffer. In either case, the privilege mode is M or lower (section 4.1 of debug spec says program buffer executes in M privilege).

You're always in Debug Mode when you're halted.  For instance, section 4.1 says "Debug Mode is a special processor mode used only when a hart is halted for external debugging" and "When executing code due to an abstract command, the hart stays in Debug Mode."  Debug Mode is a distinct mode.

Section 4.8 says "These registers are only accessible from Debug Mode."  And the privileged spec agrees that those CSR addresses are "only visible to debug mode."

However, I see the confusion.  The intention of "All operations are executed with machine mode privilege" is that memory operations behave as if they're in M mode (using bare physical addresses) and that all CSRs accessible to M mode are accessible to Debug mode.

jrahmeh:
Would it be possible to add that to the spec to make it clearer? Maybe something like: "All operations are executed in debug mode which is identical to M mode for memory access and has additional privileges than M mode for CSR access."

pdonahue-ventana:
Yes.  I'll do something like that when I get a chance.  This can remain open in the meantime.

pdonahue-ventana:
Fixed in #773.

pdonahue-ventana:
Yes.  These Sdext CSRs are listed in section 4.10 which says "These registers are only accessible from Debug Mode."

PR #773, opened 2022-11-11
Explicitly allow debug mode to access debug mode CSRs for #772.

pdonahue-ventana:
This ought to address @jrahmeh's concern.

### DIFF
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -21,10 +21,12 @@ How Debug Mode is implemented is not specified here.
 
 \begin{steps}{When executing code due to an abstract command, the hart stays in
     Debug Mode and the following apply:}
-\item All operations are executed with machine mode privilege, except that
+\item All memory accesses are executed as if in machine mode, except that
     \FcsrMstatusMprv in \Rmstatus may be ignored according to \FcsrDcsrMprven.
     Full permission checks, or a relaxed set of permission checks, will apply
     according to \FdmAbstractcsRelaxedpriv.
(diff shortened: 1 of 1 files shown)


Sdext.tex
```diff
@@ -21,10 +21,12 @@ \section{Debug Mode} \label{debugmode}
 
 \begin{steps}{When executing code due to an abstract command, the hart stays in
     Debug Mode and the following apply:}
-\item All operations are executed with machine mode privilege, except that
+\item All memory accesses are executed as if in machine mode, except that
     \FcsrMstatusMprv in \Rmstatus may be ignored according to \FcsrDcsrMprven.
     Full permission checks, or a relaxed set of permission checks, will apply
     according to \FdmAbstractcsRelaxedpriv.
+\item CSR accesses are allowed to all CSRs that can be accessed from machine
+    mode.  Debug mode CSRs are also accessible.
```
timsifive, line 29:
```suggestion
    mode.  Debug Mode CSRs are also accessible.
```

timsifive (commented):
This looks OK, but I worry that "all operations" might include some things that are neither "memory accesses" nor "CSR accesses." Are there instructions that only function in certain modes? Other things?

pdonahue-ventana:
Yes, there are probably other things that are mode-dependent.  I can't think of anything right now.  Maybe some of the things that trap instructions like sfence.vma?  If nothing else, you could have a custom extension that's M-mode-only.  Let me do some wordsmithing.

timsifive (approved):

timsifive (approved):

Issue #771, opened 2022-10-28
PMP interaction with debug spec, updated text


silabs-robin:
Hi.

Referring to the changes made in this https://github.com/riscv/riscv-debug-spec/pull/758 PR, relating to this https://github.com/riscv/riscv-debug-spec/issues/755 issue.

> Note that for debug to always be possible, the PMP must not disallow fetches, loads, or stores in the address range associated with the Debug Module when the hart is in Debug Mode, regardless of how the PMP is configured.

There are several potential issues with that sentence, as detailed below.

--------

The sentence is still ambiguous.

It could mean:
_"If an implementation requires debug to always be possible, then [..]"_
Or:
_"Debug must always be possible, therefore [...]"_

_(I could not find anywhere in the spec stating that debug must always be possible.)_

Ambiguities are important to clean up. I had, for instance, interpreted the previous text as Allen's second alternative (SW) before that ambiguity was now resolved.

--------

We are worried that overruling the PMP settings is a potential security hole, and therefore prefer if the sentence is changed to explicitly make both interpretations a legal option (i.e. overruling the PMP or not overruling the PMP).

Suppose that a debug-related signal is glitched high, then the PMP settings become overruled and we would loose the extra protection that could have been provided by the PMP.
For example, software could configure the PMP such that debug access is never allowed in the field; even if a debug-related signal is glitched the PMP would prevent debug mode execution.

--------

Lastly, should this also include the PMA and not only the PMP?

--------

ps.

I know this is about an appendix and not "the spec itself", but it is bundled with the spec and people use this as  a reference and it would be good to avoid confusion.

Regards Robin.

pdonahue-ventana:
> (I could not find anywhere in the spec stating that debug must always be possible.)

Debug must always be possible because the spec says that if you do X then Y happens, not that Y happens only if debug is possible. (e.g. an abstract command with transfer=1 and write=0 will read the register, not that it will read the register sometimes but it will be ignored other times)

> We are worried that overruling the PMP settings is a potential security hole, and therefore prefer if the sentence is changed to explicitly make both interpretations a legal option (i.e. overruling the PMP or not overruling the PMP).

You propose to allow two options:
1. Accesses made from debug mode to the debug module bypass PMP checks.
2. Accesses made from debug mode to the debug module honor the PMP settings put in place by M mode.

[...]

PR #774, opened 2022-11-11
Address #771

pdonahue-ventana:
/cc @silabs-robin

### DIFF
--- a/implementations.tex
+++ b/implementations.tex
@@ -77,9 +77,11 @@ For additional flexibility, \RdmProgbufZero, etc. are mapped into regular memory
 immediately preceding \RdmDataZero, in order to form a contiguous region of memory which
 can be used for either program execution or data transfer.
 
-Note that for debug to always be possible, the PMP must not disallow fetches,
+The PMP must not disallow fetches,
 loads, or stores in the address range associated with the Debug Module when the
-hart is in Debug Mode, regardless of how the PMP is configured.
+hart is in Debug Mode, regardless of how the PMP is configured.  The same is
+true of PMA.  Without this guarantee, the park loop would enter an infinite
+loop of traps and debug would not be possible.
 
 \section{Debug Module Interface Signals} \label{dmi_signals}
 


timsifive (approved):

Silabs-ArjanB (commented):
It would be nice if it would be made clear that several implementations are possible. E.g. the burden of 'not disallowing fetches' could either be put on hardware (i.e. bypassing PMP for certain accesses) or on software (i.e. configuring PMP taking debug into account)

Silabs-ArjanB:
Bypassing the PMP (and possibly other checks) by hardware sounds like a security hole to me. If the PMP is always being bypassed (and accesses to the debug region always allowed), then software running on the hart itself has no way of prohibiting debug mode execution. (Of course the entire debug module can be made inaccessible, but that protection is at a different level in the design)

timsifive:
I'm not talking about bypassing the entire PMP. The requirement says:
> The PMP must not disallow fetches, loads, or stores in the address range associated with the Debug Module when the hart is in Debug Mode

Can you elaborate on the security hole you see with that? I don't exactly follow it.

Silabs-ArjanB:
Suppose that software specifically forbids execution based debug by configuring the PMP to disallow execution from the 'debug region'. If these PMP settings are overruled by allowing the CPU to execute from the debug region anyway, then I would consider that a security hole. Of course the adversary would need to have JTAG access allowed to the debug module as well (which can have its own protection mechanism), but disallowing it from software via the PMP would be a nice fail safe on top of that.

>> Having debug only work if the software plays along seems like a recipe for a poor debugging experience.

Another (non debug related) example: Suppose code execution is only allowed from some internal Flash. If you use the PMP you would have to set it up to allow code execution from Flash. Failing to do so would then also be judged a 'poor execution from Flash experience'... and if the same reasoning would apply then the PMP would need to be bypassed to always allow code execution from Flash ... this doesn't make sense in my opinion. People should just properly program the PMP.

timsifive:
> Suppose code execution is only allowed from some internal Flash. If you use the PMP you would have to set it up to allow code execution from Flash. Failing to do so would then also be judged a 'poor execution from Flash experience'

True, but the existence of an address range associated with a Debug Module is an implementation detail. There is no mechanism for software to discover this exists, or where it is. What's more, Debug Mode is essentially a mode that is more privileged than Machine Mode. It seems counterintuitive to have code in one mode prevent code in a more privileged mode from doing something.

Allowing software to prevent itself from being debugged externally is not a design goal. Maybe it should be, but that would have more impact beyond what we say in this appendix.

silabs-robin (approved):
Approving based on:
1. The changes to the text resolves the ambiguity.
2. The discussion seems to have converged.

PR #777, opened 2022-11-28
Improve Native Triggers formatting.

timsifive:
Move some text out of an explanation block. It's awkward to start a Section with one, and also the text doesn't make sense without reading the explanation.


pdonahue-ventana (approved):
That will look better.

Issue #778, opened 2022-11-29
Clarification on DPC bit 0


tovine:
I have a question about DPC, and more specifically bit 0 of this register.
In the spec it's listed as a DXLEN register, R/W all the way down to bit 0 while the _PC_ is not (at least it's impossible to have bit 0 of the PC set to 1 by any legal instruction).
![image](https://user-images.githubusercontent.com/8265757/204518648-40d5aa0c-0c99-4192-971c-0c54441387e7.png)

What is the expected behavior when writing a value to DPC where bit 0 is set, and what should then be the result of resuming execution in such a case?
I would assume that DPC should follow the PC behavior and in reality only be defined for [DXLEN-1:1] with bit 0 being hardwired 0, but what's your opinion, and could we get a clarification in the spec on this one?
The only other valid alternatives as I see it would be:
a) Bit 0 of DPC is ignored when resuming, so PC is set to DPC[DXLEN-1:1]
b) The hart raises a Instruction address misaligned exception (even if the C extension is implemented, which contradicts the note in the C spec that says this exception type is impossible when compressed instructions are supported)

pdonahue-ventana:
It should be the same as mepc/sepc/vsepc where the low bit is read-only zero (or the low two bits if the C extension isn't implemented).  I made a clarification in #779.

tovine:
Thanks, new spec looks clear. 🙂

PR #779, opened 2022-11-29
Low bit of dpc is read-only.

pdonahue-ventana:
Clarification for #778.

### DIFF
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -287,6 +287,10 @@
         \end{tabular}
         \end{table}
 
+        Like \Rmepc, the low bit of \RcsrDpc is read-only zero.  On
+        implementations that support only IALIGN=32, the two low bits are
+        read-only zero.
+
         When resuming, the hart's PC is updated to the virtual address stored in
         \RcsrDpc. A debugger may write \RcsrDpc to change where the hart resumes.
         <field name="dpc" bits="DXLEN-1:0" access="R/W" reset="0" />


timsifive (approved):
Should we mandate that debuggers write this bit as 0 as well? That leaves us the possibility in the future of using the low bit for some optional feature.

tovine:
> Should we mandate that debuggers write this bit as 0 as well? That leaves us the possibility in the future of using the low bit for some optional feature.

I think that seems like a good idea for a best practice at least.
Not sure what we can fit into that one bit but could potentially be useful for something and it's better to have it and not use it than to wish we had it later. 🙂

pdonahue-ventana:
I think that we should never use dpc[0] for something else.  If some future extension ever allows pc[0] to be non-zero (like ARM did with their [Jazelle](https://en.wikipedia.org/wiki/Jazelle) extension) then mepc[0] and dpc[0] would have to be redefined and we wouldn't want to have dpc[0] already being used for something else.

Even though mepc doesn't have a requirement for software to write mepc[0]=0, I guess that it wouldn't be harmful to be more restrictive than the priv spec and impose that requirement on debuggers.  Then we could remove the requirement if/when pc[0] can have either value (at the same time that we remove the statement about it being read-only zero).  I'm not sure if that adds a lot of value, though.  If a debugger is writing dpc[0]=1 then it's either writing all ones to determine how big the register is (and it doesn't really care about the low bits) or it's probably buggy (and it's not helpful to say "don't have a bug").

timsifive:
> I think that we should never use dpc[0] for something else. If some future extension ever allows pc[0] to be non-zero (like ARM did with their [Jazelle](https://en.wikipedia.org/wiki/Jazelle) extension) then mepc[0] and dpc[0] would have to be redefined and we wouldn't want to have dpc[0] already being used for something else.

Good point.

Can we say today that has the same read/write characteristics as mepc?
Then in a clarification we can mention that as of this writing that means that dpc[0] should be wired to 0.

pdonahue-ventana:
> Can we say today that has the same read/write characteristics as mepc?

What do you think of my latest change?  It previously didn't really say anything about the upper bits of dpc, either.  I think/hope that the intention was that dpc doesn't need to be able to hold all 2^64 addresses if some of them are invalid.

tovine:
I think it looks good 👍

timsifive (changes requested):
Can you remove the PDF from this PR?

Text change looks good.

pdonahue-ventana:
> Can you remove the PDF from this PR?

Oops.  I was careful to not have the PDF in the original PR but I blindly did `git commit -a` when I did the later change.  And (as usual) I had rebuilt the PDF to take a look at my changes so it picked that up.  I fixed that now.

timsifive (approved):

timsifive:
I'll merge this Monday. If you do it earlier, make sure to select squash-and-merge so the PDF change/revert gets squashed away.

PR #781, opened 2022-12-05
Remove longtable.sty.

timsifive:
Looks like Ubuntu jammy has the bugfix described in https://www.latex-project.org/cgi-bin/ltxbugs2html.new?pr=tools/3512


pdonahue-ventana (approved):
OK.  I'm wondering if I'll still be able to build locally (on an older Ubuntu release) when I want to eyeball my changes before I make a PR.  But that's my own problem and I can always keep this file locally if necessary.

timsifive:
You will be able to build without the file. What our copy of the file does is fix a bug which means some tables might run off the end of the page. So the PDF will be wonky in a few places, but there won't be any build errors.

PR #786, opened 2022-12-09
Exceptions may update PC register in debug mode

YenHaoChen:
A sentence in Section 4.1 Debug Mode is incorrect. 
"Exceptions don’t update any registers."
The ISA registers include PC, GPRs, FPRs, VPRs, and CSRs. Exceptions updates PC to debug trap vector in debug mode. Thus, the sentence should be:
"Exceptions don't update any CSRs."

*Note: exceptions never update GPRs, FPRs, and VPRs regardless of debug mode.

### DIFF
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -26,9 +26,9 @@ How Debug Mode is implemented is not specified here.
     Full permission checks, or a relaxed set of permission checks, will apply
     according to \FdmAbstractcsRelaxedpriv.
 \item All interrupts (including NMI) are masked.
-\item Exceptions don't update any registers.  That includes {\tt cause}, {\tt
-    epc}, {\tt tval}, {\tt dpc}, and \Rmstatus. They do end execution of the
-    Program Buffer.
+\item When an exception occurs, {\tt pc} may be updated, while all other registers
+    must not be changed. That includes {\tt cause}, {\tt epc}, {\tt tval}, {\tt dpc},
+    and \Rmstatus. Exceptions do end execution of the Program Buffer.
 \item No action is taken if a trigger matches.
 \item If \FcsrDcsrStopcount is 0 then counters continue. If it is 1 then
     counters are stopped.


timsifive:
Implementing Debug Mode as a trap to a debug ROM of some kind if implementation-specific, and not visible to the debugger. The debugger has no way to directly read the PC. It can only access DPC. (A debugger could use the program buffer, if implemented, and a jal instruction to figure out what the target hart thinks the PC is. I don't think we've addressed this corner case, but if we do then we'd say something like "the debugger shouldn't make any assumptions based on this".)

pdonahue-ventana:
We kind of addressed this corner case with:
> 9. All control transfer instructions may act as illegal instructions if their destination is in the
Program Buffer. If one such instruction acts as an illegal instruction, all such instructions
must act as illegal instructions.
> 10. All control transfer instructions may act as illegal instructions if their destination is outside
the Program Buffer. If one such instruction acts as an illegal instruction, all such instructions
must act as illegal instructions.
> 11. Instructions that depend on the value of the PC (e.g. auipc) may act as illegal instructions.

Based on the current spec, I think that portable/general debuggers shouldn't be reading PC (via jal, auipc, etc.) because that might or might not work.  But maybe we should put in a sentence that the PC is UNSPECIFIED in debug mode to more clearly dissuade debuggers from relying on it.

YenHaoChen:
Thank you for the clarification. Your explanation helps a lot.

I have an impression that a PC address is related to a single instruction. Thus, it is not obvious to me that exceptions don't update the PC but change the fetching instruction. My impression seems to be wrong in debug mode. Should We clarify this in debug spec?

Consider an implementation that allows PC reading in debug mode and makes the update by exceptions observable. Does this implementation still conform to the debug spec?

YenHaoChen:
> we'd say something like "the debugger shouldn't make any assumptions based on this".

> maybe we should put in a sentence that the PC is UNSPECIFIED in debug mode to more clearly dissuade debuggers from relying on it.

We have some suggestions to clarify the behavior. Is there any update on this issue?

timsifive:
> Based on the current spec, I think that portable/general debuggers shouldn't be reading PC (via jal, auipc, etc.) because that might or might not work. But maybe we should put in a sentence that the PC is UNSPECIFIED in debug mode to more clearly dissuade debuggers from relying on it.

But the spec already says:
> The Program Buffer may be implemented as RAM which is accessible to the hart. A debugger
> can determine if this is the case by executing small programs that attempt to write and read back
> relative to pc while executing from the Program Buffer. If so, the debugger has more flexibility in
> what it can do with the program buffer.

I feel that if a debugger wants to play games by reading the PC using a carefully crafted program buffer, then more power to it.

In that case, the list in Section 4.1 should read:
> 3. When an exception occurs pc may be updated, while all other registers must not be changed. That includes cause, epc, tval, dpc, and mstatus. Exceptions do end execution of the Program Buffer.

YenHaoChen:
Thanks for the reply. Your sentence clarifies my concern.

Sdext.tex
```diff
@@ -26,9 +26,9 @@ \section{Debug Mode} \label{debugmode}
     Full permission checks, or a relaxed set of permission checks, will apply
     according to \FdmAbstractcsRelaxedpriv.
 \item All interrupts (including NMI) are masked.
-\item Exceptions don't update any registers.  That includes {\tt cause}, {\tt
-    epc}, {\tt tval}, {\tt dpc}, and \Rmstatus. They do end execution of the
-    Program Buffer.
+\item When an exception occurs, pc may be updated, while all other registers must
```
timsifive, line 29:
```suggestion
\item When an exception occurs, {\tt pc} may be updated, while all other registers must
```
We're probably not consistent on this in the spec, but the ISA spec uses that formatting.

timsifive (approved):
This works for me, but I'd like to hear @pdonahue-ventana's opinion too.

timsifive (changes requested):
Can you remove the PDF from this PR? We try not to update it all the time because git isn't great at handling binary files.

timsifive (approved):
LGTM, but I'd still like @pdonahue-ventana to take a look before merging. He's out next week but I assume he'll get to it in January.

pdonahue-ventana (commented):
@timsifive, do you think that PC is architecturally UNSPECIFIED the entire time you're in debug mode?  That has been my assumption.  The execution-based scheme described in the appendix has PC point to a ROM.  Implementations that are like what I've seen in another architecture effectively freeze everything and spoonfeed instructions from the program buffer without a PC, so something like that might actually use the PC flops to hold DPC (if there's a single location for the PC).  PC becomes predictable again when it's set to dpc upon leaving debug mode.

The spec already has some hints that debuggers should avoid things like branches and auipc but it doesn't explicitly say that PC is UNSPECIFIED.  If we said that PC is UNSPECIFIED then this present situation is a subset and this change is unnecessary.

I have two concerns with this proposal:
1. I think that many hardware implementers confuse architecture and implementation.  They think that the appendix is normative and would be confused if the PC is allowed to change randomly rather than predictably sequencing through the ROM code (or in this case predictably going to a handler on an exception).
2. Existing debuggers might put loops in the program buffer if they know that the implementation allows that.  This might be the same problem where debugger writers confuse architecture and implementation.

timsifive:
Historically I've never thought of it that way. I figured if the implementation has a ROM and some RAM for program buffer, then the PC can look sane and if the debugger wants to take advantage, then let it.

I don't think that implementations that spoonfeed instructions without updating a PC are legal in any sane way. The spec says "The debugger can write whatever program it likes (including jumps out of the Program Buffer)" and I don't see how you make that work if you have some alternate instruction feeding mechanism.

I agree that some people think the appendix is normative. I'm not sure how to make that any more clear, but it feels like a separate problem.

pdonahue-ventana:
> I don't think that implementations that spoonfeed instructions without updating a PC are legal in any sane way.

The spec explicitly contemplates such an implementation:
> The slightly inconsistent behavior with a Program Buffer of size 1 is to accommodate hardware
designs that prefer to stuff instructions directly into the pipeline when halted, instead of having
the Program Buffer exist in the address space somewhere.

Such an implementation is very similar to the EDITR register on ARM.  Implementations I've seen basically have a mux that either takes instructions from the instruction cache or from the EDITR register depending on whether you're in Debug State.

The spec says "All control transfer instructions may act as illegal instructions if their destination is in the Program Buffer."  And the same with outside the PB.  Hardware is allowed to execute these branches or not, but I expect that debuggers are written for the least common denominator and would avoid branches in the PB.

YenHaoChen:
The spec aims to cover all ISA-level behaviors. Even some of the behaviors were unexpected to be used by software debuggers. For instance, raising exceptions in D-mode is unexpected in most cases, and we generally wish to keep flexibility in this particular case, i.e., implementation-specific and unspecified.

The question here is whether to allow an implementation changing PC upon exception in D-mode. The current sentence, "Exceptions don't update any registers (PC)," restricts all RISC-V implementations from doing so. Thus, the spec does not approve an implementation that changes the PC upon exception in D-mode.

@pdonahue-ventana mentioned there exist implementations that never make the PC visible to software debuggers, or implementations may even achieve a constant PC value in D-mode. Those implementations conform to the current spec. But still, do we have to make the spec restrict all implementations from updating PC upon exception in D-mode?

pdonahue-ventana:
I'm not saying that your proposal is wrong.  I'm saying that it might not go far enough.  I would say that PC can change at any time for any reason, not just in this specific situation.

YenHaoChen:
@pdonahue-ventana points out two issues of current spec.
1. Exceptions explicitly do not allow changing the PC in debug mode, which is undesired. 
2. The PC should be allowed to change anytime in debug mode, which the current spec does not qualify. For example, exceptions are not allowed to change PC.

The first point is a special case for the second one. Before dealing with the broader issue, we should clarify that the exception may change PC to avoid conflict. We all agree on the first point and have a concrete commit to address it. Should we incrementally resolve the second point in a different PR? Or should this PR also describe that PC can change anytime for any reason? (E.g., add another item in Section 4.1 Debug Mode "PC can change at any time for any reason.")

timsifive:
> The question here is whether to allow an implementation changing PC upon exception in D-mode. The current sentence, "Exceptions don't update any registers (PC)," restricts all RISC-V implementations from doing so. Thus, the spec does not approve an implementation that changes the PC upon exception in D-mode.

An exception in Debug Mode must not affect registers in a way that is noticeable outside of Debug Mode. That is what the original language was saying.

If there is something like a PC while executing a Program Buffer, then that PC might be updated when there is an exception, but that is not visible to the debugger in any way, so it doesn't matter.

Paul pointed out that we already allow there to not be something like a PC while executing the program buffer, citing points 9--11 from Section 4.1. But there might be, and a fancy debugger might take advantage of that. In practice this seems pretty unlikely.

So I think the existing language is correct, and this change isn't necessary.

timsifive (changes requested):
Taking back my approve.

YenHaoChen:
Thanks for the clarification

PR #789, opened 2022-12-22
Don't imply that matching and firing are the same.

pdonahue-ventana:
When reviewing #776, I noticed that this implied that matching and firing are the same.  Just like the other mechanism in section 5.4 prevents matching (rather than firing), I believe that this should prevent matching.

This text was originally added in #309 when tcontrol was added.  I don't think that the match/fire distinction was as crisp back then.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -182,11 +182,11 @@
             M-mode trigger enable field.
 
             <value v="0" name="disabled">
-                Triggers with action=0 do not match/fire while the hart is in M-mode.
+                Triggers with action=0 do not match while the hart is in M-mode.
             </value>
 
             <value v="1" name="enabled">
-                Triggers do match/fire while the hart is in M-mode.
+                Triggers do match while the hart is in M-mode.
             </value>
 
             When a breakpoint trap into M-mode is taken, \FcsrTcontrolMte is set to 0. When {\tt


timsifive:
I think "match/fire" is more clear than just "match". Matching is a prerequisite for firing, but that might not be common knowledge. If we just say "match," is somebody going to read that as "firing is OK?"

pdonahue-ventana:
Doesn't #776 make it clear what matching and firing are?

Corner case that I just thought of as I made this comment: What if you're in M mode with tcontrol.mte=0 and then software writes icount.pending=1?  There was no match but the icount will fire before the next instruction under my proposal.  I don't know if we really care about this corner case but maybe the existing text is actually the best behavior (don't fire until either mte=1 or you leave M mode). Thoughts?

timsifive:
> Doesn't #776 make it clear what matching and firing are?

It does.

> Corner case that I just thought of as I made this comment: What if you're in M mode with tcontrol.mte=0 and then software writes icount.pending=1? There was no match but the icount will fire before the next instruction under my proposal. I don't know if we really care about this corner case but maybe the existing text is actually the best behavior (don't fire until either mte=1 or you leave M mode). Thoughts?

tcontrol is to help implementations that want to use mcontrol/mcontrol6 on systems where there is only one privilege mode. Do we want to help systems use icount where there is only one privilege mode? That seems reasonable, but if so we should fix that by explicitly mentioning icount in tcontrol instead of relying on some vague understanding of fire/match. (I might argue that icount matches on every instruction which causes a decrement, so it doesn't match depending on tcontrol settings, and therefore there is no problem.)

pdonahue-ventana:
> tcontrol is to help implementations that want to use mcontrol/mcontrol6 on systems where there is only one privilege mode. Do we want to help systems use icount where there is only one privilege mode? That seems reasonable, but if so we should fix that by explicitly mentioning icount in tcontrol instead of relying on some vague understanding of fire/match.

I don't think that it's vague.

> (I might argue that icount matches on every instruction which causes a decrement, so it doesn't match depending on tcontrol settings, and therefore there is no problem.)

This is already explicit.  The definition of icount includes: "This trigger matches on every instruction completed or trap taken from a privilege mode where the trigger is enabled."  And tcontrol.mte=0 suppresses matching (and firing) in M mode.  So I agree that there is no problem.  But I also don't think that any clarification is needed.

With the existing language, when tcontrol.mte=0:
- matching (decrementing icount.count) is suppressed
- firing (trapping when icount.pending=1) is suppressed

I think that all makes sense and this PR isn't necessary.  This PR would have (unfortunately) allowed this situation:
- icount.count=1 and icount.m=1
- execute something that hits a PC breakpoint which fires
- breakpoint trap to M mode
- breakpoint trap causes icount.count to decrement, sets icount.pending, and clears tcontrol.mte
- we're in M mode with icount.pending=1. Matching is suppressed.  But if firing is not suppressed (as I had proposed) then we'll take another breakpoint trap (because "When pending is set, the trigger fires just before any further instructions are executed in a mode where the trigger is enabled") which destroys mepc and defeats the purpose of this mechanism.

So let's keep the status quo which is clearly better.

PR #795, opened 2023-01-23
Expand mcontrol6.hit into hit0 and hit1, and remove mcontrol6.timing.

timsifive:
This lets an implementation be more specific about when a trigger fired in relation to the instruction that matched, to avoid user confusion when triggers fire late.

It also simplifies the spec around desired trigger timing behavior, simply asking for best effort to match the recommendation from the table.

### DIFF
--- a/introduction.tex
+++ b/introduction.tex
@@ -124,6 +124,10 @@ order to implement 1.0:}
     can be written with any value supported by any of the types this trigger
     supports. \PR{721}
     \item \RcsrTcontrol fields only apply to breakpoint traps, not any trap. \PR{723}
+    \item \FcsrMcontrolSixHitZero (previously called $|hit|$) now contains 0 when a
+        trigger fires more than one instruction after the instruction that
+        matched.  (This information is now reflected in \FcsrMcontrolSixHitOne.)
+        \PR{795}
 \end{steps}
 
 \subsubsection{Minor Changes from 0.13 to 1.0}
--- a/introduction.tex
+++ b/introduction.tex
@@ -124,10 +124,11 @@ order to implement 1.0:}
     can be written with any value supported by any of the types this trigger
     supports. \PR{721}
     \item \RcsrTcontrol fields only apply to breakpoint traps, not any trap. \PR{723}
-    \item \FcsrMcontrolSixHitZero (previously called $|hit|$) now contains 0 when a
+    \item \FcsrMcontrolSixHitZero (previously called \RcsrMcontrolSix.$|hit|$) now contains 0 when a
         trigger fires more than one instruction after the instruction that
         matched.  (This information is now reflected in \FcsrMcontrolSixHitOne.)
         \PR{795}
+    \item \RcsrMcontrolSix.$|timing|$ is now read only 0.
 \end{steps}
 
 \subsubsection{Minor Changes from 0.13 to 1.0}
@@ -155,6 +156,8 @@ incompatible, but unlikely to be noticeable:}
     \item Solutions to deal with reentrancy in Section~\ref{sec:nativetrigger}
         prevent triggers from {\em matching}, not merely {\em firing}. This primarily
         affects \RcsrIcount behavior. \PR{722}
+    \item The timing field in \RcsrMcontrolSix is now tied to 0. It was removed in
+        favor of \FcsrMcontrolSixHitZero and \FcsrMcontrolSixHitOne. \PR{795}
 \end{steps}
 
 \subsubsection{New Features from 0.13 to 1.0}
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -365,8 +365,8 @@
         <field name="timing" bits="18" access="WARL" reset="0">
             <value v="0" name="before">
             The action for this trigger will be taken just before the
-            instruction that triggered it is committed, but after all preceding
-            instructions are committed. \Rxepc or \RcsrDpc (depending
+            instruction that triggered it is retired, but after all preceding
+            instructions are retired. \Rxepc or \RcsrDpc (depending
             on \FcsrMcontrolAction) must be set to the virtual address of the
             instruction that matched.
 
@@ -380,8 +380,8 @@
 
             <value v="1" name="after">
             The action for this trigger will be taken after the instruction
-            that triggered it is committed. It should be taken before the next
-            instruction is committed, but it is better to implement triggers imprecisely
+            that triggered it is retired. It should be taken before the next
+            instruction is retired, but it is better to implement triggers imprecisely
             than to not implement them at all.  \Rxepc or
             \RcsrDpc (depending on \FcsrMcontrolAction) must be set to
             the virtual address of the next instruction that must be executed to
@@ -726,7 +726,7 @@
 
             <value v="1" name="before">
             The trigger fired just before the instruction that triggered it was
-            committed, but after all preceding instructions are committed.
+            retired, but after all preceding instructions are retired.
             \Rxepc or \RcsrDpc (depending on \FcsrMcontrolSixAction) must be set
             to the virtual address of the instruction that matched.
 
@@ -738,7 +738,7 @@
 
             <value v="2" name="after">
             The trigger fired after the instruction that triggered and at least
-            one additional instruction were committed.
+            one additional instruction were retired.
             \Rxepc or \RcsrDpc (depending on \FcsrMcontrolSixAction) must be set
             to the virtual address of the next instruction that must be executed
             to preserve the program flow.
@@ -746,7 +746,7 @@
 
             <value v="3" name="immediately after">
             The trigger fired just after the instruction that triggered it was
-            committed, but before any subsequent instructions were executed.
+            retired, but before any subsequent instructions were executed.
             \Rxepc or \RcsrDpc (depending on \FcsrMcontrolSixAction) must be set
             to the virtual address of the next instruction that must be executed
             to preserve the program flow.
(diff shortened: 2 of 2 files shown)


pdonahue-ventana:
I haven't thought through all the implications but here are some initial questions/comments:
1. How does this work with respect to action>1?  For instance, if you have a load address trigger which turns trace on, do these bits get set based on whether the first traced instruction is the load itself, the instruction immediately after the load, or some asynchronous trace start event later?

2. If an implementation sometimes stops immediately after and sometimes later (e.g. depending on whether there are pipeline bubbles), is it legal for it to always call it the generic/vague "after" instead of trying to categorize as "more than one instruction after" vs. "the first instruction after"?  It seems like it might be useful to loosen "more than one instruction after" to be "one or more" and then say that you should say "the first instruction after" if possible, but it's OK to say "one or more" if you want.

3. I think that this allows implementations that only support subsets like:
   - {false, before} (backwards compatible which is allowed because you say that hit1 can be 0)
   - {false, after} (definitely allowed because hit0 can be hardwired to 0)
   - {false, immediately after} (hit0 and hit1 read the same underlying bit) - is this allowed?
That would be visible if somebody tried to write one of the other values to these WARL fields.

4. I think that "right before" and "immediately after" should be defined.  *epc or dpc points to the instruction that hit the trigger or the next instruction after.  There was previous verbiage that could be reused.

5. This loses the warning about load value triggers with timing=before and load side-effects.  While debuggers would no longer need to be warned about asking for that, there still should be warnings so that hardware people can know that this is a bad idea and debuggers can know that they need to be careful when dealing with hardware that does this.

6. Are there any implications for chaining?  It might actually make it easier but we should think this through.

This is a nontrivial change so I still want to think about it.

YenHaoChen:
I like the idea of this PR. We generally prefer triggers fire before the instruction to give the user more insight, and we allow firing after for cases where the triggers cannot fire before. Additionally, the debugger can understand whether the instruction is executed by reading the tdata1 CSR.

I can provide a PR of hit0 and hit1 for Spike after approval of this PR.

(minor typo: "It they are implemented..." -> "If they are implemented...")

timsifive:
> How does this work with respect to action>1? For instance, if you have a load address trigger which turns trace on, do these bits get set based on whether the first traced instruction is the load itself, the instruction immediately after the load, or some asynchronous trace start event later?

I would expect that in a given implementation the trigger fires when it fires, and the timing is suggested by the table. I suppose an implementation could fire at a different time depending on the action, but that seems like added complexity with no benefit.

> If an implementation sometimes stops immediately after and sometimes later , is it legal for it to always call it the generic/vague "after" instead of trying to categorize as "more than one instruction after" vs. "the first instruction after"?

I would like implementation to be clear on whether the trigger hit immediately after, or later. That seems the most useful to a user. Otherwise everyone will just report the vague after and the distinction is lost. Is it onerous to have to report the distinction?

> (hit0 and hit1 read the same underlying bit) - is this allowed? That would be visible if somebody tried to write one of the other values to these WARL fields.

hit0/hit1 are WARL, so I think that is allowed.

> I think that "right before" and "immediately after" should be defined. *epc or dpc points to the instruction that hit the trigger or the next instruction after. There was previous verbiage that could be reused.

Done.

> This loses the warning about load value triggers with timing=before and load side-effects.

It's back.

> Are there any implications for chaining? It might actually make it easier but we should think this through.

I think it makes things a little more fuzzy, but ultimately probably simpler. Now it seems possible that someone decides to implement chains of instruction address before + load value after, which previously were prohibited (because timing didn't match).

What happens when a debugger sets a chain that can never fire because of hardware limitations? Or that can fire on some instructions but not on others?

timsifive:
@pdonahue-ventana What do you need to feel good about this PR? More discussion in the mailing list/meeting? I want to get this going, since it affects some of e.g. the vector discussion.

introduction.tex
```diff
@@ -124,6 +124,9 @@ \subsubsection{Incompatible Changes from 0.13 to 1.0}
     can be written with any value supported by any of the types this trigger
     supports. \PR{721}
     \item \RcsrTcontrol fields only apply to breakpoint traps, not any trap. \PR{723}
+    \item \FcsrMcontrolSixHitZero now contains 0 when a trigger fires more than one instruction
+        after the instruction that matched. (This information is now reflected
+        in \FcsrMcontrolSixHitOne.) \PR{795}
```
pdonahue-ventana, line 129: This should be prefaced by the fact that hit is renamed to FcsrMcontrolSixHitZero.

xml/hwbp_registers.xml
```diff
@@ -707,13 +712,45 @@
...
-            time. It is used to determine which
-            trigger(s) matched.  If the bit is not implemented, it is always 0
-            and writing it has no effect.
+        <field name="hit0" bits="22" access="WARL" reset="0">
+            It they are implemented, \FcsrMcontrolSixHitOne (MSB) and
+            \FcsrMcontrolSixHitZero (LSB) combine into a single 2-bit field that
+            indicates if and when the trigger was hit.
+
+            If either of the bits is not implemented, the unimplemented bits
+            will be wired to 0.
```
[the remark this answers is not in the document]
timsifive: Fixed.

xml/hwbp_registers.xml
```diff
@@ -707,13 +712,45 @@
...
+            <value v="1" name="before">
+            The trigger fired just before the instruction that triggered it was
+            committed, but after all preceding instructions are committed.
+            \Rxepc or \RcsrDpc (depending on \FcsrMcontrolSixAction) must be set
+            to the virtual address of the instruction that matched.
+
+            If \FcsrMcontrolSixLoad is set and \FcsrMcontrolSixSelect=1 then a
+            memory access has been performed (including any side effects of
+            performing such an access) even though the load has not updated its
+            destination register.
```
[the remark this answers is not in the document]
timsifive: That's better than what I had.

xml/hwbp_registers.xml
```diff
@@ -707,13 +712,45 @@
...
+            </value>
+
+            <value v="2" name="after">
+            The trigger fired after the instruction that triggered and at least
+            one additional instruction were committed.
```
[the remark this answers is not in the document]
timsifive: Yeah, committed is what we had previously, and still have for mcontrol. Should I replace that all with retired?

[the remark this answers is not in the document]
timsifive: Done.

introduction.tex
timsifive: That's not a backwards incompatible change, though. What matters for compatibility is the location of the bit and what values it takes when.

pdonahue-ventana:
I'm OK with this in principle.  Just wordsmithing.

xml/hwbp_registers.xml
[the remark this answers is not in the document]
pdonahue-ventana: I'd say retired instead of committed everywhere.  It's not a huge problem but it just seems more consistent with the priv spec.

introduction.tex
pdonahue-ventana: Someone here read this and was concerned about the word "now" which implies that there previously was a hit0 field and the behavior of that field is "now" different.  I was trying to address that confusion.

pdonahue-ventana: Now I understand your point.  It has to do with the way that encoding 2 would be interpreted by old debuggers.  I still think that some minor explanation of hit vs. hit0 is useful to people who may be scratching their head about why the hit0 behavior changed when there was no hit0 in the old spec.

```
    \item The hit bit is now called \FcsrMcontrolSixHitZero and it
        contains 0 when a trigger fires more than one instruction
        after the instruction that matched. (This information is now reflected
        in \FcsrMcontrolSixHitOne.) \PR{795}
```

pdonahue-ventana: Or
```
    \item \FcsrMcontrolSixHitZero (called "hit" in 0.13) now contains 0 when a trigger fires more than one instruction
        after the instruction that matched. (This information is now reflected
        in \FcsrMcontrolSixHitOne.) \PR{795}
```

timsifive: I added the 2nd version.

pdonahue-ventana (approved):
Perfect!

timsifive:
> Perfect!

Not quite. I just pushed some more tweaks.

pdonahue-ventana:
The latest version still looks good.

JamesKenneyImperas:
I haven't followed all the details here, but I'm wondering whether the same effect here can be achieved without introducing a backwards incompatibility with earlier `mcontrol6.hit` behavior.

The currently-specified behavior is:
```
hit0:0 hit1:0 -> not hit
hit0:1 hit1:0 -> hit before
hit0:1 hit1:1 -> hit immediately after
hit0:1 hit1:1 -> hit after, not immediately
```

I think that the before/after encoding here is redundant, because that is already reported by the `timing` field in the trigger. Therefore, an alternative implementation could be to preserve the original `hit` bit and rename the `hit1` bit to `imprecise` (or something) which would get set only for a trigger with timing "after" that did not occur immediately. This would give these encodings:

```
hit:0 -> not hit
hit:1 timing:0 -> hit before
hit:1 timing:1 imprecise:0 -> hit immediately after
hit:1 timing:1 imprecise:1 -> hit after, not immediately
```
 So the same information can be derived without backwards-incompatibility and with consistent meaning of "hit" in all implementations, whether or not they support the `imprecise`/`hit1` bit.

Thanks.

JamesKenneyImperas:
Ok, I think I see: the timing field is only a recommendation, not a requirement, so cannot be relied upon to determine actual timing.

timsifive:
> Ok, I think I see: the timing field is only a recommendation, not a requirement, so cannot be relied upon to determine actual timing.

And also I hope to remove the timing bit in #807.

PR #797, opened 2023-01-27
Clarify icount decrements on instruction execution

timsifive:
Previously we said "completes" which is not well defined. Also be clear that xRET instructions count.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -980,8 +980,16 @@
     <register name="Instruction Count" short="icount" address="0x7a1">
         This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 3.
 
-        This trigger matches on every instruction completed or trap taken from a
-        privilege mode where the trigger is enabled.
+        \begin{steps}{This trigger matches when:}
+            \item
+                An instruction retires after having been fetched in a privilege
+                mode where the trigger is enabled.
+                This explicitly includes {\em x}\/RET instructions.
+            \item A trap is taken from a privilege mode where the trigger is enabled.
+        \end{steps}
+
+        If more than one of the above events occur during a single instruction
+        execution, the trigger still only matches once for that instruction.
 
         When \FcsrIcountCount is greater than 1 and the trigger matches, then
         \FcsrIcountCount is decremented by 1. (This is true even if the action configured in


pdonahue-ventana (commented):
I was going to say that "executed" is vague because a load that gets a page fault might be considered by some to have executed (but not retired).  But that instruction traps and falls into the second bullet anyway.  As long as we think that it's clear that this situation would not cause the count to decrement twice: once for execution and once for the trap.  Or maybe people think that "executed" cannot possibly include this trapping load situation?

timsifive:
I've rephrased this using "retires", while (hopefully) making it clear that for the mode it matters which mode the hart was in when the instruction was fetched. Specifically for xRET

xml/hwbp_registers.xml
```diff
@@ -980,8 +980,12 @@
     <register name="Instruction Count" short="icount" address="0x7a1">
         This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 3.
 
-        This trigger matches on every instruction completed or trap taken from a
-        privilege mode where the trigger is enabled.
+        \begin{steps}{This trigger matches when:}
+            \item An instruction is fetched in a privilege mode where the
+                trigger is enabled and then retires.
+                This explicitly includes {\em x}\/RET instructions.
```
pdonahue-ventana, line 986: I find the wording awkward (particularly the "and then retires" being split from "fetched" by the intervening phrase).  Is there a way to say that it retires in an enabled mode (leaving out fetch) and then the second sentence says that in this context xRET is considered to retire in the mode it's coming from (or fetched in) rather than the mode it's going to?  Or even "An instruction retires after having been fetched in a privilege mode where the trigger is enabled" would be better in my view.

The content is good.  Just wordsmithing.

timsifive: I like your second wording, because it makes the line about xRET be simply a clarification and not a special case. If there are/will be other instructions that change the privilege mode we want them to also be included.

xml/hwbp_registers.xml
```diff
@@ -980,8 +980,13 @@
     <register name="Instruction Count" short="icount" address="0x7a1">
         This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 3.
 
-        This trigger matches on every instruction completed or trap taken from a
-        privilege mode where the trigger is enabled.
+        \begin{steps}{This trigger matches at most once per instruction when:}
```
pdonahue-ventana, line 983: Now I'm happy with the bullets but I'm concerned that "at most once per instruction" sounds like it's optional whether it matches at all.  (Maybe an implementation could decide to match zero times per instruction retire in some situations.)  Maybe something like "Unless otherwise disabled/prohibited/specified, this trigger matches when..."

timsifive: What I want to underline is that the trigger matches once even if somehow both of these conditions are met by the same instruction. It doesn't match twice in that case.

pdonahue-ventana (approved):
👍

PR #798, opened 2023-01-27
Suggest debugger behavior when icount stepping over restarting instruction

timsifive:
Tons of mailing list discussion, e.g. https://lists.riscv.org/g/tech-debug/message/1016

### DIFF
--- a/debugger_implementation.tex
+++ b/debugger_implementation.tex
@@ -606,7 +606,7 @@ the OS or debug stub writes \FcsrIcountCount=1, \FcsrIcountAction=0,
 \FcsrIcountM=0 before returning control to the lower user program with an
 {\tt mret} instruction.
 
-On tiny systems which only supports M-Mode single step is doable, but tricky to
+On tiny systems which only support M-Mode single step is doable, but tricky to
 get right. To single step, the debug stub would execute something like:
 \begin{verbatim}
     li    t0, \FcsrIcountCount=4, \FcsrIcountAction=0, \FcsrIcountM=1
--- a/debugger_implementation.tex
+++ b/debugger_implementation.tex
@@ -615,3 +615,30 @@ get right. To single step, the debug stub would execute something like:
     lw    sp, 0(sp)     /* Restore sp, count decrements to 2 */
     mret                /* Return to program being debugged. count decrements to 1 */
 \end{verbatim}
+
+There is an additional problem with using \RcsrIcount to single step.  An
+instruction may cause an exception into a more privileged mode where the trigger
+is not enabled.  The exception handler might address the cause of the exception,
+and then restart the instruction.  Examples of this include page faults, FPU
+instructions when the FPU is not yet enabled, and interrupts.
+When a user is single stepping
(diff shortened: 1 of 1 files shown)


debugger_implementation.tex
```diff
@@ -615,3 +615,32 @@ \subsection{Single Step} \label{nativestep}
...
+
+To help users out, debuggers should detect when a single step restarted an
+instruction, and then step again. This way the users see the expected behavior
+of stepping over the instruction. Ideally the debugger would notify the user
+that an exception handler executed the first time.
+
+\begin{steps}{The debugger should perform this extra step if all the following
+        conditions are satisfied:}
+    \item The PC did not change during the first step.
+    \item The PC does not equal the {\tt epc} register for the current privilege mode.
```
pdonahue-ventana, line 636: This condition isn't immediately obvious to me.  If you're stepping S mode, PC=X, you step, and PC is still X then step one more time.  Why do we also need to check if sepc=X?

timsifive: That should be `tvec`, not `epc`.

The condition I'm worried about is if you're stepping the first instruction in an exception handler, that itself causes that exception handler to be called. But maybe that doesn't matter similar to jal to self makes no changes after the first invocation.

I'd love some more input into my thinking here.

pdonahue-ventana: tvec makes more sense.  Say that you're stepping U+S modes and breakpoints go to M.  The following happens:
- You step X and it traps.
- The trap sends you to M mode with mepc=X.
- PC did change so you wait for the debugger user to request another step.  (Don't automatically do the extra step you're describing.)
- User says to step.  Your vector is an illegal instruction (or whatever) so it traps.
- The trap sends you to M mode with mepc=stvec.

You're saying that the debugger should not do the extra step here.  I agree that you don't want to do the extra step.  However, if you did then you'd hit the same illegal instruction and come back to M mode with mepc=stvec so there's no real harm.

pdonahue-ventana:
Modulo the tvec details, I think that this is generally good.  If you get a page fault on a load, the debugger should step twice because the PC didn't advance.  If you get an interrupt on a nop, the debugger should also step twice because the PC didn't advance.  If you get an illegal instruction on something that's emulated, the debugger will naturally only step once because the PC did advance.  No handler needs to do any fixup and all the work is pushed to the debugger which presumably isn't as performance-critical as one of the handlers.

timsifive:
I think we might not need the `tvec` rule at all. Then all the debugger needs to check is whether the PC changed. This works, iff the following statement is true: Every RISC-V instruction changes the PC or has no side effects when repeated.

Is that a true statement?

pdonahue-ventana:
This isn't just caused by exceptions like page faults.  Interrupts also can cause this behavior on otherwise innocuous instructions.  The solution is the same, but I would mention both exceptions and interrupts in the description just so everyone is aware of the full spectrum of cases.

timsifive:
I've added interrupts as another source of this problem.

pdonahue-ventana (approved):
Looks good.

PR #800, opened 2023-02-06
Add mcontrol6.{uncertain,uncertainen} bits.

timsifive:
Uncertain trigger matches occur when the hardware can't do a precise comparison, e.g. on batched vector accesses, and far AMO data.

This lets the user select whether they want the trigger to fire if a match is uncertain, and the hardware can indicate whether a given match was certain or not.

Lots of mailing list discussion prompted this:
https://lists.riscv.org/g/tech-debug/topic/94864865#1030

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -700,9 +700,36 @@
         If \RcsrTextraThirtytwo or \RcsrTextraSixtyfour are implemented for this
         trigger, it only matches when the conditions set there are satisfied.
 
+        \begin{commentary}
+        \FcsrMcontrolSixUncertain and \FcsrMcontrolSixUncertainen exist to
+        accommodate systems where not every memory access is fully observed by
+        the Trigger Module. Possible examples include data values in far AMOs,
+        and the address/data/size of accesses by instructions that perform
+        multiple memory accesses, such as vector, push, and pop instructions.
+
+        While the uncertain mechanism exists to deal with these situations, it
+        can lead to an unusable number of false positives. Users will get a much
+        better debug experience if the TM does have perfect visibility into the
+        details of every memory access.
+        \end{commentary}
+
         <field name="type" bits="XLEN-1:XLEN-4" access="R" reset="6" />
         <field name="dmode" bits="XLEN-5" access="WARL" reset="0" />
-        <field name="0" bits="XLEN-6:26" access="R" reset="0" />
+        <field name="0" bits="XLEN-6:27" access="R" reset="0" />
+        <field name="uncertain" bits="26" access="WARL" reset="0">
+            This bit should be updated every time that \FcsrMcontrolSixHitZero
+            or \FcsrMcontrolSixHitOne is updated.
+
+            <value v="0" name="certain">
+                The trigger that fired satisfied the configured conditions, or
+                this bit is not implemented.
+            </value>
+            <value v="1" name="uncertain">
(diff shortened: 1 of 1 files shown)


sifiverobert:
As these bits (applicable to at least both far-AMO and vectors) are generic I like them.

I would maybe specifically mention these two use-cases (as examples ...). These are mentioned in commit comments, but not in text itself. Also phrase 'in a situation' is a bit confusing (but I a not native English speaker ...).

I also assume (maybe good to say it directly ...), that when 'uncertainen' is set to 0 (default for existing tools), trigger must ONLY fire as defined (including all before/after/size etc.). Only when tool (or user ...) wants uncertain behavior they should explicitly allow this. Is that right?

timsifive:
> I would maybe specifically mention these two use-cases (as examples ...). These are mentioned in commit comments, but not in text itself.

I'm adding a commentary that mentions the use cases.

> Also phrase 'in a situation' is a bit confusing (but I a not native English speaker ...).

Agreed, but I couldn't think of anything better. :-(

> I also assume (maybe good to say it directly ...), that when 'uncertainen' is set to 0 (default for existing tools), trigger must ONLY fire as defined (including all before/after/size etc.). Only when tool (or user ...) wants uncertain behavior they should explicitly allow this. Is that right?

That is definitely right. And I think that's what the text says, but I'll try to make it more clear.

bcstrongx (commented):
Is uncertain read-only (access="R")?  That will cause problems with context switching of trigger regs.  All implemented bits should be writable at least when the trigger is not enabled, but arguably anytime (with the understanding that writing mcontrol6 for an enabled trigger could cause race conditions and potentially lost hit/uncertain updates).

timsifive:
> Is uncertain read-only (access="R")?

Good catch. I made it WARL.

xml/hwbp_registers.xml
```diff
@@ -700,9 +700,32 @@
         If \RcsrTextraThirtytwo or \RcsrTextraSixtyfour are implemented for this
         trigger, it only matches when the conditions set there are satisfied.
 
+        \begin{commentary}
+        \FcsrMcontrolSixUncertain and \FcsrMcontrolSixUncertainen exist to
+        accommodate system where not every memory access is fully observed by
```
pdonahue-ventana, line 705: "systems"

pdonahue-ventana (commented):
If you don't implement the optional hit0/hit1 bits then presumably uncertain is also not implemented, right?  But uncertainen could still be implemented because you might want to limit the uncertainty, right?

For instance, a competing architecture doesn't have hit bits.  I'm not sure how it handles vector loads/stores but let's assume it allows uncertainty.  The uncertain bit is useless without hit bits because the uncertain bit is related to a specific trigger and you don't know which trigger to look at (since there's no hit information).  But uncertainen still does still seem useful because debuggers may want to make different tradeoffs with false positives vs. false negatives.  (Such debuggers presumably do some software matching of the instruction in question against all triggers.)

timsifive:
I think uncertain is useful without the hit bits, because it lets the debugger select desired behavior.

Not implementing hit bits seems like a poor choice to me. Can we require them since we bumped tinfo.version?

pdonahue-ventana:
Let's go with this as-is.  I'm not a big fan of this type of uncertainty but it seems unavoidable (unless we eliminate other features or lock things down in ways that would be controversial).

pdonahue-ventana (approved):

PR #801, opened 2023-02-10
Debuggers should accept mcontrol6.20 reading back 1.

timsifive:
This bit used to be used to request/acknowledge trigger timing information. This change ensures triggers that implements the spec before #795 will still work.


timsifive:
#807 solves this problem in a better way.

Issue #796, opened 2023-01-26
Priority of previous-instruction debug exceptions and interrupts


JamesKenneyImperas:
Section 5.3 (table 5.2) indicates the priority of synchronous exceptions with respect to trigger-initiated debug exceptions.

Should the priority with respect to _pending interrupts_ also be defined? For example, if there is a pending trigger-initiated debug exception after the previous instruction (the first row of table 5.2) _and_ a pending enabled interrupt, which is taken? A specific case of this is when the just-executed instruction has enabled interrupts.

Thanks.

timsifive:
I think pending interrupts should be below debug interrupts. But it doesn't really matter since the timing of external interrupts is such that they don't come in when a particular instruction is being executed.

If you feel this really matters, can you make a PR that makes the spec more complete/readable?

JamesKenneyImperas:
Hi Tim,

There is this statement in the Privileged Specification about interrupts (section 3.1.9):

> These conditions for an interrupt trap to occur must be evaluated in a bounded amount of time from when an interrupt becomes, or ceases to be, pending in mip, **and must also be evaluated immediately following the execution of an xRET instruction or an explicit write to a CSR on which these interrupt trap conditions expressly depend (including mip, mie, mstatus, and mideleg)**.

This means that already-pending interrupts are treated as _synchronous_ traps when enabled because of an explicit CSR change of some kind, so I think it is worth including them in priority tables. I also feel that they should be lower priority than debug interrupts. I'll have a go at making a pull request.

James.

timsifive:
I missed that text. It's too bad they didn't include external interrupts it in the priority table in the same spec. Thanks for taking this on.

JamesKenneyImperas:
Thinking further before attempting any documentation update: the statement above implies interrupts are similar to debug events on the previous instruction with timing "after". In the absence of the debugger, the interrupt will be taken without performing any actions associated with the next instruction. Specifically, interrupts have higher priority than any page, permission, or illegal instruction fault on the next instruction.

I think this means they must be handled _after_ the "mcontrol/mcontrol6 after" stage in table 5.2 but _before_ any following stages. If the debugger is active, I think a step instruction breakpoint on the just-completed instruction should have higher priority than an interrupt on the next instruction. The various debug halt signals (`haltreq` etc) should also be higher priority than an interrupt.

PR #802, opened 2023-02-13
tentative attempt to fix issue 796

JamesKenneyImperas:
Hi @timsifive  - I've made an attempt at this (reference #796)


timsifive (changes requested):
The spec changes look right to me. I have two requests for the PR:
1. Can you give it a more descriptive title? Something like "define asynchronous interrupt priorities" or anything where someone reading just the title could guess what the change relates to.
2. Don't include the PDF in the PR. It just leads to conflicts and a lot of churn.

timsifive (approved):
Looks good, but I'd like @pdonahue-ventana's opinion too. He has a better head for stuff like this.

pdonahue-ventana:
This table in the debug spec mirrors [table 3.7](https://github.com/riscv/riscv-isa-manual/blob/b39b1bec376ee661bfb05eebe4e7ea781fe4d803/src/machine.tex#L2128) of the [latest priv spec](https://github.com/riscv/riscv-isa-manual/releases/download/draft-20230131-c0b298a/riscv-privileged.pdf) which doesn't have interrupts.  This PR specifies the priority of interrupts with respect to non-debug exceptions, not just debug exceptions.  I don't think that the priv spec defines that interrupt vs. synchronous exception priority and it's not appropriate for the debug spec to define that.

JamesKenneyImperas:
Hi @pdonahue-ventana,

Table 3.7 doesn't mention interrupts because they are not synchronous traps _on the current instruction_ resulting from execution of that instruction. The behaviour of interrupts in the case we are considering here (enabling a pending interrupt by execution of `xret` or explicit CSR write) is explicitly covered by paragraph 3.1.9 (see issue #796). The Privileged Specification is clear that enabling a pending interrupt in this way results in the interrupt occurring at higher priority than anything else before the next instruction is executed. In debug terminology, it is effectively a high priority synchronous event with timing 'after' the current instruction.

The issue is that the Debug Specification introduces _other_ high priority synchronous events with timing 'after' (e.g. stepi trigger events), and therefore needs to define the priority of interrupts with respect to those. For example, if you stepi over an `xret` that enables interrupts, is it the interrupt that happens next or the stepi breakpoint?

I agree that it feels a bit awkward inserting interrupts into the table as I have tried to do; can you think of a better way to make clear where interrupts sit?

Thanks,

James.

pdonahue-ventana:
I see your point.  Let me think about it.

timsifive:
Is this a bugfix, clarification, or something else?

JamesKenneyImperas:
It's an attempt to specify something that is missing entirely at the moment - see our previous discussion on #796. Not sure whether that counts as a bugfix or clarification.

YenHaoChen:
This PR clarifies the priority between precise/synchronous interrupts and triggers with timing=after. The current Spike does not provide this behavior. Is there any update on this clarification?

PR #806, opened 2023-02-16
Reinstate timing

timsifive:
I still think this should happen, but it needs to be backwards compatible. Discussion on how to do this is ongoing, and I don't want someone to grab the current spec and then not implement timing at all.

Rebuild the PDF as well, to avoid people who just grab the PDF (which is probably everyone) from seeing the incomplete work.


timsifive:
@pdonahue-ventana Can you prioritize this review? I don't like the fact that the current git mainline points to something that we're definitely not going to do that way.

timsifive:
Force merging this because people are already implementing the partially baked feature. https://github.com/riscv-software-src/riscv-isa-sim/pull/1257

pdonahue-ventana:
Sorry.  I was out yesterday (holiday) and then I'm reading emails out of order.  Reverting this was fine.

PR #807, opened 2023-02-17
Remove timing bit again. Introduce tinfo.version.

timsifive:
Lots of email discussion, subject:"removed mcontrol6.timing,
compatibility"

Now that hit0/hit1 clearly describe when a trigger hit, we don't need a
separate mechanism to communicate whether triggers hit before/after. The
only remaining benefit timing provided was that it would let the
debugger know when the trigger is set what the timing behavior would
probably be. (Emphasis on probably, since it's not possible to honor the
requested timing in all situations.)

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -164,6 +164,8 @@
                     \FcsrMcontrolTiming
                 \item \FcsrMcontrolSixHitZero behaves just as \FcsrMcontrolHit.
                 \item \FcsrMcontrolSixHitOne is read-only 0.
+                \item Encodings for \FcsrMcontrolSixSize for access sizes larger
+                than 64 bits are different.
                 \end{steps}
             </value>
             <value v="1" name="1">
@@ -755,8 +757,8 @@
             Any bits beyond the size of the data access will contain 0.
             </value>
         </field>
-        <field name="0" bits="20" access="R" reset="0" />
-        <field name="size" bits="19:16" access="WARL" reset="0">
+        <field name="0" bits="20:19" access="R" reset="0" />
+        <field name="size" bits="18:16" access="WARL" reset="0">
             <value v="0" name="any">
             The trigger will attempt to match against an access of any size.
             The behavior is only well-defined if $|select|=0$, or if the access
@@ -786,19 +788,7 @@
             execution of 64-bit instructions.
             </value>
 
-            <value v="6" name="80bit">
-            The trigger will only match against execution of 80-bit instructions.
-            </value>
-
-            <value v="7" name="96bit">
-            The trigger will only match against execution of 96-bit instructions.
-            </value>
-
-            <value v="8" name="112bit">
-            The trigger will only match against execution of 112-bit instructions.
-            </value>
-
-            <value v="9" name="128bit">
+            <value v="6" name="128bit">
             The trigger will only match against 128-bit memory accesses or
             execution of 128-bit instructions.
             </value>
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -147,11 +147,31 @@
     </register>
 
     <register name="Trigger Info" short="tinfo" address="0x7a4">
-        This register is optional if no triggers are implemented, or if \FcsrTdataOneType
-        is not writable. In this case the debugger can read the only supported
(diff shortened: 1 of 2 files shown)


xml/hwbp_registers.xml
```diff
@@ -610,13 +630,18 @@
     <register name="Match Control Type 6" short="mcontrol6" address="0x7a1">
         This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 6.
 
+        Implementing this trigger as described here requires that
+        \FcsrTinfoVersion is 1 or higher, which in turns means \RcsrTinfo must
```
pdonahue-ventana, line 634: "in turn"

pdonahue-ventana (commented):
Should there be some (possibly non-normative) text about what debuggers should expect with respect to hit0/1, timing, tinfo.version, etc.?  Something along the lines of the discussion from the last few days.  That seems better than having people read two different (but 99% identical) versions of the spec to figure out what they have to put in their `if (tinfo.version == 0)` statement.

xml/hwbp_registers.xml
```diff
@@ -147,11 +147,31 @@
...
-        type from \RcsrTdataOne.
-
-        <field name="0" bits="XLEN-1:16" access="R" reset="0" />
+        This register is optional if no triggers are implemented, or if
+        \FcsrTdataOneType is not writable and \FcsrTinfoVersion would be 0. In
+        this case the debugger can read the only supported type from
+        \RcsrTdataOne.
+
+        <field name="0" bits="XLEN-1:32" access="R" reset="0" />
+        <field name="version" bits="31:24" access="R" reset="Preset">
```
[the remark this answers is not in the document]
pdonahue-ventana: I'm reading things out of order.  You already addressed this.

If we went with 4 bits and expanded to 8 later then we could only have 31 versions in 8 bits.  0-0xE are 15 versions, then 0xF would say to look at the other 4 bits which encode another 16 versions.  I agree that we should plan for 256 versions and reclaim some bits later if necessary.

xml/hwbp_registers.xml
```diff
@@ -147,11 +147,31 @@
...
+            Contains the version of the Sdtrig extension implemented.
+            <value v="0" name="0">
+                Supports triggers as described in this spec from commit 5a5c078,
```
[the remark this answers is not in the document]
timsifive: "at commit"? On and before implies any time before, and that's not right either.

xml/hwbp_registers.xml
```diff
@@ -147,11 +147,33 @@
...
+            <value v="0" name="0">
+                Supports triggers as described in this spec at commit 5a5c078,
+                made on February 2, 2023.
+
+                \begin{steps}{In these older versions:}
+                \item \RcsrMcontrolSix has a timing bit identical to
+                    \FcsrMcontrolTiming
+                \item \FcsrMcontrolSixHitZero behaves just as \FcsrMcontrolHit.
+                \item \FcsrMcontrolSixHitOne is read-only 0.
+                \item Encodings for \FcsrMcontrolSixSize for access larger than
```
pdonahue-ventana, line 167: "accesses" or "access sizes" or something plural.

xml/hwbp_registers.xml
```diff
@@ -770,9 +797,7 @@
             Any bits beyond the size of the data access will contain 0.
             </value>
         </field>
-        <field name="0" bits="20" access="WARL" reset="0">
-            This used to be used for timing, but is currently read only 0.
-        </field>
+        <field name="0" bits="20" access="R" reset="0" />
         <field name="size" bits="19:16" access="WARL" reset="0">
```
pdonahue-ventana, line 801: Should this be 18:16?

pdonahue-ventana (commented):
Removing the 80/96/112 encodings is fine with me.  They can be added later (with different encoding values) if anybody ever needs them.

Also, remove the PDF.

xml/hwbp_registers.xml
timsifive: I don't think it matters. Either way we can use that bit for more size encodings, or for a new feature. So that means it's better to be explicit and reduce the size like you suggest.

pdonahue-ventana (approved):
OK.

PR #810, opened 2023-02-21
Refactor AMO discussion.

timsifive:
Now it's only in one place, instead of duplicated for mcontrol and mcontrol6.


timsifive:
Yes, eventually. I did this because I'm starting to write how vector load/store and cm.push/cm.pop should act, and it seemed silly to add yet more duplicate text.

I'd like to not chew this all off in a single PR.

pdonahue-ventana (approved):
OK.  I thought that it was weird that there was only a single subsection.  If this is going to have more content then that makes more sense.

PR #821, opened 2023-03-14
Describe how CMO instructions match triggers.

timsifive:
See e-mail subject:"Sdtrig and CMO".

This partially contradicts what the CMO spec says, so we'll need to figure that out before merging this.

### DIFF
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -275,6 +275,28 @@ individually. E.g. a vector load should be treated as if it performed multiple
 loads of size SEW (selected element width), and {\tt cm.push} should be treated
 as if it performed multiple stores of size XLEN.
 
+\subsection{Cache Operations}
+
+Cache operations are unusual events, and code that uses them can have
+hard-to-find bugs. To help debug such code, cache operations that affect
+consistency between memory and caches should be treated as if they are stores to
+all the addresses that are affected by the operation.  Operations that store to
+valid cache lines should also match as if they are stores.
+
+The access size for such operations is unknown to the TM, so only triggers with
+\FcsrMcontrolSixSize=0 will match.
+
+\begin{commentary}
+    \begin{steps}{For RISC-V Base Cache Management Operation ISA Extensions
+            1.0.1, this means the following:}
+        \item \inst{cbo.clean}, \inst{cbo.flush}, and \inst{cbo.inval} match as
+            if they are stores because they affect consistency.
+        \item \inst{cbo.zero} matches as if it is a store because it stores to
+            valid cache lines.
+        \item The prefetch instructions don't match at all.
+    \end{steps}
+\end{commentary}
+
 \section{Trigger Registers}
 
 These registers are CSRs, accessible using the RISC-V {\tt csr} opcodes and
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -286,13 +286,18 @@ otherwise non-coherent implicit and explicit memory accesses.
 \item Cache operations that perform block writes of constant data.
 \end{steps}
 
-Only triggers with \FcsrMcontrolSixSize=0 and \FcsrMcontrolSixSelect=0 will
-match. These triggers must at least be compared against the lowest virtual
-address covered by any of the affected cache lines, or the virtual address that
-is the argument
-to the instruction. Comparing against both of these is better, and best of all
-is to match against any virtual address covered by any of the affected cache
-lines.
+\begin{steps}{Only triggers with \FcsrMcontrolSixSize=0 and
+\FcsrMcontrolSixSelect=0 will match. Since cache operations affect multiple
+addresses, there are multiple possible values to compare against.
+Implementations must implement one of the following options.  From most
+desirable to least desirable, they are:}
+\item Every virtual address that is affected by the cache operation is a compare
+    value.
+\item The lowest virtual address that is affected by the cache operation is a
+    compare value.
+\item The virtual address that is the argument to the instruction is a compare
+    value.
+\end{steps}
 
 Cache operations encoded as HINTs do not match debug triggers.
 
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -277,22 +277,30 @@ as if it performed multiple stores of size XLEN.
 
 \subsection{Cache Operations}
 
+\begin{steps}{
 Cache operations are unusual events, and code that uses them can have
-hard-to-find bugs. To help debug such code, cache operations that affect
-consistency between memory and caches should be treated as if they are stores to
-all the addresses that are affected by the operation.  Operations that store to
-valid cache lines should also match as if they are stores.
-
+hard-to-find bugs. For the purposes of debug triggers, two classes of cache
+operations match as stores against any of the affected addresses:}
+\item Cache operations that enable software to maintain coherence between
+otherwise non-coherent implicit and explicit memory accesses.
+\item Cache operations that perform block writes of constant data.
+\end{steps}
 The access size for such operations is unknown to the TM, so only triggers with
 \FcsrMcontrolSixSize=0 will match.
 
+Cache operations encoded as HINTs do not match debug triggers.
+
 \begin{commentary}
+    The above language intends to capture the trigger behavior with respect to
+    the cache operations to be introduced in a forthcoming I/D consistency
+    extension.
+
     \begin{steps}{For RISC-V Base Cache Management Operation ISA Extensions
             1.0.1, this means the following:}
         \item \inst{cbo.clean}, \inst{cbo.flush}, and \inst{cbo.inval} match as
             if they are stores because they affect consistency.
-        \item \inst{cbo.zero} matches as if it is a store because it stores to
-            valid cache lines.
+        \item \inst{cbo.zero} matches as if it is a store because it performs a
+            block write of constant data.
         \item The prefetch instructions don't match at all.
     \end{steps}
 \end{commentary}
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -278,7 +278,7 @@ as if it performed multiple stores of size XLEN.
 \subsection{Cache Operations}
 
 \begin{steps}{
-Cache operations are unusual events, and code that uses them can have
+Cache operations are infrequently performed, and code that uses them can have
 hard-to-find bugs. For the purposes of debug triggers, two classes of cache
 operations match as stores against any of the affected addresses:}
 \item Cache operations that enable software to maintain coherence between
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -291,11 +291,7 @@ otherwise non-coherent implicit and explicit memory accesses.
 addresses, there are multiple possible values to compare against.
 Implementations must implement one of the following options.  From most
 desirable to least desirable, they are:}
-\item Every virtual address that is affected by the cache operation is a compare
-    value. (This is hard to implement if caches are physically tagged and
-    address translation means a single cache block might correspond to multiple
-    virtual addresses.)
-\item Every virtual address from the effective address rounded down to the
+\item Every address from the effective address rounded down to the
     nearest cache block boundary (inclusive) to the effective address rounded up to
     the nearest cache block boundary (exclusive) is a compare value.
 \item The effective address rounded down to the nearest cache block boundary is a
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -288,9 +288,11 @@ otherwise non-coherent implicit and explicit memory accesses.
 
 Only triggers with \FcsrMcontrolSixSize=0 and \FcsrMcontrolSixSelect=0 will
 match. These triggers must at least be compared against the lowest virtual
-address of all affected cache lines, or the virtual address that is the argument
+address covered by any of the affected cache lines, or the virtual address that
+is the argument
 to the instruction. Comparing against both of these is better, and best of all
-is to match against the virtual address covered by all affected cache lines.
+is to match against any virtual address covered by any of the affected cache
+lines.
 
 Cache operations encoded as HINTs do not match debug triggers.
 
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -280,13 +280,17 @@ as if it performed multiple stores of size XLEN.
 \begin{steps}{
 Cache operations are infrequently performed, and code that uses them can have
 hard-to-find bugs. For the purposes of debug triggers, two classes of cache
-operations match as stores against any of the affected addresses:}
+operations must match as stores:}
 \item Cache operations that enable software to maintain coherence between
 otherwise non-coherent implicit and explicit memory accesses.
 \item Cache operations that perform block writes of constant data.
 \end{steps}
-The access size for such operations is unknown to the TM, so only triggers with
-\FcsrMcontrolSixSize=0 will match.
+
+Only triggers with \FcsrMcontrolSixSize=0 and \FcsrMcontrolSixSelect=0 will
+match. These triggers must at least be compared against the lowest virtual
+address of all affected cache lines, or the virtual address that is the argument
+to the instruction. Comparing against both of these is better, and best of all
+is to match against the virtual address covered by all affected cache lines.
 
 Cache operations encoded as HINTs do not match debug triggers.
 
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -292,11 +292,15 @@ addresses, there are multiple possible values to compare against.
 Implementations must implement one of the following options.  From most
 desirable to least desirable, they are:}
 \item Every virtual address that is affected by the cache operation is a compare
-    value.
-\item The lowest virtual address that is affected by the cache operation is a
+    value. (This is hard to implement if caches are physically tagged and
+    address translation means a single cache block might correspond to multiple
+    virtual addresses.)
+\item Every virtual address from the effective address rounded down to the
+    nearest cache block boundary (inclusive) to the effective address rounded up to
+    the nearest cache block boundary (exclusive) is a compare value.
+\item The effective address rounded down to the nearest cache block boundary is a
     compare value.
-\item The virtual address that is the argument to the instruction is a compare
-    value.
+\item The effective address of the instruction is a compare value.
 \end{steps}
 
 Cache operations encoded as HINTs do not match debug triggers.
--- a/riscv-debug-spec.tex
+++ b/riscv-debug-spec.tex
@@ -17,8 +17,11 @@
 \makeindex
 
 \usepackage{xspace}
-\newcommand{\defregname}[2]{\providecommand{#1}{{\tt #2}\xspace}}
-\newcommand{\deffieldname}[2]{\providecommand{#1}{{$|#2|$}\xspace}}
+\newcommand{\inst}[1]{{\tt #1}\xspace}
+\newcommand{\reg}[1]{{\tt #1}\xspace}
+\newcommand{\field}[1]{{$|#1|$}\xspace}
+\newcommand{\defregname}[2]{\providecommand{#1}{\reg{#2}}}
+\newcommand{\deffieldname}[2]{\providecommand{#1}{\field{#2}}}
 \deffieldname{\FcsrMstatusMprv}{MPRV}
 \deffieldname{\FcsrMstatusMie}{MIE}
 \deffieldname{\FcsrSstatusSie}{SIE}


pdonahue-ventana (commented):
What if menvcfg.CBIE=01 and we do a cbo.inval?  I think that it should be optional whether it matches or not.  I imagine that some implementations will allow CBIE to be writable and will not want to qualify with CBIE, so they'll match.  Other implementations may hardwire CBIE[1]=0 in which case they'll want to treat cbo.inval exactly the same as cbo.flush (other than looking at CBIE[0] vs. CBCFE[0] at instruction decode time) so they'll want to not match.

This is also not clear whether "did XLEN-bit stores to memory" means that a cbo.inval did a single XLEN-bit store to memory or did a sequence of XLEN-bit stores up to the cache block size.

timsifive:
> What if menvcfg.CBIE=01 and we do a cbo.inval?

Good point. In that case it's not a bunch of stores, so shouldn't match, right? But this is really getting into details of the cache spec that I'd like to not have in the debug spec. I'll mention menvcfg.CBIE, but also move the specific instruction stuff into a commentary section.

I'd like the debug spec to layout principles, that can then be followed by other specs.

pdonahue-ventana:
> > What if menvcfg.CBIE=01 and we do a cbo.inval?

> Good point. In that case it's not a bunch of stores, so shouldn't match, right? But this is really getting into details of the cache spec that I'd like to not have in the debug spec. I'll mention menvcfg.CBIE, but also move the specific instruction stuff into a commentary section.

You're right.  I missed "Cache operations that change how memory looks to the hart" and read it as either cache operations in general or cache instructions that could possibly change how memory looks.

I also wouldn't say "how memory looks to the hart" but rather something like "the contents of memory as seen by the hart" since the way it "looks" is kind of vague (the value, some PMA attributes, etc.).

> I'd like the debug spec to layout principles, that can then be followed by other specs.

Yes, we should principally focus on principles. 😄 

I also want to discuss:
> This is also not clear whether "did XLEN-bit stores to memory" means that a cbo.inval did a single XLEN-bit store to memory or did a sequence of XLEN-bit stores up to the cache block size.

It's not clear to me that treating it like XLEN is better than what the TG suggested:
> Unless an encoding for a cache block is added to the mcontrol6.size field, an address trigger should
only match a memory access from a CBO instruction if mcontrol6.size=0

timsifive:
> I also wouldn't say "how memory looks to the hart" but rather something like "the contents of memory as seen by the hart" since the way it "looks" is kind of vague (the value, some PMA attributes, etc.).

Hmm... I didn't want to refer to the contents, because the contents don't actually change. Its just the hart's perception that changes. But "how memory looks" could be read as changing a memory mapping or something. 

How about "cache operations that change the apparent contents of memory"?

pdonahue-ventana:
I think that "contents of memory as seen by the hart" is distinct from "contents of memory" but I'm OK with "apparent contents of memory" if you prefer that.

Sdtrig.tex
```diff
@@ -275,6 +275,27 @@ \subsection{Combined Accesses}
 loads of size SEW (selected element width), and {\tt cm.push} should be treated
 as if it performed multiple stores of size XLEN.
 
+\subsection{Cache Operations}
+
+As far as triggers are concerned, caches don't exist. Cache operations that
+change the apparent contents of memory should match triggers as if those
+operations did XLEN-bit stores to memory. No other cache operations should match
+triggers.
```
pdonahue-ventana, line 283: If memory happened to already contain 0 and somebody does a cbo.zero then it doesn't change the apparent contents of memory.  I assume that we agree that address triggers on cbo.zero should always match without doing any data comparison.

Should we just spell out all the cases from the cache management spec instead of having this brief description?  We might elaborate on the Zicbom/Zicbop cases to say that what they generally say is "optional" means that actual changes to the apparent contents of memory should match like stores and non-changes (like flushes or prefetches) are not required to or are encouraged not to match.  (I don't think that we need to require that prefetches must not match.  Maybe it's much easier for some implementation to match on those or somebody has some special reason to care about prefetches.)

pdonahue-ventana:
@dkruckemyer-ventana is more of the guru in this space.

timsifive:
We have one week until we need to get this in and frozen. Can you corner @dkruckemyer-ventana in a hallway or something?

dkruckemyer-ventana:
My primary comment is that we should treat cbo.inval more like cbo.clean and cbo.flush and less like cbo.zero. Inval really isn't a store in the sense that it explicitly modifies memory (cbo.zero clearly does), and it really operates on cached copies with the side-effect that it may expose older data values (it may also expose the same values). So I think it's better to take all the management operations as a package, i.e. either trigger on them or don't.

On that particular note, I do think it's valuable to trigger on all the management operations to help debug code that is performing non-coherent communication with another agent. If you have a memory location that is being used for non-coherent communication, I would think you would want to trigger on all operations to that address to know whether you implemented the proper SW sequence. Looking ahead, I would think we want to incorporate the I/D coherence operations as well, since that extension will introduce additional sequences that need to be performed correctly. So my suggestion would be to trigger on all management CBOs.

Finally, does the statement about XLEN imply that a trigger will match on a CBO only if the data size is equal to XLEN (assuming all the other conditions are met) and won't match otherwise? That seems a little non-intuitive to me, especially since cbo.zero (like the management CBOs) operates on a full cache line (e.g. 64B) instead of 4B or 8B. Also, does the matching address need to be aligned to XLEN? The CBO spec does not require any address alignment, so requiring that for the triggers would potentially miss some CBO instructions with unaligned addresses.

timsifive:
> My primary comment is that we should treat cbo.inval more like cbo.clean and cbo.flush and less like cbo.zero. Inval really isn't a store in the sense that it explicitly modifies memory (cbo.zero clearly does), and it really operates on cached copies with the side-effect that it may expose older data values (it may also expose the same values). So I think it's better to take all the management operations as a package, i.e. either trigger on them or don't.

That sounds reasonable, but I'm still left with this feeling:
If you do a load, then perform another instruction, and do another load, you should get the same value (memory-mapped devices aside). If you don't get the same value, then the other instruction modified memory, right?

> On that particular note, I do think it's valuable to trigger on all the management operations to help debug code that is performing non-coherent communication with another agent. If you have a memory location that is being used for non-coherent communication, I would think you would want to trigger on all operations to that address to know whether you implemented the proper SW sequence. Looking ahead, I would think we want to incorporate the I/D coherence operations as well, since that extension will introduce additional sequences that need to be performed correctly. So my suggestion would be to trigger on all management CBOs.

So conceptually, you would like triggers to say they trigger on load, store, execute, or a new 4th special class that includes cache operations. Is that right? That sounds reasonable to me. We have a bit that we could use. What should the new category be called?

> Finally, does the statement about XLEN imply that a trigger will match on a CBO only if the data size is equal to XLEN (assuming all the other conditions are met) and won't match otherwise? That seems a little non-intuitive to me, especially since cbo.zero (like the management CBOs) operates on a full cache line (e.g. 64B) instead of 4B or 8B. Also, does the matching address need to be aligned to XLEN? The CBO spec does not require any address alignment, so requiring that for the triggers would potentially miss some CBO instructions with unaligned addresses.

Rereading the definition of mcontrol6.size, it sounds like size=0 can work on data matches, as long as the access size is XLEN. I'm happy to remove the XLEN language from this PR. Noone seems to agree with me in any case. :)

pdonahue-ventana:
> Looking ahead, I would think we want to incorporate the I/D coherence operations as well

FYI, this is referring to [Zjid](https://github.com/riscv/riscv-j-extension/tree/master/zjid).

dkruckemyer-ventana:
> That sounds reasonable, but I'm still left with this feeling: If you do a load, then perform another instruction, and do another load, you should get the same value (memory-mapped devices aside). If you don't get the same value, then the other instruction modified memory, right?

The instruction really *unmodified* memory in the case that no non-coherent agent write the memory location directly. However, a primary use-case for inval (as well as flush) is to eliminate any cached copies that may be stale because a non-coherent agent did write newer values into memory. So triggering on inval (and other management ops) makes sense as part of the non-coherent communication sequence.

> So conceptually, you would like triggers to say they trigger on load, store, execute, or a new 4th special class that includes cache operations. Is that right? That sounds reasonable to me. We have a bit that we could use. What should the new category be called?

I suppose you could use the bit for an "other" category, but we kinda envisioned the execution behavior of management and zero CBOs to be similar to stores (i.e. from a memory ordering, trapping, speculation, etc. point of view -- permissions are different, however). So from that perspective, I'd be OK to say that CBOs match when the store bit is set, rather than use up another bit.

> Rereading the definition of mcontrol6.size, it sounds like size=0 can work on data matches, as long as the access size is XLEN. I'm happy to remove the XLEN language from this PR. Noone seems to agree with me in any case. :)

Sounds good to me.

timsifive:
> I suppose you could use the bit for an "other" category, but we kinda envisioned the execution behavior of management and zero CBOs to be similar to stores (i.e. from a memory ordering, trapping, speculation, etc. point of view -- permissions are different, however). So from that perspective, I'd be OK to say that CBOs match when the store bit is set, rather than use up another bit.

My concern is that someone wants to debug a process, and doesn't want the trigger to fire every time there's a cache operation that's really a nop as far as that process is concerned. But maybe this doesn't matter because the process would execute in a different privilege mode than the cache operations.

Does the same logic apply to debugging a kernel? Everything is M-Mode, and you just want to find when some memory is modified. Do you care to be notified every time that memory is evicted from the cache? I think you probably do, since presumably you don't use the same virtual addresses in processes as for the kernel, so there's no need to evict every time you switch to a new process.

Are you convinced that treating CMO as stores will never cause an annoying number of false positives, or that it commonly will? Are there cases I didn't think of?

pdonahue-ventana:
> My concern is that someone wants to debug a process, and doesn't want the trigger to fire every time there's a cache operation that's really a nop as far as that process is concerned. But maybe this doesn't matter because the process would execute in a different privilege mode than the cache operations.

If the process is doing cache ops then presumably there's something special going on.  Matching on cache ops allows you to debug the special thing.  If you're doing cache ops for no particular reason then you might also want to know about that so you can ask yourself why those cache ops are in there at all.

> Does the same logic apply to debugging a kernel? Everything is M-Mode, and you just want to find when some memory is modified. Do you care to be notified every time that memory is evicted from the cache? I think you probably do, since presumably you don't use the same virtual addresses in processes as for the kernel, so there's no need to evict every time you switch to a new process.

For instance, you might be debugging noncoherent I/O.  The hart needs to flush everything out of the caches so everyone can see the correct data that some noncoherent device writes to a buffer or reads from a buffer.  If you're seeing stale data then you probably forgot the flush.  Stopping on the flush allows you to debug the recipe.

The other major RISC architecture that I'm familiar with treats cache invalidate instructions as stores for purposes of watchpoints.

timsifive:
I think I've addressed everyone's thoughts with this update. Please take a look. Please especially verify that this text will already answer the question as to what the behavior for future I/D-cache coherence instructions should be.

Sdtrig.tex
```diff
@@ -275,6 +275,36 @@ \subsection{Combined Accesses}
...
+\begin{steps}{
+Cache operations are unusual events, and code that uses them can have
+hard-to-find bugs. For the purposes of debug triggers, two classes of cache
+operations match as stores against any of the affected addresses:}
+\item Cache operations that enable software to maintain coherence between
+otherwise non-coherent implicit and explicit memory accesses.
+\item Cache operations that perform block writes of constant data.
+\end{steps}
+The access size for such operations is unknown to the TM, so only triggers with
+\FcsrMcontrolSixSize=0 will match.
```
pdonahue-ventana, line 289: It's not clear to me that the size is always unknown to the TM.  The recommendation in select=0 (to compare on all addresses that are affected by an operation) mean that you would ideally be comparing on all 64 addresses (or however long the cache block size is) which means that a recommended implementation actually would know the size.

This also brings up a question.  Consider the case where we match on store address and tdata2=4.  If you do a cache op with rs1=4 then is this required to match?  My question has to do with "the lowest virtual address of the access" in the select=0 text.  If you only have one comparison then does that comparison need to clear the low 6 bits before comparing (since the lowest VA of the access is 0) or are you allowed to match on the address provided in the register?  The former would not match if rs1=4 even though tdata2=4 but the latter would (and would not match on the lowest VA). 
 This is about a minimally compliant implementation, not one that complies with the recommendation.

pdonahue-ventana: I think that "is unknown to the TM" is just wordsmithing which can maybe be done post-freeze.  The other question about lowest VA is probably more substantive.

My inclination is to change "lowest virtual address of the access" in select=0 to say "lowest virtual address of the access or in the case of cache operations the base address from the register."  Basically, it's always the output of the base+offset calculation.  Then nothing in this section needs to change (other than minor clarifications).

timsifive: > It's not clear to me that the size is always unknown to the TM. The recommendation in select=0 (to compare on all addresses that are affected by an operation) mean that you would ideally be comparing on all 64 addresses (or however long the cache block size is) which means that a recommended implementation actually would know the size.

This is referring to the *access* size, not the number of bytes affected by the operation. The number of bytes should be known.

pdonahue-ventana: I can see arguments either way.  It's a little awkward to add a masking capability before the comparison just for this one case.

However, the CMO spec says:
> the compare values
follow from the addresses of the NAPOT memory region corresponding to the cache block containing
the effective address

To me, that means that the CMO group was aware of this concept of multiple compare values, that the lowest VA of the line is included in that set of compare values, and that they didn't say anything along the lines of my proposal.

I'll drop my question and let's go with the status quo of lowest VA.

My only (minor) concern now is "The access size for such operations is unknown to the TM" which just doesn't seem necessarily true.

timsifive: > I can see arguments either way. It's a little awkward to add a masking capability before the comparison just for this one case.

Don't you already do that masking to find the correct cache line? On the other hand, maybe the cache is doing the masking, and the instruction execution unit doesn't actually know the cache line size or anything. In which case all it can do is go by the value given to the cache.

So... should (but not must) match against the lowest address in the cache line?

pdonahue-ventana: I'm OK with should.

pdonahue-ventana (approved):
The wording of the second paragraph isn't my favorite but I'm OK with the sentiment.  Wording can possibly be tweaked later but we need to get the actual normative requirements in here first.

dkruckemyer-ventana:
I'm confused about what is meant by "all affected cache lines." Do you mean all addresses within a cache block?

pdonahue-ventana:
Good question.  I thought that it said the address of each byte that's affected but that's not what it actually says.

timsifive:
I assume all current instructions affect a single cache line. E.g. invalidate at virtual address, etc. What if future instructions affect multiple cache lines? I want triggers to to ideally match on every address in any of those cache lines. Does that make sense? What phrase describes that well?

dkruckemyer-ventana:
Doesn't the NAPOT match mode cover all of this? I thought that let the debugger set a range of addresses to match against.

timsifive:
Yes, but NAPOT is optional, and even if it is implemented the user might not use it for the trigger.

The question here is what happens when someone sets a store breakpoint at 0x80011, and there's a cache op that clears the line containing 0x80008. With 32-byte cache lines, that means the starting address of the affected line is 0x80000. Ideally you'd trigger, even though neither of those addresses exactly match the store breakpoint.

dkruckemyer-ventana:
IIRC, we (the CMO group) expected that NAPOT matches would be used if you wanted to match on the range of addresses in a cache block, especially since cache block size may differ on different harts. Otherwise, the debug triggers have to understand the cache block size for management operations, zero operations, etc. (These may all be different in some implementations in the future.) But the debug software could always discover the block size and program the NAPOT regions appropriately.

timsifive:
Using NAPOT doesn't solve everything either, because then you might end up with a lot of false positives for regular stores that you might not want to trigger on.

Regardless, we do need to answer the question of what happens when the store trigger is a simple address trigger. I'll rephrase the PR again and see if it makes more sense.

dkruckemyer-ventana:
If you care about a range of addresses, I'm not sure how you can avoid false positives on stores unless you want an explicit CMO or "other" bit in the match control register. On the other hand, if you're performing CMOs on a range, you probably want to trigger on loads and stores to that region anyway either to make sure the associated loads and stores are being executed correctly or to make sure that spurious loads and stores are not being performed.

timsifive:
In this case you don't care about a range. You only care about one address. So you set a trigger on that one address. Ideally, if any CMO affects that address, you want the trigger to fire. But that's not always easy to implement, so we're giving implementers some options.

This discussion doesn't apply to when you explicitly care about the entire range and use e.g. NAPOT to trigger on it.

dkruckemyer-ventana:
Got it. Yes, my perspective was from the that of the CMO, not an individual address that may get caught up by a CMO (though one might reasonably expect if there were any CMOs in the vicinity, one would cast the NAPOT net wide enough, but that may not be all use-cases).

The existing text is fine, though it's not particularly clear what a "compare value" is in this context. Is the compare value the contents of the trigger register or is the compare value the addresses associated with the operation that is compared against the contents of the trigger register?

Also, do you mean effective address (i.e. the address generated by the instruction) rather than virtual address? Depending on the translation modes enabled, you may or may not have a true virtual address.

timsifive:
> The existing text is fine, though it's not particularly clear what a "compare value" is in this context. Is the compare value the contents of the trigger register or is the compare value the addresses associated with the operation that is compared against the contents of the trigger register?

That's fair. The compare value are the addresses associated with the operation. Their selection is documented at https://github.com/riscv/riscv-debug-spec/blob/master/xml/hwbp_registers.xml#L769, and are referenced with the various match operations at https://github.com/riscv/riscv-debug-spec/blob/master/xml/hwbp_registers.xml#L885.

I generally assume that people implementing triggers are familiar with that. This is not really great, since it occurs later in the document, but I don't want to make this particular PR any larger. Should I file an issue to clarify what exactly compare values are? I can do that after this one merges, since that would just be a clarification.

> Also, do you mean effective address (i.e. the address generated by the instruction) rather than virtual address? Depending on the translation modes enabled, you may or may not have a true virtual address.

I don't know what would be a true virtual address versus a regular. Our current address triggers just refer to "virtual address."

Looking at the CMO spec again, "effective" address is what I mean. It's the address that you get just by looking at the instruction and GPR values.

I then assume that effective address is the virtual address, that is translated into a physical address.

What if the caches are physically tagged? In that case a single cache line could correspond to multiple virtual addresses. Requiring triggers to figure out what all the virtual addresses are that match that cache line is unreasonable. So I guess I need a term that means "virtual base address of the cache line(s) specified by the effective address, even if the cache is physically tagged." Is there a term like that?

dkruckemyer-ventana:
> > The existing text is fine, though it's not particularly clear what a "compare value" is in this context. Is the compare value the contents of the trigger register or is the compare value the addresses associated with the operation that is compared against the contents of the trigger register?
> 
> That's fair. The compare value are the addresses associated with the operation. Their selection is documented at https://github.com/riscv/riscv-debug-spec/blob/master/xml/hwbp_registers.xml#L769, and are referenced with the various match operations at https://github.com/riscv/riscv-debug-spec/blob/master/xml/hwbp_registers.xml#L885.
> 
> I generally assume that people implementing triggers are familiar with that. This is not really great, since it occurs later in the document, but I don't want to make this particular PR any larger. Should I file an issue to clarify what exactly compare values are? I can do that after this one merges, since that would just be a clarification.

No need. If the term is defined elsewhere, that's OK. I am admittedly reading out of context.

> 
> > Also, do you mean effective address (i.e. the address generated by the instruction) rather than virtual address? Depending on the translation modes enabled, you may or may not have a true virtual address.
> 
> I don't know what would be a true virtual address versus a regular. Our current address triggers just refer to "virtual address."

By "true virtual address," I mean an actual virtual address, i.e. the input address to an enabled one or two stage translation mechanism. Depending on execution mode or whether translation is enabled or disabled, the address generated by the instruction can be interpreted as a virtual address, guest physical address, or physical address. I generally prefer "effective address" (rather than "virtual address") as the name for the address generated by an instruction (e.g. rs1, rs1+offset, etc.) when no context about execution modes or address translation is given.

> Looking at the CMO spec again, "effective" address is what I mean. It's the address that you get just by looking at the instruction and GPR values.
> 
> I then assume that effective address is the virtual address, that is translated into a physical address.

I would say the effective address *may* be a virtual address, depending on how it's used, but I generally figured the text meant that triggers matched on the address generated by an instruction (otherwise, one interpretation could be that M-mode triggers never match because M-mode instructions never generate virtual addresses). Just wanted to clarify the meaning.

> What if the caches are physically tagged? In that case a single cache line could correspond to multiple virtual addresses. Requiring triggers to figure out what all the virtual addresses are that match that cache line is unreasonable. So I guess I need a term that means "virtual base address of the cache line(s) specified by the effective address, even if the cache is physically tagged." Is there a term like that?

The CBO operations ultimately work on copies of cache blocks identified by physical addresses, whether those result from address translation or not. How the caches are tagged is otherwise invisible to software. From a trigger point of view, one could encounter virtual aliasing due to comparisons with the effective address, but that's true of regular loads and stores, too. I don't believe there's a requirement to determine all the virtual aliases for a given cache block (though I might be misinterpreting what you're getting at).

timsifive:
I've opened #837 to find uses of "virtual address" in this spec. My gut feeling is that we just want "effective address" everywhere that we said "virtual address."

> From a trigger point of view, one could encounter virtual aliasing due to comparisons with the effective address, but that's true of regular loads and stores, too.

Good point that this equally applies to load and stores. We don't mention this problem there, so there's no need to mention it for CMOs.

> I don't believe there's a requirement to determine all the virtual aliases for a given cache block (though I might be misinterpreting what you're getting at).

That is what I was getting at, and it does seem like some ideal implementation, but also nobody will ever do it.

So the language now is:
```
\begin{steps}{Only triggers with \FcsrMcontrolSixSize=0 and
\FcsrMcontrolSixSelect=0 will match. Since cache operations affect multiple
addresses, there are multiple possible values to compare against.
Implementations must implement one of the following options.  From most
desirable to least desirable, they are:}
\item Every address from the effective address rounded down to the
    nearest cache block boundary (inclusive) to the effective address rounded up to
    the nearest cache block boundary (exclusive) is a compare value.
\item The effective address rounded down to the nearest cache block boundary is a
    compare value.
\item The effective address of the instruction is a compare value.
\end{steps}
```

I'm making an assumption that effective addresses for CMOs always affect only the cache block that that effective address falls into. Is that true? Is my round up/round down phrasing correct?

timsifive:
Merging because today is the day we freeze substantive changes. I'm pretty sure this is in spirit what we want, even though we'll tweak the language later.

pdonahue-ventana:
I like that new list of possible compare values.  There might be some possible cleaning up around the edges but the concept is fine.

Issue #822, opened 2023-03-15
When trigger(load/store, action=1) match vector load/store instruction's element address,  should entry into Debug Mode?


GZYyyyy:
If any element of the vector load/store instruction matches the set trigger breakpoint, do I enter debug mode? The debug register information is saved for the vector instruction? For example, dpc holds the next vector instruction.

timsifive:
Very recently this clarification was added: https://github.com/riscv/riscv-debug-spec/blob/master/Sdtrig.tex#L269
I think that answers your first question.

I made https://github.com/riscv/riscv-debug-spec/pull/823 which hopefully answers your second question.

If not, let me know.

GZYyyyy:
Thank you.

PR #823, opened 2023-03-15
Describe what happens on halt.

timsifive:
This way I have a place to describe what happens with vector instructions.

Hopefully also answers part of #822.

### DIFF
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -149,6 +149,21 @@ enter Debug Mode before executing any instructions, but after performing any
 initialization that would usually happen before the first instruction is
 executed.
 
+\section{Halt}
+
+\begin{steps}{When a hart halts:}
+    \item \FcsrDcsrCause is updated.
+    \item \FcsrDcsrPrv and \FcsrDcsrV are set to reflect current privilege mode.
+    \item \RcsrDpc is set to the next instruction that should be executed.
+    \item If the current instruction can be partially executed and should be
+        restarted to complete, then the relevant state for that is updated. E.g.
+        if a halt occurs during a partially executed vector instruction, then
+        {\tt vstart} is updated, and \RcsrDpc is updated to the address of the
+        partially executed instruction. This is analogous to how vector
+        instructions behave for exceptions.
+    \item The hart enters Debug Mode.
+\end{steps}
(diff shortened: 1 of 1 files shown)


Sdext.tex
```diff
@@ -149,6 +149,20 @@ \section{Reset}
...
+
+\begin{steps}{When a hart halts:}
+    \item \FcsrDcsrCause is updated.
+    \item \FcsrDcsrPrv is set to the current privilege mode.
+    \item \RcsrDpc is set to the next instruction that should be executed.
+    \item If the current instruction can be partially executed and should be
+        restarted to complete, then the relevant state for that is updated. E.g.
+        if a halt occurs during a partially executed vector instruction, then
+        {\tt vstart} is updated, and \RcsrDpc is updated to the address of the
+        partially executed instruction.
```
pdonahue-ventana, line 162: Can we say something like "vstart is updated just like it would if it had any other exception (like a page fault) and dpc is updated just like mepc/sepc/vsepc would have been updated on any other exception (like a page fault)"?  (That needs some wordsmithing but I think it might be helpful for people to see that this basically uses existing mechanisms.)

Sdext.tex
```diff
@@ -149,6 +149,21 @@ \section{Reset}
...
+\begin{steps}{When a hart halts:}
+    \item \FcsrDcsrCause is updated.
+    \item \FcsrDcsrPrv and \FcsrDcsrV are set to reflect current privilege mode.
+    \item \RcsrDpc is set to the next instruction that should be executed.
+    \item If the current instruction can be partially executed and should be
+        restarted to complete, then the relevant state for that is updated. E.g.
+        if a halt occurs during a partially executed vector instruction, then
+        {\tt vstart} is updated, and \RcsrDpc is updated to the address of the
+        partially executed instruction. This is analogous to how vector
+        instruction behave for other exceptions.
```
pdonahue-ventana, line 163: While I think of halting as an exception, it's technically something different.  So removing the word "other" would keep the spirit of what you're saying and prevent an email question later.

pdonahue-ventana (approved):
Thanks.  I think that this will lead to fewer questions later.

Issue #824, opened 2023-03-23
Can DPC be updated on all committed instructions?


tariqkurd-repo:
The debug spec says:

`Upon entry to debug mode, dpc is updated with the virtual address of the next instruction to be executed.`

This doesn't preclude updating DPC at other times. In particlar I'm interested in allowing the debugger to monitor the current PC without halting the core, and reading DPC seems like a sensible way to achieve this.

Can there be a non-normative note added to the debug spec saying that this is legal behaviour?

Thanks

pdonahue-ventana:
Sure.  Tim created #825.

pdonahue-ventana:
#825 is now merged.

PR #825, opened 2023-03-23
Document behavior for abstract dpc access while running.

timsifive:
Addresses https://github.com/riscv/riscv-debug-spec/issues/824.

### DIFF
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -288,6 +288,20 @@
         \end{tabular}
         \end{table}
 
+        Executing the Program Buffer may cause the value of \RcsrDpc to become
+        \unspecified. If that is the case, it must be possible to read/write
+        \RcsrDpc using an abstract command with \FacAccessregisterPostexec not
+        set.  The debugger must attempt to save \RcsrDpc between halting and
+        executing a Program Buffer, and then restore \RcsrDpc before leaving
+        Debug Mode.
+
+        \begin{commentary}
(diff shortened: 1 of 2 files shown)


xml/core_registers.xml
```diff
@@ -288,6 +288,28 @@
...
+            Program Buffer.
+        \end{commentary}
+
+        If the Access Register abstract command supports reading \RcsrDpc while
+        the hart is running, then the value read should be the address of a
+        recently executed instruction.
+
+        If the Access Register abstract command supports writing \RcsrDpc while
+        the hart is running, then the executing program should jump to the
+        written address shortly after the write occurs.
```
pdonahue-ventana, line 311: The read behavior sounds good but this write behavior seems kind of weird.  Does it seem useful for debuggers to do this or would it be OK to just make writes have UNSPECIFIED behavior?

timsifive: I agree it's weird, but if write is implemented then this is the sensible thing IMO. I wouldn't expect anyone to support writing DPC while running.

pdonahue-ventana: I haven't thought about abstract reads/writes in a while so I forgot that it's allowable to support reads but not support writes.  This is fine since it has the "if" clause.

pdonahue-ventana (approved):

tariqkurd-repo:
That looks good to me.

Issue #812, opened 2023-03-02
Should icount fire when textra{32,64} don't match?


timsifive:
https://github.com/riscv-software-src/riscv-isa-sim/commit/d69a788be7feea8f529cb2dab0e9854f665055f5 says:

> Debug spec says about icount: "When pending is set, the trigger fires
> just before any further instructions are executed in a mode where the
> trigger is enabled"
> 
> And about icount textra: "If textra32 or textra64 are implemented for
> this trigger, it only matches when the conditions set there are
> satisfied."
> 
> I interpret that to mean that count should only be decremented when
> textra matches, but once pending, it should fire even if textra
> doesn't match anymore.

That is what the spec says, but I'm not sure it's what we want.

pending exists so that when you step an instruction which decrements icount to 0, that you then stop in the mode where the trigger is actually enabled instead of e.g. at an interrupt handler that happens to be entered then. textra.sselect and textra.mhselect do the same thing for supervisor/hypervisor context as the U, M, S, etc. bits do for execution modes. So pending should behave the same way.

Is that right?

pdonahue-ventana:
This will be fixed when #826 is merged.

timsifive:
#826 did merge.

pdonahue-ventana:
Yes, this should say "matching or firing" as you suggest.  Matching and firing are usually correlated but not with icount.  You don't want, for instance:
- M-mode-only implementation
- we're in a handler with MIE=0 (which is suppressing icount matching)
- set icount to count a single instruction.  This won't match until after we're out of the handler.
- mret back to non-handler code.  MIE=1 after the mret.
- take an interrupt before executing anything
- the interrupt does an icount match (but not a fire) and sets icount.pending
- we enter the interrupt handler with MIE=0 and pending=1 which (if not suppressed) will cause icount to fire

If icount fires then we'll go to the breakpoint handler and mepc will point to the first instruction of the interrupt handler.  We'll be able to return from the breakpoint handler to the interrupt handler but we lost the address that the interrupt handler should return to.  Firing needs to be suppressed until MIE=1 so we know that the interrupt handler has saved mepc on the stack (or whatever it needs to do).

pdonahue-ventana:
After all the PRs, I think that we're done with this one.

PR #826, opened 2023-03-23
textra inhibits firing in addition to matching.

timsifive:
Those two are distinct in the case of icount, where pending becomes set on match, and firing happens later. We want to make sure not to fire in a place where match is disallowed.

Addresses #812.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -1346,14 +1346,16 @@
             </value>
 
             <value v="4" name="mcontext">
-            This trigger will only match if the low bits of
+            This trigger will only match or fire if the low bits of
             \RcsrMcontext/\RcsrHcontext equal \FcsrTextraThirtytwoMhvalue.
             </value>
 
-            1, 5 (mcontext\_select): This trigger will only match if the low bits of
+            1, 5 (mcontext\_select): This trigger will only match or fire if the
+            low bits of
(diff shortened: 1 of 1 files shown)


pdonahue-ventana:
There are probably different implementations but it seems that you have:
- Somewhere around the instruction fetch (front of the pipe) you need to check the icount.pending bit and also the icount mode filters and textra.  Fire if the conditions warrant.
- Somewhere around retire (back of the pipe) you need to check icount mode filters and textra.  Decrement count and set icount.pending if the conditions warrant.

It seems unnatural to have the mode filters and textra being used both as the very first thing in the pipeline and also the very last thing.

Thoughts?  Is it just that debuggers want this and mode filters already work that way so this is just more of the same?

timsifive:
That's not ideal, but I don't really know how to do it any other way.

Having textra only affect matching or firing doesn't make sense. We need the pending bit to handle steps followed by an exception.

Simplest implementation might be to do all the checks up front, and then carry a single bit through the pipeline that indicates whether to decrement `icount` (and possibly set `pending`) at the end. That doesn't sound cheap either.

pdonahue-ventana (approved):
Yeah, this is probably not a huge problem.  Two different places might need copies of a couple of bits.  If we were to architect things from scratch we might have come up with a different way of doing mode filtering and/or holding icount pending etc.  With the constraints we have, this is fine.

Issue #813, opened 2023-03-02
Should tcontrol inhibit icount counting?


scottj97:
The spec says that when `tcontrol.mte == 0`:

> Triggers with action=0 do not match/fire while the hart is in M-mode.

And it says about icount:

> When count is greater than 1 and the trigger matches, then count is decremented by 1. (This is true even if the action configured in the trigger is inhibited for some reason.)

The first sentence there suggests that when inhibited by `tcontrol.mte` the count should not decrement, because `tcontrol` inhibits "match/fire" and this uses the same verb "matches".

But the second sentence makes me wonder if `tcontrol` is intended only to inhibit the action configured, and therefore a positive count should still be decremented.

Since the purpose of `tcontrol` is to inhibit reentrancy, it doesn't seem necessary (or desirable?) to inhibit the counting too.

scottj97:
Some interesting discussion on this in #555 and #557. IIUC when we added the `pending` bit that issue became moot, so #557 was reverted by #574. But the current text still suggests that counting should be squashed, when that apparently was not the intention after `pending` was added.

timsifive:
I don't think we considered icount of M-Mode when a trap handler is being executed.

I'm struggling to come up with an actual use case for it. I guess if you have a debug stub on the target, and it wants to single step, it can use icount for that. So it very carefully crafts code than can set icount to the proper value so that when it returns icount.count=1 and exactly one instruction executes before entering the debug stub handler. Except another exception happens. In that case you'd prefer for icount.count not to be decremented.

Or maybe you're using icount.count to turn trace on and off, in which case... who knows whether you want these trap instructions to be counted or not. If you just want N of your program instructions, then not. If you just want M instructions because that's all that fits in your trace buffer, then you do.

All of this is for some system where you're debugging M-Mode from M-Mode itself, which is clearly asking for trouble and tricky bugs in the debugger. Which is me just wondering if it really matters. Are people going to implement M-Mode only systems and then also icount triggers? Adding U-Mode is probably a lot more useful than icount triggers, and not much more expensive.

timsifive:
For consistency, it makes sense to have tcontrol and textra behave the same way.

scottj97:
This is confounded by this parenthetical statement in the icount spec:

> When count is greater than 1 and the trigger matches, then count is decremented by 1. (This is true even if the action configured in the trigger is inhibited for some reason.)

Why might the action be inhibited?

[...]

PR #829, opened 2023-03-30
icount: Remove confusing statement about inhibited trigger.

timsifive:
"Inhibited" is not well-defined, and this language is no longer necessary. It was added when the original sentence said:
> When \FcsrIcountCount is greater than 1, every instruction completed or
> trap taken from a mode where the trigger is enabled decrements
> \FcsrIcountCount by 1.

Now there is no more reference to a mode, so the clarification isn't needed either.

Addresses #813

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -1000,8 +1000,7 @@
         execution, the trigger still only matches once for that instruction.
 
         When \FcsrIcountCount is greater than 1 and the trigger matches, then
-        \FcsrIcountCount is decremented by 1. (This is true even if the action configured in
-        the trigger is inhibited for some reason.)
+        \FcsrIcountCount is decremented by 1.
 
         When \FcsrIcountCount is 1 and the trigger matches, then \FcsrIcountPending
         becomes set. In addition \FcsrIcountCount will become 0 unless it is


pdonahue-ventana (approved):
Yes.  If the trigger matches then decrement.  If the trigger doesn't match* then don't decrement. 
 That's it and we don't need to imply that there are exceptions.  

\* It doesn't matter whether it doesn't match because you're not retiring/trapping in an enabled mode (the "normal" case) or because you have some higher level match disable in tcontrol or mstatus.MIE (the "Native Triggers" case).

pdonahue-ventana:
> Does this imply that both mechanisms in section 5.4 “Native Triggers” (i.e. mstatus.mie or tcontrol.mte) should also inhibit icount decrementing?

I think so.

If you're in the handler (as indicated by either mechanism) and you have enabled icount in that mode for some reason then we all agree that you don't want to fire due to the reentrancy problem.  Do you want to match and set the pending bit so that you come back to the handler immediately upon leaving the handler (either via changing mie/mte directly while remaining in the handler's mode or by doing mret/sret) or do you want to step one instruction (or N instructions) after leaving the handler?  It seems preferable to only count the non-handler instructions as steps.  That would allow an M-mode-only implementation to actually use icount in a fairly straightforward way (without requiring count to be able to be >1 in hardware and having the code know how many instructions between the write to icount and the mret).

scottj97 (approved):

PR #832, opened 2023-04-05
Document native single step options.

timsifive:
Now include both mechanisms suggested for native triggers in single-privilege-mode systems, instead of just carefully counting instruction.

Addresses #830.

---
Improve single step description for different debug features

This commit updates the description of single step implementation in the debugger_implementation.tex file. It now provides a more detailed and clear explanation on how to single step through code running in the same privilege mode as the debugger, depending on the availability of \FcsrTcontrolMpte, \FcsrTcontrolMte, and automatically preventing \FcsrMcontrolSixAction=0 triggers from matching when entering a trap handler.
(gpt-4)

### DIFF
--- a/debugger_implementation.tex
+++ b/debugger_implementation.tex
@@ -606,7 +606,19 @@ the OS or debug stub writes \FcsrIcountCount=1, \FcsrIcountAction=0,
 \FcsrIcountM=0 before returning control to the lower user program with an
 {\tt mret} instruction.
 
-On tiny systems which only support M-Mode single step is doable, but tricky to
+Stepping code running in the same privilege mode as the debugger is more
+complicated, depending on what other debug features are implemented.
+
+If hardware implements \FcsrTcontrolMpte and \FcsrTcontrolMte, then stepping
+through non-trap code which doesn't allow for nested interrupts is also
+straightforward.
+
+If hardware automatically prevents \FcsrMcontrolSixAction=0 triggers from
+matching when entering a trap handler as described in
+Section~\ref{sec:nativetrigger}, then a carefully written trap handler can
+ensure that interrupts are disabled whenever the icount trigger must not match.
+
+If neither of these features exist, then single step is doable, but tricky to
 get right. To single step, the debug stub would execute something like:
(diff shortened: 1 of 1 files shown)


debugger_implementation.tex
```diff
@@ -606,7 +606,19 @@ \subsection{Single Step} \label{nativestep}
...
+complicated, depending on what other debug features are implemented.
+
+If hardware implements \FcsrTcontrolMpte and \FcsrTcontrolMte, then stepping
+through non-trap code which doesn't allow for nested interrupts is also
+straightforward.
+
+If hardware automatically prevents \FcsrMcontrolSixAction=0 triggers from
+matching when entering a trap handler as described in
+Section~\ref{sec:nativetrigger}, then a carefully written trap handler can
+ensure that interrupts are enabled whenever the icount trigger is enbled.
```
pdonahue-ventana, line 619:
```suggestion
ensure that interrupts are enabled whenever the icount trigger is enabled.
```

I wouldn't say that the handler should make sure that interrupts are enabled whenever the icount trigger is enabled.  It should make sure that interrupts are enabled whenever you want to match.  Or to count instructions.  Or whenever the trap handler wants icount to be active.  Or "armed" (though we don't use that word anywhere else so defining that opens a whole new challenge).

The word "enabled" is just too ambiguous for me.  My initial interpretation is just m=1.

timsifive: Updated. A little more double-negative than I prefer, but I'm not sure how to do it otherwise.

pdonahue-ventana: I see what you mean.  But I think it gets the point across and is unambiguous.

pdonahue-ventana (approved):

pdonahue-ventana:
By the way, did you write that summary of the change above or was that some github GPT autogenerated thing?

timsifive:
I wrote the stuff above the line. I've been experimenting with how well GPT4 makes commit messages. It's a bit mixed, but entertaining. It's almost always more verbose than I would be, which is not a bad thing.

Issue #783, opened 2022-12-07
Action 8 - 9


Silabs-ArjanB:
In 'table 5.1 action encoding' the values 8 - 9 are described as follows:

> Signal the firing of the trigger to other blocks within the hart (e.g. as countable events to hpmcounters). Use external debug trigger output 0 or 1 (respectively).

Why does it specifically state 'within the hart'? That seems to conflict with 'use external debug trigger output 0 or 1' as external triggers are 'abstract concepts that can signal the DM and/or receive signals from the DM' (i.e. there is no notion of 'within the hart' in respect to external triggers.

Could the description simply be as follows?

> Signal the firing of the trigger to other blocks via external debug trigger output 0 or 1 (respectively).

pdonahue-ventana:
tmexttrigger is a trigger that's external to the TM and that is defined to connect to something within the hart.  dmexttrigger is a trigger that's external to the DM and is intended to connect to something else in the system such as another DM.

An example of the former is given in the text.  You might want to have HPM count how many times you've executed a particular PC, loaded from a particular address, etc.

An example of the latter is if you have multiple DMs in the system and you want the benefit of halt groups and resume groups.  With a single DM, you can have the halting of a single hart cause other harts to halt.  With multiple DMs, you can still have the single halting hart use halt groups to halt other harts (if any) controlled by the same DM but you can simultaneously use dmexttrigger to have that DM send a trigger to other DMs which can then be programmed to halt a halt group.  This inter-DM cross triggering allows all harts in the system to halt.

The spec could theoretically allow TM external trigger outputs to connect to DM external trigger inputs.  I don't think that's useful since there's nothing you can do with these DM inputs other than halt/resume groups.  The only reason I can see to connect things like this is to have the fact that hart A executed a particular PC cause harts B and C to halt but not cause hart A to halt (by having B and C but not A in a halt group which is triggered by the input).  If you want to halt hart A then you'd just use action=1 (and optionally add B and C to a halt group if you want them to also halt).  This seems like a strange thing to do but I may be missing something.

Silabs-ArjanB:
Hello @pdonahue-ventana Thank you for your answer. Let me give an example of what kind of functionality I am looking for. Suppose I want to toggle some chip general purpose IO pin when the hart is accessing a certain address or executing an instruction at a specific address. Specifically we do not want to enter debug mode when this happens as that is considered too intrusive.

Do I though I could do that as follows:

[...]

PR #834, opened 2023-04-07
Clarify action 8/9 signal TM external triggers.

timsifive:
Also be more consistent in using "TM external triggers" in definition of tmexttrigger.

Addresses #783

### DIFF
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -86,12 +86,17 @@ Value & Description \\
 \hline
 5 & Reserved for use by the trace specification. \\
 \hline
-8 -- 9 & Signal the firing of the trigger to other blocks within the hart (e.g. as countable events to hpmcounters).  Use external debug trigger output 0 or 1 (respectively). \\
+8 -- 9 & Send a signal to TM external trigger output 0 or 1 (respectively). \\
 \hline
 other & Reserved for future use. \\
 \hline
 \end{tabular}
 \end{table}
+
+\begin{commentary}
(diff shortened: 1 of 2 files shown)


Sdtrig.tex
```diff
@@ -86,13 +86,19 @@ \section{Actions}
 \hline
 5 & Reserved for use by the trace specification. \\
 \hline
-8 -- 9 & Signal the firing of the trigger to other blocks within the hart (e.g. as countable events to hpmcounters).  Use external debug trigger output 0 or 1 (respectively). \\
+8 -- 9 & Send a signal to TM external trigger output 0 or 1 (respectively).  the
+firing of the trigger to blocks. \\
```
pdonahue-ventana, line 90: This second sentence looks like an editing error.

Sdtrig.tex
```diff
@@ -86,13 +86,19 @@ \section{Actions}
...
 \hline
 other & Reserved for future use. \\
 \hline
 \end{tabular}
 \end{table}
 
+\begin{commentary}
+    Actions 8 and 9 are intended to increment custom event counters, but these
+    signals could also be brought out to pins for use by external logic.
```
pdonahue-ventana, line 99: Maybe this is just me but I think that `s/out to pins/to outputs/` would be better.

Sdtrig.tex
timsifive: Thanks.

Sdtrig.tex
timsifive: That's more general. Thanks.

PR #836, opened 2023-04-13
Clarify that stepie applies to dcsr.step but not icount.

pdonahue-ventana:
A coworker assumed this morning that stepie affected single stepping with icount (which is not preferred but is allowed).  The spec wasn't clear one way or the other but I assume that this is the intent.

### DIFF
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -130,12 +130,14 @@
         </field>
         <field name="stepie" bits="11" access="WARL" reset="0">
             <value v="0" name="interrupts disabled">
-            Interrupts (including NMI) are disabled during single stepping.
+            Interrupts (including NMI) are disabled during single stepping
+            with \FcsrDcsrStep set.
             This value should be supported.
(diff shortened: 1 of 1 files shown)


timsifive (approved):

xiaoweish:
`Interrupts (including NMI) are disabled during single stepping with \FcsrDcsrStep set.`

Hi @pdonahue-ventana ,
Just to confirm my understanding -- does this mean that when `dcsr.step == 1`, the core is considered to be in single stepping mode, even though it may be entering and exiting debug mode for each instruction?

Thanks a lot!

Xiaowei

pdonahue-ventana:
> does this mean that when dcsr.step == 1, the core is considered to be in single stepping mode, even though it may be entering and exiting debug mode for each instruction?

Yes.  Entering and exiting debug mode for each instruction is the definition of single stepping mode.  See the "Single Step" section of the spec and the "Step Bit in Dcsr" subsection.

PR #838, opened 2023-04-17
Debug interaction with wrs.sto/wrs.nto

ved-rivos:
The `wrs.sto` and `wrs.nto` instructions follow the behavior of `wfi` for interaction with debug.
- Are `nop` in debug mode
- Are `nop` when stepped over
- Leave stalled stated on a halt request

### DIFF
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -91,6 +91,12 @@ If halt is requested while {\tt wfi} is executing, then the hart must leave the
 stalled state, completing this instruction's execution, and then enter Debug
 Mode.
 
+\section{Wait-on-Reservation-Set Instructions}
+
+If halt is requested while {\tt wrs.sto} or {\tt wrs.nto} is executing, then the
+hart must leave the stalled state, completing this instruction's execution, and
+then enter Debug Mode.
+
 \section{Single Step}
 
 \subsection{Step Bit In Dcsr} \label{stepBit}
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -36,7 +36,7 @@ How Debug Mode is implemented is not specified here.
 \item If \FcsrDcsrStoptime is 0 then \Rtime continues to update. If
     it is 1 then \Rtime will not update. It will resynchronize with
     \Rmtime after leaving Debug Mode.
-\item The {\tt wfi} instruction acts as a {\tt nop}.
+\item The {\tt wfi}, {\tt wrs.sto}, {\tt wrs.nto} instructions act as a {\tt nop}.
 \item Almost all instructions that change the privilege mode have \unspecified\
     behavior.  This includes {\tt ecall}, {\tt mret}, {\tt sret}, and {\tt uret}.
     (To change the privilege mode, the debugger can write
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -125,8 +125,9 @@ an instruction fetch causes an exception, that exception does not occur until
 the next time the hart is resumed. Similarly, a trigger at the new address does
 not fire until the hart actually attempts to execute that instruction.
 
-If the instruction being stepped over is {\tt wfi} and would normally stall the
-hart, then instead the instruction is treated as {\tt nop}.
+If the instruction being stepped over is {\tt wfi}, {\tt wrs.sto}, or {\tt wrs.nto}
+and would normally stall the hart, then instead the instruction is treated as
+{\tt nop}.
 
 \subsection{Icount Trigger}
 


pdonahue-ventana (approved):
This looks sensible to me.

pdonahue-ventana:
@timsifive: I normally let you merge things.  This is basically a clarification about how the new Zawrs extension fits in (WRS.* are the same as WFI except for an additional wakeup mechanism or two).  This is how you'd naturally want to do it.

Sdext.tex
```diff
@@ -36,7 +36,7 @@ \section{Debug Mode} \label{debugmode}
 \item If \FcsrDcsrStoptime is 0 then \Rtime continues to update. If
     it is 1 then \Rtime will not update. It will resynchronize with
     \Rmtime after leaving Debug Mode.
-\item The {\tt wfi} instruction acts as a {\tt nop}.
+\item The {\tt wfi}, {\tt wrs.sto}, {\tt wrs.nto} instructions act as a {\tt nop}.
```
timsifive, line 39:
```suggestion
\item Instructions that place the hart into a stalled state act as a {\tt nop}.
    This includes {\tt wfi}, {\tt wrs.sto}, and {\tt wrs.nto}).
```

timsifive (changes requested):
Phrase this in terms of the rule to follow and then specific examples, which hopefully is more future proof.

Sdext.tex
ved-rivos: Updated in this commit:
https://github.com/riscv/riscv-debug-spec/pull/838/commits/7cddfb3769a47bd0c9ea6158749832e3bfd75e78

pdonahue-ventana (commented):
I like that this is written more generically.

This now seems to include the pause instruction from Zihintpause.  It makes me realize that the purpose of this clause is about functionality rather than performance.  If wfi stalls/sleeps then there's probably no way to wake it up which would be very bad.  If the pause instruction or even wrs.sto stalls/sleeps then we know at some point it is guaranteed to unstall and keep going.

So do we need to include wrs.sto here?  An implementation can stall for a finite time for any reason at all so even if an implementation doesn't treat wrs.sto as a nop, a debugger can't really distinguish whether it violated this clause or if it happened to stall for some other reason.  If there's really no way to verify compliance then the requirement doesn't seem useful.

If, on the other hand, we think that wrs.sto should stay then should we also include pause here?  We might as well include all known examples at this time.

timsifive:
I agree with @pdonahue-ventana. `wrs.sto` is defined as being "bounded by an implementation defined short timeout." Such a short timeout seems fine when responding to a halt. Probably the generic language should say something like "Instructions that may place the hart into a stalled state for more than one second act as a {\tt nop}."

This is a corollary to existing text in the debug spec which says "When halt or resume is requested, a hart must respond in less than one second."

pdonahue-ventana:
@ved-rivos: Thoughts?

ved-rivos:
Agree. But we should also allow for the case where a stalled instruction actually does want to be nop. For example, for many implementations, like one I am working on - its more work to ignore the halt request depending on type of stall than to act as nop when there is a halt requested. 
So how about?
"Instructions that may place the hart into a stalled state for more than one second MUST act as a {\tt nop} and those that stall for less than one second MAY act as a {\tt nop}."

pdonahue-ventana:
I agree with your point in theory but aren't PAUSE and WRS.STO (and WFI and WRS.NTO) always allowed to act as NOPs even in non-debug mode?  (More specifically, they can wake up for any reason at all which includes waking up immediately without ever having gone to sleep.)

ved-rivos:
Agree with that reasoning. We could just drop the `wrs.sto` from the enumerated list - similar to how `pause` is not in that list.

ved-rivos:
I think the specification makes a subtle differentiation between PAUSE vs. WFI/WRS.*. While for PAUSE it states "rate of instruction retirement should be temporarily reduced or paused", for WFI/WRS.* it states "current hart can be stalled". So a pause only controls the instruction retirement rate but does not actually stall execution if we interpret it that way. But it may be reading too much into the english text... 
The only reason I suggested not invoking "longer than 1 second" was to avoid bringing a magic number into the spec. Enumerating only wfi and wrs.nto may then suffice to be clear that other instructions like wrs.sto and pause are not covered by this rule.

pdonahue-ventana:
It depends on the definition of stall.  Of course, things stall all the time for a few cycles but that's obviously not what we're talking about here.  The PAUSE instruction seems to fall into that situation of stalling for a few tens or maybe hundreds of cycles but not anything noticeable at a human scale like this one second thing we're talking about (and if it does then debug mode is the least of your problems).  So I'm OK with leaving PAUSE off the list since it's just a NOP that's a little bit slow.

WRS.STO essentially is already defined to be a NOP (though one that's much slower than PAUSE) which might take a long, though guaranteed finite, amount of time to execute.  In contrast, if you disable interrupts and never lose the reservation then WFI and WRS.NTO might take an infinite time to execute (which is why they must be listed).

I'm OK with removing WRS.STO but I'm also OK with leaving it there (and not making any changes to this PR at all).  Since WRS.STO is already a NOP, if it stays in the list then implementations can either never go to sleep (a fast NOP) or wait for the timer to expire (a slow NOP).  If it's not in the list then implementations also can either never go to sleep (which is always a valid WRS.STO behavior) or wait for the timer to expire.

ved-rivos:
So how do we proceed?

timsifive:
I think our options are:
1. Leave as-is. `wrs.sto` will have to be a nop. Downside is that this might slightly complicate an implementation that has to give `wrs.sto` different behavior in Debug Mode.
2. Rewrite the generic language in some way that allows short (how short?) stalls in Debug Mode.

I'm OK with either one. I don't know a great way to phrase option 2, and the downside of option 1 seems pretty minor. So that's my preference.

pdonahue-ventana:
Let's go with option 1.  I already approved this PR and I'll stand by that.

timsifive (approved):

PR #839, opened 2023-04-18
Reentrant trigger prevention inhibits match *or fire*

timsifive:
This is a substantive change, but a bugfix, so OK to merge.

Addresses #812.


pdonahue-ventana (approved):
Looks good.

Issue #840, opened 2023-04-19
Software control of mcontrol6.hit0 and mcontrol6.hit1 and clearing of mcontrol6.hit0 and mcontrol6.hit1 


Silabs-ArjanB:
Before introduction of the hit0 and hit1 bits in mcontrol6 the 'old style' mcontrol.hit had a description allowing user control of the hit bit. Similarly other 'hit' bits (e.g. legacy mcontrol.hit and icount.hit) have descriptions as the following, highlighting that the trigger's user (I assume that means software or debugger) can set/clear hit as well:

> If this bit is implemented then it must become set
> when this trigger fires and may become set when
> this trigger matches. The trigger’s user can set or
> clear it at any time.

This 'the trigger’s user can set or clear it at any time' phrase is no longer part of the mcontrol6 hit0/hit1 descriptions. Instead there is now the encoding '0 (false) the trigger did not fire'. That seems quite ambiguous as I would expect that once a trigger fires that its hit0/hit1 info would remain available until either software clears it or that same trigger fires with a different timing reason.

In particular the 'trigger did not fire' condition will typically hold let's say 99.9% of the time (as normally triggers won't fire) and therefore the hit0/hit1 field would end up as 0 before the debugger can inspect it.

Example: We have 4 mcontrol6 triggers set up for load address matching. Initially for some reasons has hit* bitfields of these 4 triggers are set to 3. Now suppose there is a load that does not match any of the triggers. I would expect that all hit* bitfields remain set to 3 (as opposed to all of them getting set to 0 as the Debug specification seems to imply). 

Now, assuming that again the 4 mcontrol6 bitfields are initially 3 and suppose that only instance 0 has a match with before timing and fires for some load instruction and the other 3 triggers do neither match nor fire. 

Now I would think that there are two sensible outcomes:

1) Set hit* of instance 0 to 1 and leave the hit* bitfields of the other 3 instances unchanged (so value 3).
2) Set hit* of instance 0 to 1 and set the hit* bitfields of the other 3 instances to value 0 (indicating that some trigger fired, but not these ones).

I also would assume that normally the debugger software would clear all hit* bitfields to 0 after having handled an initial firing, right (assuming 'firing into debug mode)?

Could you please clarify the hit* operation a bit further and especially whether hardware and/or software is reponsible and/or able to clear the hit* bitfield?

Best regards,
Arjan

pdonahue-ventana:
[...]

PR #841, opened 2023-04-28
Clarify when hit0/hit1 are updated by TM.

timsifive:
Addresses #840.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -727,9 +727,11 @@
             virtualization mode.
         </field>
         <field name="hit0" bits="22" access="WARL" reset="0">
-            It they are implemented, \FcsrMcontrolSixHitOne (MSB) and
-            \FcsrMcontrolSixHitZero (LSB) combine into a single 2-bit field that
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -728,8 +728,8 @@
         </field>
         <field name="hit0" bits="22" access="WARL" reset="0">
             It they are implemented, \FcsrMcontrolSixHitOne (MSB) and
```
pdonahue-ventana, line 730:
```suggestion
            If they are implemented, \FcsrMcontrolSixHitOne (MSB) and
```
I didn't notice that previously.

pdonahue-ventana (commented):
Strictly speaking, the value 1 (or 2 or 3) doesn't mean that the trigger fired.  It means that either the trigger fired or the debugger wrote a 1.  But debuggers shouldn't really be doing that.  Maybe a little more explanation such as "The debugger will normally write 0 to this field after it observes the trigger having fired.  If the field becomes non-zero later then the trigger must have fired."

pdonahue-ventana (approved):
OK

Issue #842, opened 2023-05-03
Does icount trigger match on entering Debug Mode due to triggers?


YenHaoChen:
The spec says "This (icount) trigger matches on every instruction completed or trap taken ..."
![image](https://user-images.githubusercontent.com/39526191/235826646-38168239-f46d-478c-93f5-d52afcb5fba6.png)
To my reading, the icount will decrease on ebreak instructions but will not decrease on entering Debug Mode.

I want to clarify the following conditions.
1. Does the icount decrease on firing trigger action=breakpoint when icount.count=1?
2. Does the icount decrease on firing trigger action=debug_mode when icount.count=1?

timsifive:
> To my reading, the icount will decrease on ebreak instructions but will not decrease on entering Debug Mode.

Correct. Maybe the spec should be more clear, but in Debug Mode triggers are completely inactive.

icount matches "on every instruction completed or trap taken" and a trap refers to "the transfer of control to a trap handler caused by either an exception or an interrupt" (unpriv spec section 1.6).

> 1. Does the icount decrease on firing trigger action=breakpoint when icount.count=1?

Yes. action=breakpoint explicitly raises "a breakpoint exception" which then results in a trap taken. (I don't think you can mask the breakpoint exception.)

> 2. Does the icount decrease on firing trigger action=debug_mode when icount.count=1?

Only if the instruction is executed. Entering Debug Mode is not a trap, even though in systems with a Debug ROM it looks a lot like it. Debug Spec section A.2:

"When the halt request bit is set, the Debug Module raises a special interrupt to the selected harts. This interrupt causes each hart to enter Debug Mode and jump to a defined memory region that is serviced by the DM and is only accessible to the harts in Debug Mode. When taking this jump, pc is saved to dpc and cause is updated in dcsr. This jump is similar to a trap but it is not architecturally considered a trap, so for instance doesn’t count as a trap for trigger behavior."

YenHaoChen:
Thank you

YenHaoChen:
Another quick question. Does the icount trigger decrease on NMI?

timsifive:
I don't think the spec says. My inclination is to say "no" since taking an NMI doesn't execute an instruction (except the ones in the trap handler, which will be counted depending on the modes where icount is enabled).

I'll make a PR and we'll see what people think.

timsifive:
[...]

PR #843, opened 2023-05-04
Clarify that icount decrements on interrupt traps.

timsifive:
The rationale is that if icount is enabled in the current mode and the mode of the trap handler, when you single step you don't want to step past the first instruction of the trap handler, which you didn't know was going to be executed. Having to step multiple times to get past an interrupt to actually execute the instruction you tended to step is addressed the same way as already documented in the appendix.
    
Addresses a question from #842.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -994,6 +994,7 @@
                 mode where the trigger is enabled.
                 This explicitly includes {\em x}\/RET instructions.
             \item A trap is taken from a privilege mode where the trigger is enabled.
+                This explicitly includes traps taken due to interrupts.
         \end{steps}
 
         If more than one of the above events occur during a single instruction


pdonahue-ventana (commented):
I didn't think that this is true.  There was a lot of discussion in https://lists.riscv.org/g/tech-debug/topic/96011849 about how interrupts DO cause icount to decrement, how that can be a problem, and how the debugger is supposed to detect that and do an extra step.

timsifive:
There was a lot of discussion about exceptions, and interrupts were kind of hand-waved in there. The end result of all that discussion was https://github.com/riscv/riscv-debug-spec/pull/798, which does list interrupts as a case where icount decrements. Note that it's in the appendix, not the spec itself.

So the spec itself does not state whether interrupts decrement icount. We need a change that clarifies, one way or the other.

Here's Beeman laying out the argument for decrementing icount on interrupts: https://docs.google.com/spreadsheets/d/1o_pZ0MMqCkbsAvFpqdmdkl1fRPYOzR6QFXuYz5ky2uo/edit?disco=AAAAm3byacU. The idea is that if icount is enabled in the mode the trap is enabled in, then you do want to decrement icount when an interrupt is taken so that a single step doesn't step past the first instruction in that trap handler.

I've reasoned my way over to your position. I'll change the PR to reflect that.

pdonahue-ventana (approved):
The unpriv spec says:
> We use the term trap to refer to the transfer of control to a trap handler caused by either an
exception or an interrupt.

So this bullet always included traps due to interrupts.  But most people don't remember definitions from the introduction chapter of another spec so I agree that it's better to be explicit.

YenHaoChen:
Thank you for the clarification. (I thought interrupt traps were included from my reading on the current spec.)
My original question was about the NMI. I assume the icount also decreases on NMI since the NMI is an interrupt.

pdonahue-ventana:
> Thank you for the clarification. (I thought interrupt traps were included from my reading on the current spec.) My original question was about the NMI. I assume the icount also decreases on NMI since the NMI is an interrupt.

I agree.  For instance, the Smrnmi spec (in the priv spec) mentions the "the NMI trap handler" so NMIs are traps which are handled by a trap handler.

PR #846, opened 2023-05-10
Add dtmcs.errinfo.

timsifive:
This is specifically to handle DMs that power down, which could be handled very differently by a debugger than an actual failure talking to a powered up DM. In the former case report to the user and wait until it's back. In the latter report to the user and abort.

This is certainly a substantive (albeit backwards compatible) change. Can it go into 1.0?


xml/jtag_registers.xml
```diff
@@ -26,7 +26,29 @@
...
+
+            <value v="1" name="dmipower">
+                The error occurred because the DMI interface is powered down.
+            </value>
+
+            <value v="2" name="dmpower">
+                The error occurred because the DM is powered down.
+            </value>
+
+            TODO: Other errors? Parity?
```
ernie-sifive, line 48: Another useful code would be that an error was reported from the DM.  An example might be if a debugger attempted to access a non-existent DMI register, or if a DMI request or response was corrupted (e.g. parity error).

sifiverobert:
What is DM is powered-up, but not clocked? I think **DM is not accessible** would be more generic description.
We should state when this bit is cleared. As error in DM is 'latched' it should be cleared at 'dmreset' (as DM error-state is).
I also do not see how DMI powered-down may respond to JTAG scan (of 'dtmcs' ...).
And single DMI may handle several DMs, so this error should handle ALL DMs connected via this DMI. But then we are getting into different DMs signalling different reasons. Maybe then single bit (or bits ...) being logical-or of all DMs would be enough?
As far as down-compatibility, some (already released ...) debuggers may not like reading non-0 in this field.
So, we should have a bit, which is enabling non-0 reporting of detailed error. If it is never set (what old debuggers will not do ...), it will remain 0.

fborisovskii:
Hello,
May I join? Interesting question by Robert. I thought using multiple DMs with one DMI is possible only if all DMs are accessible.
Because nextdm is a DM register. Is it really possible to power down one of many DMs controlled by one DMI?
Is it possible to reconnect to this system? Seems it becomes undiscoverable. There is op == 2 in DMI read that means DM reports error. Does DM must report error on DMI if it is powered off? Or it is supposed to be busy error because there is no clock and Clock domain crossing can't be done?

xml/jtag_registers.xml
timsifive: Added. Thanks.

timsifive:
> What is DM is powered-up, but not clocked? I think DM is not accessible would be more generic description.

Changed. Thanks.


> We should state when this bit is cleared. As error in DM is 'latched' it should be cleared at 'dmreset' (as DM error-state is).

It already says the field "is updated whenever \FdtmDmiOp is updated by the hardware."

> I also do not see how DMI powered-down may respond to JTAG scan (of 'dtmcs' ...).

dtmcs is on the TAP controller.

> And single DMI may handle several DMs, so this error should handle ALL DMs connected via this DMI.

I disagree. This field simply gives more information about why a given access failed (if it did fail). There is no confusion 

> But then we are getting into different DMs signalling different reasons. Maybe then single bit (or bits ...) being logical-or of all DMs would be enough?

This is not an issue if errinfo only provides information for the latest access.

> As far as down-compatibility, some (already released ...) debuggers may not like reading non-0 in this field.

Then they are non-compliant. Section 1.3.3 says "Parts of the register which are currently unused are labeled with the number 0. Software must only write 0 to those fields, and ignore their value while reading. Hardware must return 0 when those fields are read, and ignore the value written to them."

timsifive:
@fborisovskii, you raise some good points, but none of these really pertain to this specific PR. It's best to discuss those on the mailing list, or by creating a new PR.

markuslauterbach:
> value v="3" name="dm error"
>                 The DM reported an error, e.g. because a non-existent register was accessed.

Shouldn't an access to a non-existent debug register succeed and return 0? Mentioning non-existent registers as an example sounds as if we are implying that it is normal that an access to a non-existent debug register throws an error.

timsifive:
@markuslauterbach, you are right.

"When read, unimplemented or non-existent Debug Module DMI Registers return 0. Writing them has no effect."

Should "dm error" then be "bus error" instead? If there's a parity error somewhere I'd expect the bus to check/report that, not the DM. Do I misunderstand how hardware works? I can't think of a case where the DM would report an error.

pdonahue-ventana:
I think that other devices besides DMs are allowed to live on DMI (as long as they're not addresses 0x00-0x7f or within 0x00-0x7f of any nextdm value).  Those other devices could presumably return errors even if a DM cannot.

It seems like there are 3 possible errors:
1. DTM has some problem communicating with DMI
2. DMI has some problem communicating with a device (such as DM)
3. The device received the transaction but wants to return an error (for a device other than a DM since they can't return errors).

Cases 1,2,3 map to errinfo values 1,2,3.  If DMI is AXI then I think that DECERR maps to 2 and SLVERR maps to 3.  The descriptions of those values would have to change to allow that, though.

I'm not entirely sure how errinfo values 5 and 6 are expected to be used.  Is the idea that you try a transaction, hardware sends an error response and kicks off some state machine to fix the problem (e.g. powering something back up) after sending the error?  Why not just have hardware hold off on responding until the state machine is done (since responses are allowed to have variable latency)?  That would use the existing op=busy mechanism to tell the debugger to retry rather than a new op=failed + errinfo=temporary mechanism.

sifiverobert:
Thanks @timsifive - I agree with your response to my earlier post.

Two points A) and B) ...

A) I would elaborate more about 'errinfo' being 'sticky' because 'op' is sticky.

Consider the following (executing transaction is a batch is useful ...):

1. You execute transaction which fails and 'errinfo' is set.
2. You execute another transaction (which is ignored as error in 'op' is sticky).
3. What is value of 'errinfo' (unchanged from 1 as 'op' is same?).

B) Adding WARL 'enable' bit would allow debugger to check if 'errinfo' is implemented or not. If 0 is reported we do not know if this is because details are unknown or 'errinfo' is not implemented. Maybe 'errinfo=0' should mean 'not implemented in HW'?

Thanks
-Robert

markuslauterbach:
> I think that other devices besides DMs are allowed to live on DMI (as long as they're not addresses 0x00-0x7f or within 0x00-0x7f of any nextdm value). Those other devices could presumably return errors even if a DM cannot.

Good point. For other non-DM devices on DMI, a bus error may of course be legal.

> Should "dm error" then be "bus error" instead?

Yeah maybe that is better. In general, given that DMI can not only address DMs, the error codes should not (only) be specific to DMs, but about any devices behind DMI in general?

timsifive:
> I'm not entirely sure how errinfo values 5 and 6 are expected to be used. Is the idea that you try a transaction, hardware sends an error response and kicks off some state machine to fix the problem (e.g. powering something back up) after sending the error? Why not just have hardware hold off on responding until the state machine is done (since responses are allowed to have variable latency)?

This is my attempt to have states that are useful when a DM is powered down (which is kind of intentional, and might be resolved if you try again). In that case what you want is for the debugger to tell the user "this thing failed, but I'm going to keep trying until you tell me otherwise" which is different from "this thing failed, and it should not have failed, and the house of cards is coming down."

I'm open to other phrasing etc. I'd just like a difference between debugger-can-die and debugger-should-keep-retrying.

timsifive:
I've rephrased some of this based on @pdonahue-ventana's comments, and addressed @sifiverobert's point B.

Still need to think about 5 and 6.

timsifive:
I've removed the "temporary" states. The only example I can think of is regarding power down, but I don't want to acknowledge anywhere in the spec that that might happen, because it shouldn't.

When a debugger sees an error, it will have to decide by itself whether to keep trying or whether to fail. The safe behavior is to keep trying, while simultaneously notifying the user, giving them control over when to stop trying.

I still think there is value in this change because it gives more information about what goes wrong.

timsifive:
@pdonahue-ventana, are you able to rereview this? And do you have thoughts on whether this goes into 1.0 or into a future version?

pdonahue-ventana (commented):
Reading it again, I have some new thoughts.

The reset value is 0.  That means that the field is not implemented.  I don't think that's necessarily right.

It's not entirely clear whether or how dmireset affects this field.  When dmireset is written with a 1, does dmi.op get updated by hardware or is that update done by software/debugger?  I could envision two approaches.  One is that this field becomes 4 whenever we do a dmireset.  The other is that this field is only valid if op=2 so there's no need to ever clear this field (and probably no need to have encoding 4) since the next time hardware writes 2 it will update this field.

My preference is to say that it resets to 4 or 0 depending on whether it's implemented or not and that dmireset causes this to become the reset value (which should be explicitly stated here and probably also in the dmireset description).

As for the 1.0 vs. future question, I'm a little torn.  It seems straightforward, backward-compatible, possibly helpful and at least not harmful.  On the other hand, it's getting late and if there are any weirdnesses then we might not know about them until after it's baked into the spec.  Overall, I'm OK with it being in 1.0, though.

timsifive:
@pdonahue-ventana I agree about reset. I've updated the language.

> As for the 1.0 vs. future question, I'm a little torn. It seems straightforward, backward-compatible, possibly helpful and at least not harmful. On the other hand, it's getting late and if there are any weirdnesses then we might not know about them until after it's baked into the spec. Overall, I'm OK with it being in 1.0, though.

Worst case, we merge it and immediately AR approves the spec, and we freeze right then. Then there's still the public review period where we could back this out if necessary.

pdonahue-ventana (approved):
OK.  We can go with this.  And it's fine for 1.0 since AR hasn't started yet.

PR #847, opened 2023-05-10
D-Mode exceptions may modify DPC

timsifive:
Because the definition of DPC already says "Executing the Program Buffer may cause the value of dpc to become unspecified."

This fixes a contradiction in the spec.

### DIFF
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -28,7 +28,7 @@ How Debug Mode is implemented is not specified here.
     according to \FdmAbstractcsRelaxedpriv.
 \item All interrupts (including NMI) are masked.
 \item Exceptions don't update any registers.  That includes {\tt cause}, {\tt
-    epc}, {\tt tval}, {\tt dpc}, and \Rmstatus. They do end execution of the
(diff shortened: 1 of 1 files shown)


pdonahue-ventana (commented):
I haven't thought about this before but how do you get dpc out?  You're only guaranteed to have abstract access to the integer registers.  If you want access to CSR X then you normally use the PB to read X into (for instance) x1 and then transfer x1 out.  If you need to use the PB to read dpc and executing the PB causes dpc to be trashed (possibly before the CSR read) then there's a problem.

timsifive:
I think this is already accounted for. 4.9.2:

> Executing the Program Buffer may cause the value of dpc to become unspecified. *If that is the
> case, it must be possible to read/write dpc using an abstract command with postexec not set.* The
> debugger must attempt to save dpc between halting and executing a Program Buffer, and then
> restore dpc before leaving Debug Mode.

pdonahue-ventana:
OK.  That's what I get for not reading the whole thing (and not remembering all the nuances).

pdonahue-ventana (approved):

PR #848, opened 2023-05-16
Clarify trap in progbuf behavior.

timsifive:
Allow partial register updates for instructions that already do that.

Based on lots of mailing list discussion:
https://lists.riscv.org/g/tech-debug/topic/98622044#1178

### DIFF
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -21,15 +21,23 @@ How Debug Mode is implemented is not specified here.
 
 \begin{steps}{When executing code due to an abstract command, the hart stays in
     Debug Mode and the following apply:}
+\item All implemented instructions operate just as they do in M-mode, unless
+    an exception is mentioned in this section.
 \item All operations are executed with machine mode privilege, except that
     additional Debug Mode CSRs are accessible and
     \FcsrMstatusMprv in \Rmstatus may be ignored according to \FcsrDcsrMprven.
     Full permission checks, or a relaxed set of permission checks, will apply
     according to \FdmAbstractcsRelaxedpriv.
 \item All interrupts (including NMI) are masked.
-\item Exceptions don't update any registers.  That includes {\tt cause}, {\tt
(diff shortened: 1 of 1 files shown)


Sdext.tex
```diff
@@ -21,15 +21,23 @@ \section{Debug Mode} \label{debugmode}
...
     additional Debug Mode CSRs are accessible and
     \FcsrMstatusMprv in \Rmstatus may be ignored according to \FcsrDcsrMprven.
     Full permission checks, or a relaxed set of permission checks, will apply
     according to \FdmAbstractcsRelaxedpriv.
 \item All interrupts (including NMI) are masked.
-\item Exceptions don't update any registers.  That includes {\tt cause}, {\tt
-    epc}, {\tt tval}, and \Rmstatus. They do end execution of the
-    Program Buffer.
+\item Traps don't take place. Instead, they end execution of the program buffer
+    and the hart remains in debug mode. Because they do not trap to M mode, they
```
pdonahue-ventana, line 33:
```suggestion
    and the hart remains in Debug Mode. Because they do not trap to M-mode, they
```
We're not entirely consistent with how these things are done, but it appears that in most places Debug Mode is capitalized and M-mode is hyphenated and not capitalized (matching what's in the priv spec).

Sdext.tex
```diff
@@ -21,15 +21,23 @@ \section{Debug Mode} \label{debugmode}
 
 \begin{steps}{When executing code due to an abstract command, the hart stays in
     Debug Mode and the following apply:}
+\item All implemented instructions operate just as they do in M-Mode, unless
```
pdonahue-ventana, line 24:
```suggestion
\item All implemented instructions operate just as they do in M-mode, unless
```
To be consistent with the other bullet below (and the capitalization in the priv spec).

pdonahue-ventana (approved):

PR #850, opened 2023-06-12
External trigger signals inputs don't cause trigger output.

timsifive:
This fixes a bug introduced in #506.

See e-mail: https://lists.riscv.org/g/tech-debug/topic/99398782#1291


pdonahue-ventana (approved):
This seems right.

Issue #851, opened 2023-06-22
clarification on how execution from Program Buffer affects triggers


aap-sc:
Hi!

Chapter 4 "Debug Mode" says the following:

```
When executing code due to an abstract command, the hart stays in Debug Mode and the following
apply:
1. All operations are executed with machine mode privilege, except that additional Debug Mode
CSRs are accessible and MPRV in mstatus may be ignored according to mprven. Full per-
mission checks, or a relaxed set of permission checks, will apply according to relaxedpriv.
2. All interrupts (including NMI) are masked.
3. Exceptions don’t update any registers. That includes cause, epc, tval, dpc, and mstatus.
They do end execution of the Program Buffer.
4. No action is taken if a trigger matches.
...
```

Could someone clarify on the bullet number "4"? Which action the specification is talking about? Is it that the action defined by the trigger discarded/ignored? Or does the spec implies that the program in Program Buffer stops executing/aborts?

To be more specific, consider the following case:
1. Imagine a data breakpoint (aka watchpoint) being set at some address using trigger interface.
2. A hart is in debug state. Now we have a program in the Program Buffer that access the address (via load/store operation) that matches the watchpoint.
3. Will the load/store be completed successfully?

pdonahue-ventana:
The word "action" has the same meaning as every other use of that word in the Sdext and Sdtrig chapters.  It is the thing that is encoded in the action field of the trigger as described in the "Actions" section.  For instance, if you have a trigger on a load address (as in your example) and the program buffer does a load from that address then the action (halt, trap, turn trace on, etc.) doesn't happen.  Since there is no halt or trap, it just does the load.  To act otherwise would be very burdensome for a debugger since it might have to disable all of the triggers before doing anything and then reenable them before resuming.

timsifive:
Hmm... So does the hit bits get set when a trigger matches even though the action isn't taken? I suspect this line predates `hit` bits. Really it should say triggers don't match or fire, which is more current language.

aap-sc:
@pdonahue-ventana , @timsifive thanks for the clarification! Resolving the issue.

PR #852, opened 2023-06-23
In Debug Mode, triggers don't match or fire.

timsifive:
The existing language implied that hit bits would be updated, which doesn't make any sense. That's because that line was written before we had hit bits.

Clarifies #851.


pdonahue-ventana (approved):
OK.  Just 3 hours ago I told Greg that no further spec changes were pending.  This clarification seems valuable and I assume this won't set us back since they haven't started reviewing it yet.

PR #855, opened 2023-07-08
Add support for Zicfilp and the hart ELP state

sorear:
[Zicfilp](https://github.com/riscv/riscv-cfi/blob/main/cfi_forward.adoc) adds a new ELP state to each hart, which is handled like the privilege mode: it affects all instructions, is not visible in any CSR, but is saved into CSR fields by traps and restored by trap returns.

Add it to the debug spec in the same fashion as the privilege mode, and specify that it does not affect execution in Debug Mode.

(Draft because Zicfilp is not frozen.)

### DIFF
--- a/implementations.tex
+++ b/implementations.tex
@@ -67,7 +67,7 @@ To resume execution, the debug module sets a flag which causes the hart to execu
 not executing from the Program Buffer.
 Its recommended encoding is 0x7b200073.
 When {\tt dret} is executed, \Rpc is restored from \RcsrDpc and normal execution resumes at the
-privilege set by \FcsrDcsrPrv and \FcsrDcsrV.
+privilege set by \FcsrDcsrPrv and \FcsrDcsrV and the ELP state set by \FcsrDcsrElp.
 
 \RdmDataZero etc. are mapped into regular memory at an address relative to \Rzero
 with only a 12-bit {\tt imm}. The exact address is an implementation
--- a/xml/sw_registers.xml
+++ b/xml/sw_registers.xml
@@ -12,8 +12,9 @@
         Users can write this register to change the privilege mode that
         the hart will run in when it resumes.
 
-        This register contains \FcsrDcsrPrv and \FcsrDcsrV from \RcsrDcsr, but in a place that the user
-        is expected to access. The user should not access \RcsrDcsr directly,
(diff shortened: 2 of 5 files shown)


sorear:
@ved-rivos

implementations.tex
```diff
@@ -67,7 +67,7 @@ \section{Execution Based} \label{execution_based}
 not executing from the Program Buffer.
 Its recommended encoding is 0x7b200073.
 When {\tt dret} is executed, \Rpc is restored from \RcsrDpc and normal execution resumes at the
-privilege set by \FcsrDcsrPrv and \FcsrDcsrV.
+privilege set by \FcsrDcsrPrv and \FcsrDcsrV and the ELP state set by \FcsrDcsrElp.
```
timsifive, line 70:
```suggestion
privilege set by \FcsrDcsrPrv and \FcsrDcsrV, and the ELP state set by \FcsrDcsrElp.
```

xml/sw_registers.xml
```diff
@@ -38,6 +39,28 @@
...
+        \begin{tabular}{|c|l|}
+        \hline
+        elp &amp; Name \\
+        \hline
+        0 &amp; {\tt NO\_LP\_EXPECTED} \\
+        \hline
+        1 &amp; {\tt LP\_EXPECTED} \\
+        \hline
+        \end{tabular}
+        \end{table}
```
timsifive, line 55: I think I prefer these values inlined in `dcsr` and `priv` using the XML value tags. `v` is a bit different since a single table contains `v`, `prv`, and what they mean depending on whether H is supported.

timsifive (changes requested):
Overall looks good and thorough.

I wonder if we should rename `priv` to reflect its more general use now, or if we should leave it to not make any unnecessary waves.

xml/sw_registers.xml
ved-rivos: Could also just reference Zicfilp for the syntax  and semantics of `ELP`

sorear: If it's literally just a gdb user interface thing then we might as well split out `$elp` from `$priv` but I wasn't sure how much of the stack splitting the two would touch and whether the new register might cause problems with debug middleware that doesn't otherwise need changes.

ved-rivos:
@rtwfroody - this PR can be closed without merge as the #1071 that integrates the Zicfilp extension is merged.

PR #858, opened 2023-07-19
Add register printers to debug_defines.h

en-sc:
This commit implements facilities to decode register fields for each register described in debug_defines.h.

Namely:

define <REG_NAME>_FORMAT_STR
macro <REG_NAME>_FORMAT_ARGS(value)

The intended use case:
```
reg_type reg_value = ...;
printf(... <REG_NAME>_FORMAT_STR ...,
		..., <REG_NAME>_FORMAT_ARGS(reg_value), ...);
```


timsifive:
For some simple and complex registers, can you show what ends up being generated by this change?

timsifive:
Never mind, I see it in https://github.com/riscv/riscv-openocd/pull/892/files

registers.py
```diff
@@ -299,18 +299,20 @@ def sympy_to_c(expression):
         else:
             return "-0x%xULL" % -expression
     elif isinstance(expression, sympy.Symbol):
-        return str(expression)
+        return "(" + str(expression) + ")"
```
timsifive, line 302: Good catch.

registers.py
```diff
@@ -320,10 +322,15 @@ def write_cheader( fd, registers ):
         try:
             if r.width() <= 32:
                 suffix = "U"
+                format_specifyer = "PRIx32"
```
timsifive, line 325:
```suggestion
                format_specifier = "PRIx32"
```

registers.py
```diff
@@ -350,6 +357,22 @@ def write_cheader( fd, registers ):
...
+                field_value_str = f"((value) & {sympy_to_c(mask.body())}) >> {sympy_to_c(offset.body())}"
+                if len(f.values):
+                    format_string += [f"\"{f.name}=%s\""]
+                    get_name_from_value = [
+                            f"({field_value_str} == {str_or_sympy_to_c(v.value)}) ? \"{v.name}\" :"
+                            for v in f.values if v.value
+                            ] + ["\"unknown\""]
+                    format_args.append(("\\\n\t".join(get_name_from_value)))
+                else:
+                    format_string.append((f"\"{f.name}=0x%\" {format_specifyer}"))
```
timsifive, line 369: If the field is 3 bits wide or less, can you omit "0x" in the format?

registers.py
```diff
@@ -350,6 +357,22 @@ def write_cheader( fd, registers ):
                 for v in f.values:
                     definitions += v.to_c_definitions(prefix)
 
+                field_value_str = f"((value) & {sympy_to_c(mask.body())}) >> {sympy_to_c(offset.body())}"
+                if len(f.values):
+                    format_string += [f"\"{f.name}=%s\""]
+                    get_name_from_value = [
+                            f"({field_value_str} == {str_or_sympy_to_c(v.value)}) ? \"{v.name}\" :"
+                            for v in f.values if v.value
+                            ] + ["\"unknown\""]
```
timsifive, line 366:
```suggestion
                            ] + ['"unknown"']
```

registers.py
```diff
@@ -350,6 +357,22 @@ def write_cheader( fd, registers ):
                 for v in f.values:
                     definitions += v.to_c_definitions(prefix)
 
+                field_value_str = f"((value) & {sympy_to_c(mask.body())}) >> {sympy_to_c(offset.body())}"
+                if len(f.values):
+                    format_string += [f"\"{f.name}=%s\""]
```
timsifive, line 362:
```suggestion
                    format_string += [f'"{f.name}=%s"']
```

registers.py
timsifive: Ideally we'd show the numeric value instead of "unknown". I don't know how to do that reasonably inside a macro.

timsifive (commented):
I love the idea behind this change. Great use of having all this data in machine-readable format.

Using macros for everything is a little limiting. E.g. we have to show "unknown" instead of the value, and I guess checkpatch doesn't like that `value` is evaluated multiple times.

Would it make sense to generate static inline functions instead, with something like:
```
int ac_access_register_format(char *str, size_t size, uint32_t value) {
   return snprintf(str, size, FORMAT, ARGS);
}
```

JanMatCodasip:
Overall, I like the idea of the auto-generated printers.

> Tim: ... Would it make sense to generate static inline functions instead ...

I support this. For code cleanliness and proper encapsulation, I recommend to use functions that would contain handling of both the printf format string and the arguments. I am not too happy to see these complex expressions, even multiple function arguments, exposed to the user in form of macros. 

I assume that the reason for the use of macros was the flexibility - to allow putting the text into an arbitrary point in the message, along with other printf arguments.

The question is whether we actually need that degree of flexibility. Maybe it would be sufficient to have functions that just dump the register values as a debug print (`riscv_print_tadata1(target, tdata1_value)  --> calls LOG_TARGET_DEBUG()`). And any user message can go to a separate log print.

timsifive:
> Maybe it would be sufficient to have functions that just dump the register values as a debug print (riscv_print_tadata1(target, tdata1_value)  --> calls LOG_TARGET_DEBUG())? And any user message can be printed as a separate log print.

Note that this PR is in the debug spec, not in OpenOCD source. So we shouldn't limit these functions to something that only works in OpenOCD.

JanMatCodasip:
> > Maybe it would be sufficient to have functions that just dump the register values as a debug print (riscv_print_tadata1(target, tdata1_value)  --> calls LOG_TARGET_DEBUG())? And any user message can be printed as a separate log print.
> 
> Note that this PR is in the debug spec, not in OpenOCD source. So we shouldn't limit these functions to something that only works in OpenOCD.

Good point. If we are looking for a more generic solution, then you may consider dynamic allocation of a character buffer, and passing the resulting string in this form. The calling application can then use it as it pleases (and is responsible for freeing it).

JanMatCodasip:
> I prefer having the caller pass in the buffer and the length ...

Sounds fine. I am in for any solution that has the string composition encapsulated inside functions with a clean and easy to use API.

registers.py
en-sc: Thanks!

registers.py
en-sc: Addressed here: https://github.com/riscv/riscv-openocd/pull/892/files#diff-4187474b258a4ed647428725e3d087bc7f60b0e8a2648270beca34dcaa76498cR26

registers.py
en-sc: Addressed.
The issue with unknown values is addressed here: https://github.com/riscv/riscv-openocd/pull/892/files#diff-4187474b258a4ed647428725e3d087bc7f60b0e8a2648270beca34dcaa76498cR40

registers.py
en-sc: Addressed.

en-sc:
@timsifive, @JanMatCodasip, I have changed the approach. Please, take a look.

JanMatCodasip (changes requested):
I've done only superficial review so far. I think the overall idea is good. Some aspects of this should probably be improved.

1) Does it make sense to couple the register decoding functions to debug_defines.h or to put it into another C module (another pair of .h & .c files)? It is possible that some users are interested only in the constants but do not wish to bring the register decoding functions into their project. You may consider splitting it. And opinions of others are welcome :)

2) Why to create new enum in debug_defines.h - `riscv_debug_reg_ordinal`? Why aren't the already existing constants used instead, e.g. `DM_DMSTATUS`?

3) Furthermore, why to combine CSR registers, JTAG registers and debug module registers into one enum and one set of functions? They are by definition different register groups (namespaces), so I recommend to have separate decoding functions for them.

4) Functions in the generated .c file that are not to be callable from the outside should be defined as `static`. For smaller helper functions, you can also consider `static inline` (as an optimization hint for the compiler).

5) Include guards (#ifndef + #define) have been added to debug_defines.h but they are generated in the middle of the file and cover only a part of the file. 

```
....
#define DMI_SERRX_DATA_LENGTH               0x20
#define DMI_SERRX_DATA                      0xffffffffU
#ifndef DEBUG_DEFINES_H
#define DEBUG_DEFINES_H
enum riscv_debug_reg_ordinal {
	DTM_IDCODE_ORDINAL,
	DTM_DTMCS_ORDINAL,
....
```

I'd expect them to cover the whole .h file.

6) For functions `*_reg_field_value_name`: naming them `*_reg_decode_field` would make the name shorter and IMO clearer.

7) When the functions are split to three groups (for JTAG registers, CSRs and DM registers), it is likely that the newly introduced structure `riscv_debug_reg_ctx` will not be needed anymore. Instead, you can just pass XLEN, abits or DXLEN as plain integers.

8) The argument `context` (or potentially `xlen`, `abits`, `dxlen` after addressing the above suggestions) should not be passed to the functions that do not use them. Otherwise users of these files would get warnings about unused function arguments.

9) Instead of `unsigned long long`, can we use `uint64_t`? Or have you used the former for compatibility with pre-C99 compilers? That would make sense to me.

10) Please, pay attention also to formatting of the generated C code. You can include blank lines between function definitions, auto-generated comments for functions, etc.

11) IIUC, the original aim of this change was to provide a function that allows to decode a whole register into user-readable textual form. I don't see any such function there anymore, only functions that deal with individual register fields. Do you plan to add them there?

I have not reviewed the Python changes yet.

en-sc:
@JanMatCodasip, thank you for the review!

> 1. Does it make sense to couple the register decoding functions to debug_defines.h or to put it into another C module (another pair of .h & .c files)? It is possible that some users are interested only in the constants but do not wish to bring the register decoding functions into their project. You may consider splitting it. And opinions of others are welcome :)

The reason for separating the definitions is to hide auxiliary functions (like `DTM_DMI_*`).

> 2. Why to create new enum in debug_defines.h - `riscv_debug_reg_ordinal`? Why aren't the already existing constants used instead, e.g. `DM_DMSTATUS`?

`enum riscv_debug_reg_ordinal` provides a way to specify the register. Address (e.g. `DM_DMSTATUS`) can not be used for this purpose, since some registers share a common address (e.g. `tdata1` and `mcontrol`).

> 3. Furthermore, why to combine CSR registers, JTAG registers and debug module registers into one enum and one set of functions? They are by definition different register groups (namespaces), so I recommend to have separate decoding functions for them.

Please, consider [`riscv_debug_reg_to_s()`](https://github.com/riscv/riscv-openocd/pull/892/files#diff-4187474b258a4ed647428725e3d087bc7f60b0e8a2648270beca34dcaa76498cR58). If one is to separate decoding functions as you are suggesting, this function should either be rewritten to call the decoding function for the appropriate register group or it should be copied a several times (one for each register group) with the only difference being the specific part of the context that is needed to decode registers in that group. IMHO this is an unnecessary complification of the interface.

> 4. Functions in the generated .c file that are not to be callable from the outside should be defined as `static`. For smaller helper functions, you can also consider `static inline` (as an optimization hint for the compiler).

> 5. Include guards (#ifndef + #define) have been added to debug_defines.h but they are generated in the middle of the file and cover only a part of the file.
> ...
> I'd expect them to cover the whole .h file.

> 6. For functions `*_reg_field_value_name`: naming them `*_reg_decode_field` would make the name shorter and IMO clearer.

> 7. When the functions are split to three groups (for JTAG registers, CSRs and DM registers), it is likely that the newly introduced structure `riscv_debug_reg_ctx` will not be needed anymore. Instead, you can just pass XLEN, abits or DXLEN as plain integers.

Lets consider the function `riscv_debug_reg_width()` for `dmi` register. It should return `abits + 34`. However, for other DMI registers it is not required to know the value of `abits` to get the width of the register. So, which value should the function argument have? Furthermore, it is quite hard to derive from the spec in an automated manner valid values of such a parameter, so checking if the argument is well defined if it is needed is hard and without such checks the code would be error-prone. In the current implementation the user is required to set the `is_set` flag, so it is clear where and how the value can be validated. Furthermore, if some new register is added and it requires a new parameter for decoding, all the call sites for the functions need to be modified, regardless of the register that is decoded in that call.

> 8. The argument `context` (or potentially `xlen`, `abits`, `dxlen` after addressing the above suggestions) should not be passed to the functions that do not use them. Otherwise users of these files would get warnings about unused function arguments.

> 9. Instead of `unsigned long long`, can we use `uint64_t`? Or have you used the former for compatibility with pre-C99 compilers? That would make sense to me.

I've used `unsigned long long` since it was already used in `sympy_to_c()` (more specifically `ULL` specifier). If you insist, I'll change it to `uint64_t`.

> 10. Please, pay attention also to formatting of the generated C code. You can include blank lines between function definitions, auto-generated comments for functions, etc.

> 11. IIUC, the original aim of this change was to provide a function that allows to decode a whole register into user-readable textual form. I don't see any such function there anymore, only functions that deal with individual register fields. Do you plan to add them there?

Please, see [the comment](https://github.com/riscv/riscv-openocd/pull/892#issuecomment-1660703635) and [related files](https://github.com/riscv/riscv-openocd/pull/892/files#diff-4187474b258a4ed647428725e3d087bc7f60b0e8a2648270beca34dcaa76498c).

JanMatCodasip:
@en-sc Thank you for your reply and the changes made.

To the individual points from https://github.com/riscv/riscv-debug-spec/pull/858#issuecomment-1678742086:

1) I am not sure if my comment was understood. I was thinking whether it makes sense to leave `debug_defines.h` as is, and instead generate new files `debug_reg_printers.{h,c}` that would contain the new register decoding functionality. I don't have a strong opinion here, just suggesting it for consideration (and also wondering what others think about this point).

2) The `tdata1` register (which has multiple "aliases" like `mcontrol`, `mcontrol6`, ...) seem to be a special case. Do we have any other registers like that? It would seem proper to create a separate printing function for tdata1 that takes its special nature into account.

3) Still, creation of a new enumeration that contains registers from different "namespaces" does not feel right. One solution could be to add an extra argument which tells what register type that is: `riscv_debug_reg_decode_field(RISCV_REG_CSR, RISCV_CSR_DCSR, ...);`. 

5) If it is not too difficult, it would be good to fix the include guards so that they cover the whole file, as it is usual in C.

7) The odd thing about `riscv_debug_reg_ctx` structure is that it contains a mix of data for multiple contexts. Why would I need to pass the value of `abits` in order to decode a CSR register, for example? CSR parsing should only require CSR related info from the user (e.g. XLEN). If redundant data needs to be passed to that function, it indicates that the overall architecture can probably be reworked and improved.

8) If there won't be any unused argument warnings displayed to the user of this auto-generated code, that should be good enough.

9) Unsigned long long seems OK.

10) OK

11) How about putting the `riscv_debug_reg_to_s()` function to the generated printers?

registers.py
```diff
@@ -359,10 +396,231 @@ def write_cheader( fd, registers ):
...
-            if not isinstance(value, str):
-                value = sympy_to_c(value)
+            value = sympy_to_c(value)
             fd.write( "#define %-35s %s\n" % ( name, value ) )
 
+def c_function_definition( return_type, name, args, body, fd=None):
+    if (return_type[-1] != "*"):
+        return_type += " "
+    decl = f"{return_type}{name}"
+    "({args})\n"
```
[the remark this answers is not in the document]
en-sc: Thanks for the catch! This is a mistake, the `args` are added to function declaration later via `add_arg` function. Addressed.

registers.py
```diff
@@ -359,10 +396,231 @@ def write_cheader( fd, registers ):
...
+
+    decl = add_arg(decl, args[0], "(")
+    for arg in args[1:]:
+        decl = add_arg(decl, arg)
+    decl += ")"
+
+    if fd:
+        fd.write(decl + ";\n")
+
+    args = ", ".join(args)
```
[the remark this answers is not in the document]
en-sc: Yes. Thanks for the catch! Addressed.

registers.py
```diff
@@ -359,10 +396,231 @@ def write_cheader( fd, registers ):
...
+    body_string = ""
+    for case_result in body.keys():
+        if case_result == default:
+            continue
+        for c in body[case_result]:
+            body_string += f"case {c}:\n"
+        body_string += "\t"
+        body_string += case_result.replace("\n", "\n\t") + "\n"
+
+    if (default is not None):
```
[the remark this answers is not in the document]
en-sc: Addressed.

registers.py
```diff
@@ -359,10 +396,231 @@ def write_cheader( fd, registers ):
...
+    fd_c.write(f'#include "{file_name}.h"\n#include <stddef.h>\n#include <assert.h>\n')
+    all_regs = [r for registers in registers_list for r in registers.registers if r.fields]
+    fd_h.write("#ifndef DEBUG_DEFINES_H\n#define DEBUG_DEFINES_H\n")
+    fd_h.write("enum riscv_debug_reg_ordinal {\n\t"
+          + ",\n\t".join(
+              r.ordinal_name()
+              for r in all_regs
+              )
+          + "\n};\n")
+    join_sets = lambda a, b: a | b
```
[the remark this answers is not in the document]
en-sc: Addressed.

registers.py
```diff
@@ -359,10 +396,231 @@ def write_cheader( fd, registers ):
...
+    return c_switch_stmnt(
+            "offset",
+            ((f.offset_define_name(), gen_body_for_field(f, default))
+             for f in r.fields if f.to_c_filter() and not f.symbols()),
+            default
+            )
+
+def print_cgetters_for_reg(r, all_symbols, fd):
+    prefix = r.address_define_name()
+    context_type = "void"
```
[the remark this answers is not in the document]
en-sc: Addressed.

registers.py
```diff
@@ -359,10 +396,231 @@ def write_cheader( fd, registers ):
...
+
+    if (default is not None):
+        if body_string == "":
+            return default
+        body_string += "default:\n\t"
+        body_string += default.replace("\n", "\n\t")
+    return f"switch ({expr}) ""{"f"\n{body_string}\n""}"
+
+def c_if_stmnt( expr, then_block):
+    return f"if (" + expr  + ")\n\t" + then_block.replace("\n", "\n\t")
```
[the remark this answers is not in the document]
en-sc: Addressed.

registers.py
```diff
@@ -359,10 +396,231 @@ def write_cheader( fd, registers ):
...
+            body_string += f"case {c}:\n"
+        body_string += "\t"
+        body_string += case_result.replace("\n", "\n\t") + "\n"
+
+    if (default is not None):
+        if body_string == "":
+            return default
+        body_string += "default:\n\t"
+        body_string += default.replace("\n", "\n\t")
+    return f"switch ({expr}) ""{"f"\n{body_string}\n""}"
```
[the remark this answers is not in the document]
en-sc: Addressed.

en-sc:
@JanMatCodasip, thank you for the swift response!

> 2- The `tdata1` register (which has multiple "aliases" like `mcontrol`, `mcontrol6`, ...) seem to be a special case. Do we have any other registers like that? It would seem proper to create a separate printing function for tdata1 that takes its special nature into account.

* For now there is one other register address shared between `textra32` and `textra64`.
* There is also an issue with virtual registers (for now only the `priv` register). Since they do no have an address, the corresponding define is as follows:
```
#define VIRT_PRIV                           virtual
```
Furthermore, I do not understand what is the issue with the enumeration.

> 3- Still, creation of a new enumeration that contains registers from different "namespaces" does not feel right. One solution could be to add an extra argument which tells what register type that is: `riscv_debug_reg_decode_field(RISCV_REG_CSR, RISCV_CSR_DCSR, ...);`.

This would require the user to specify the correct register group, since, as far as I understand, there is no way to determine a group of the register automatically (without creating a common enumeration either explicitly or implicitly). IMHO, the interface to these functions should be as simple as possible.

> 5- If it is not too difficult, it would be good to fix the include guards so that they cover the whole file, as it is usual in C

Addressed.
 
> 7- The odd thing about `riscv_debug_reg_ctx` structure is that it contains a mix of data for multiple contexts. Why would I need to pass the value of `abits` in order to decode a CSR register, for example? CSR parsing should only require CSR related info from the user (e.g. XLEN). If redundant data needs to be passed to that function, it indicates that the overall architecture can probably be reworked and improved.

One does not need to specify `abits.value` to decode a CSR. `abits.is_set` can be set to `false`.
I would like to leave the interface (function type) uniform across all registers to simplify the generation and usage.

> 11- How about putting the `riscv_debug_reg_to_s()` function to the generated printers?

I would like to refrain from doing so.
In my opinion, the `riscv_debug_reg_to_s()` function is not general enough (e.g. the user does not specify the format (field separators and so on)) to be moved to this repo. I would also like to minimize the amount of auto-generated code.

registers.py
```diff
@@ -281,8 +386,11 @@ def prototype(self):
     def body(self):
         return self.expression
 
-def sympy_to_c(expression):
+def sympy_to_c(expression, sym_to_c = lambda s: "(" + str(s) + ")"):
+    stc = lambda x : sympy_to_c(x, sym_to_c)
     """Implement our own string function, so we can replace 2** with 1<<."""
```
timsifive, line 391:
```suggestion
    """Implement our own string function, so we can replace 2** with 1<<."""
    stc = lambda x : sympy_to_c(x, sym_to_c)
```
Docstrings should go at the start of the function.

timsifive (changes requested):
Overall looks good.

I had assumed that debug_register_printers.* from https://github.com/riscv/riscv-openocd/pull/892 would also originate here. The code looks generally useful, and not tied to anything OpenOCD-specific. Can you place it here as well?

Can you visually check a handful of the bit diagrams in the PDF that's generated? Some of these changes might have affected them.

Can you update README.md to (very briefly) mention this new feature, and also fix the make target name?

en-sc:
> Can you visually check a handful of the bit diagrams in the PDF that's generated? Some of these changes might have affected them.

To check this l have done the following:
```
> git checkout master
> make all
> mkdir master && cp *.tex master/.
> git checkout en-sc/gen-printers
> make all
> for texfile in $(ls *.tex); do diff $texfile master/$texfile; done
```
The output of the last command:
```
3d2
< \vhEntry{d67cca8}{2023-07-17}{Evgeniy Naydanov}{Add decoding facilities into debug\_defines.h}
3c3
<     \gdef\GITHash{d67cca89b402cb8644fbe105d3f867883da88aad}    \gdef\GITAbrHash{d67cca8}    \gdef\GITAuthorDate{Mon Jul 17 21:07:22 2023 +0300}    \gdef\GITAuthorName{Evgeniy Naydanov}
\ No newline at end of file
---
>     \gdef\GITHash{c13d9bd3e494708b0275d9a0dbd5cbe13478766e}    \gdef\GITAbrHash{c13d9bd}    \gdef\GITAuthorDate{Tue Aug 22 14:26:27 2023 -0700}    \gdef\GITAuthorName{Tim Newsome}
\ No newline at end of file
```
So as far as I understand there is no difference in PDFs, since the TEX files are the same.
Is this check sufficient?

registers.py
en-sc: Addressed.

en-sc:
> I had assumed that debug_register_printers.* from [riscv/riscv-openocd#892](https://github.com/riscv/riscv-openocd/pull/892) would also originate here. The code looks generally useful, and not tied to anything OpenOCD-specific. Can you place it here as well?

Addressed.

> Can you update README.md to (very briefly) mention this new feature, and also fix the make target name?

Addressed.

en-sc:
@timsifive, @JanMatCodasip, @MarekVCodasip could you please take a look?

timsifive:
> Is this check sufficient?

Yes, that's perfect.

timsifive (approved):

PR #860, opened 2023-07-26
Fix typo in execution appendix: execute -> postexec

### LINKED ISSUE #859: Possible Typo - "If execute is set"
In "A.2 Execution Based" there is a paragraph first referencing "transfer" (AKA accessregister.transfer), then it references "execute" (AKA mcontrol.execute).

MicrosoftTeams-image

But mcontrol.execute is not really related to this behavior, and right next to "transfer" we have "postexec" which seems to match the description.

MicrosoftTeams-image (1)

--------

The relevant lines are here: https://github.com/riscv/riscv-debug-spec/blob/master/implementations.tex#L50

Shouldn't `If \FcsrMcontrolExecute is set` be changed to `If \FacAccessregisterPostexec is set`?

timsifive:
Fixes #859.


pdonahue-ventana (approved):
Yep.

PR #861, opened 2023-07-31
AR: Use manager/subordinate for master/slave.

timsifive:
Address AR feedback:
> Master/slave terminology is obviously being replaced across RV specs -
> in a somewhat haphazard manner as far as what is being used for
> replacements.  The ARC suggests using the following (and will be
> proposing this to be the standard replacement across RV specs):
> Manager/subordinate  (which remains intuitive and happens to maintain
> the M/S starting letters fwiw)


pdonahue-ventana (approved):
Fine.

PR #865, opened 2023-08-15
AR: Define "native debug" in Background section.

timsifive:


introduction.tex
```diff
@@ -300,29 +300,30 @@ \subsection{Register Definition Format}
 
 \section{Background}
 
-There are several use cases for dedicated debugging hardware, both
-internal to a CPU core and with an external connection.
+There are several use cases for dedicated debugging hardware, both in native
+debug and external debug. Native debug (sometimes called self-hosted debug)
+refers to debug software running on a RISC-V platform which debugs the same
+platform.  The Trigger Module provides features that are useful for native
+debug.  External debug refers to debug software running on somewhere else,
```
pdonahue-ventana, line 307:
```suggestion
debug.  External debug refers to debug software running somewhere else,
```

pdonahue-ventana (commented):
In addition to saying that the entire document provides features that are useful for external debug, can we say that the TM is optional and the JTAG DTM is optional?  While most implementations will have TM, very small ones might not (relying only on async haltreq and single stepping) and they may not understand that it's not required.  And somebody may want a non-JTAG DTM.

introduction.tex
timsifive: Done.

timsifive:
What do you want to say about JTAG? I think this phrasing only mentions it for external debug. I would like there to be non-JTAG transports, but currently we don't have any other options.

pdonahue-ventana:
I agree that non-JTAG transports should be an option (either custom ones or future standard ones).

My earlier proposal was to specify what hardware needs to implement if you want external debug:
> Outside the hart, hardware support for external debug involves implementing at least one Debug Module.  It also requires the JTAG Debug Transport Module and/or a custom Debug Transport Module.  Inside the hart, external debug support involves implementing the Sdext extension and, optionally, the Sdtrig extension with support for action=1.

I think that the TM (Sdtrig) is optional if you want very minimal external debug support.  And you can have either JTAG or something custom or both, but you have to have at least one transport.

timsifive:
> My earlier proposal was to specify what hardware needs to implement if you want external debug:

I don't see a need for that. Is anyone going to look at the system overview diagram and conclude that they don't need a DTM? The point of this section (and the AR request) is to describe the difference between native and external debug.

If you want a section that describes what is optional and what is not (similar to the intro to the DM chapter), then I think the System Overview chapter is the right place for that.

pdonahue-ventana (approved):
Frankly, I haven't looked at that System Overview chapter in a looong time.  That seems clear about JTAG being optional.  It also has a dotted line around the TM which indicates that the TM is optional.  Good enough.

I'm fine with what you have.

PR #866, opened 2023-08-15
Recommend debug ROM is not cached.

timsifive:
The performance of this code doesn't matter (compared to how slow JTAG is), and the fewer side effects debugging has, the better.


pdonahue-ventana (approved):
That's OK with me.  The only thing that makes me mildly nervous is that changes like this (even though it's no big deal and in a non-normative appendix) could be an excuse to further delay things.

timsifive:
Good point. Let's hold it until 1.1.

timsifive:
We've allowed in a couple other clarification changes, so I'm merging this one too.

Issue #863, opened 2023-08-15
Program buffer & requirement to insert only single instruction - how to interpret?


JanMatCodasip:
How shall we interpret the highlighted text from the spec, telling that the debugger is required to insert no more than one instruction into the program buffer? Is it:

a) a general requirement, applicable to any progbufsize?
b) a requirement applicable only to the case of progbufsize == 1?

My understanding is that b) is right but I'd like to get a confirmation.

Depending on the outcome of this discussion, I'd submit a change to clarify that text.

Thank you.

![image](https://github.com/riscv/riscv-debug-spec/assets/50193733/35174a4a-c9a1-4c0d-bc8f-9dd7df01fb5c)

pdonahue-ventana:
I think that you're right.  It basically means:
> If progbufsize is 1 then:
> - impebreak must be 1
> - It is possible that the Program Buffer can hold only one 32-or 16-bit instruction, so the debugger must only write a single instruction in this case, regardless of its size. This instruction can be a 32-bit instruction, or a compressed instruction in the lower 16 bits accompanied by a compressed nop in the upper 16 bits.

If the last two sentences were independent of progbufsize then there would be no point in allowing progbufsize>1.

JanMatCodasip:
@pdonahue-ventana, thank you for your response.

I have submitted a pull request that should make the text clearer: https://github.com/riscv/riscv-debug-spec/pull/868

JanMatCodasip:
Closing this issue -- it is resolved by https://github.com/riscv/riscv-debug-spec/pull/868.

PR #868, opened 2023-08-16
Program Buffer: Clarify requirements for case progbufsize==1

JanMatCodasip:
- Clarified that both the concerned requirements apply only to the case of progbufsize==1, as discussed in
https://github.com/riscv/riscv-debug-spec/issues/863

- Minor reword and simplification of the sentences

- Changed the order of paragraphs so that:
  - Generic requirements are stated first (Debug Mode, cmderr, ...)
  - Special requirements for progbufsize==1 are mentioned later

### DIFF
--- a/debug_module.tex
+++ b/debug_module.tex
@@ -463,26 +463,29 @@ an implicit {\tt ebreak} that is executed when a hart runs off the end of the
 Program Buffer. This is indicated by \FdmDmstatusImpebreak. With this feature, a Program
 Buffer of just 2 32-bit words can offer efficient debugging.
 
-If \FdmAbstractcsProgbufsize is 1, \FdmDmstatusImpebreak must be 1. It is possible that the Program
-Buffer can hold only one 32- or 16-bit instruction, so the debugger must only
-write a single instruction in this case, regardless of its size.
-This instruction can be a 32-bit
-instruction, or a compressed instruction in the lower 16 bits accompanied by a
-compressed {\tt nop} in the upper 16 bits.
-
-\begin{commentary}
-    The slightly inconsistent behavior with a Program Buffer of size 1 is to
-    accommodate hardware designs that prefer to stuff instructions directly
-    into the pipeline when halted, instead of having the Program Buffer exist
-    in the address space somewhere.
-\end{commentary}
-
 While these programs are executed, the hart does not leave Debug Mode (see
 Section~\ref{debugmode}).  If an exception is encountered during execution of
 the Program Buffer, no more instructions are executed, the hart remains in Debug
 Mode, and \FdmAbstractcsCmderr is set to 3 ({\tt exception error}).  If the debugger
 executes a program that doesn't terminate with an {\tt ebreak} instruction, the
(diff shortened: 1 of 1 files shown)


debug_module.tex
```diff
@@ -463,27 +463,30 @@ \section{Program Buffer} \label{programbuffer}
...
 While these programs are executed, the hart does not leave Debug Mode (see
 Section~\ref{debugmode}).  If an exception is encountered during execution of
 the Program Buffer, no more instructions are executed, the hart remains in Debug
 Mode, and \FdmAbstractcsCmderr is set to 3 ({\tt exception error}).  If the debugger
 executes a program that doesn't terminate with an {\tt ebreak} instruction, the
 hart will remain in Debug Mode and the debugger will lose control of the hart.
 
+If \FdmAbstractcsProgbufsize is 1 then the following apply:
+
+\begin{enumerate}[nosep]
```
timsifive, line 475:
```suggestion
\begin{steps}{If \FdmAbstractcsProgbufsize is 1 then the following apply:}
```
(and corresponding `\end`)

debug_module.tex
```diff
@@ -463,27 +463,30 @@ \section{Program Buffer} \label{programbuffer}
...
+    \item \FdmDmstatusImpebreak must be 1. 
+    \item The debugger must only write a single instruction into the Program Buffer,
+        regardless of the instruction size. It can be a 32-bit instruction, or 
+        a compressed instruction in the lower 16 bits accompanied by a compressed
+        {\tt nop} in the upper 16 bits.
```
timsifive, line 480: I know this is the original language, but it reads awkward to me. "a compressed instruction in the lower 16 bits accompanied by a compressed nop in the upper 16 bits" is 2 instructions.

How about:
> If the debugger writes a compressed instruction in the lower 16 bits,
> then the upper 16 must contain a compressed {\tt nop}.

timsifive (changes requested):

debug_module.tex
JanMatCodasip: Done as suggested.

debug_module.tex
JanMatCodasip: I have modified the wording slightly:

> If the debugger writes a compressed instruction into the Program Buffer,
        it must be placed into the lower 16 bits and accompanied by a compressed
        {\tt nop} in the upper 16 bits.

timsifive (approved):

pdonahue-ventana (approved):
This is also OK with me.

Note that just today there was a [message from Krste](https://lists.riscv.org/g/tech-unprivileged/message/551
) that says:
> Longer instruction encodings (48b/64b) are inevitable for
  future expansion of vector/matrix instructions regardless of this
  decision, just as all other ISAs have had to expand beyond 32b
  encodings for vector/matrix instructions.

That may cause problems in these instruction-stuffing implementations if they can only have 32b instructions, unless the longer instructions are not necessary for debugging.  Time will tell...

JanMatCodasip:
Note: I have made one last push to revert an unintended whitespace change in `debug_module.tex`. There has been no change in the actual text.

PR #869, opened 2023-08-21
AR: Don't use "non-existent" in reference to register.

timsifive:
We can always be more clear using other terminology.


pdonahue-ventana (approved):
OK.  I think that they also have concerns about the "raise an illegal instruction exception" part of a couple of those sentences, but we can address that separately.

PR #870, opened 2023-08-22
AR: Comment on unimplemented CSR behavior.

timsifive:
After a very long e-mail discussion, we landed on maintaining
    unimplemented CSR behavior as-is. Just add a quick note to be explicit
    about that intention. The end of the discussion is basically:
    1. If Sdtrig CSRs are accessed when they are not implemented or Sdtrig
       is not implemented, the hart will trap. This is to maintain backwards
       compatibility with existing implementations that use this for
       discovery.
    2. If Sdext CSRs are accessed when they are not implemented, the hart
       will trap. (No need to specify behavior when Sdext is not
       implemented, because in that case Debug Mode doesn't exist and the
       CSRs aren't accessible anyway.) This is to be consistent with Sdtrig
       behavior.

### DIFF
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -358,4 +358,12 @@ time than is expected.
 Attempts to access an unimplemented Trigger Register raise an illegal instruction
 exception.
 
+\begin{commentary}
+    Privileged spec v1.12 requires accesses to unimplemented CSRs to raise
+    illegal instruction exceptions, but that may change in the future. Even if
+    it does change, the behavior here overrides it for Sdtrig CSRs for backwards
+    compatibility. This is even true if Sdtrig is not implemented at all, so the
+    discovery algorithm in Section~\ref{sec:trigger} still works.
+\end{commentary}
+
 \input{hwbp_registers.tex}


pdonahue-ventana:
It is my understanding that there is no requirement for any particular behavior in the (as yet undocumented) behavior on accesses to unimplemented CSRs.  They no longer need to trap but they also don't need to return 0.

Consider implementation X:
- no triggers are implemented
- nothing ever causes an illegal instruction exception.  Look at all the gates we save.  And medeleg[2] can be hardwired, saving one whole flop.
- tinfo isn't implemented.  Why would we if we have no triggers and we're skimping elsewhere?
- reads of tinfo return the value from pmpcfg4 (or whatever).  That's a nearby encoding and doing this saves a gate or two on this super-cheapskate implementation.

We get to step 4 of the enumeration recipe.  Reading tinfo didn't cause an exception and it returned non-zero.  The recipe says to parse the read value to discover the trigger type but we're actually parsing something that's not tinfo at all.  I see no possible way that this can work.  While the recipe will work on sane implementations, I think the general requirement is that the debugger needs to already know what is implemented.  Unified discovery or maybe reading JTAG IDCODE and looking up in some database that the debugger as access to.

I think that it might be best to leave the recipe as-is and preface it all with "If an implementation raises an illegal instruction exception on accesses to unimplemented CSRs then the enumeration recipe is below.  If it does not raise illegal instruction exceptions then the enumeration method is beyond the scope of this specification."  That might get the ARC's attention about their contemplated backward-incompatible change to the priv spec.

timsifive:
This version is awkward, but I dislike it less than simply removing the requirement to raise an exception, because deep down I believe that the privspec won't be changed in a way that makes discovery impossible.

Sdext.tex
```diff
@@ -203,8 +203,13 @@ \section{Core Debug Registers} \label{debreg}
...
 optionally also using abstract debug commands.
 
-Attempts to access an unimplemented Core Debug Register raise an illegal
-instruction exception.
+The behavior when accessing unimplemented Core Debug Registers follows what is
+defined in the Privileged Spec.
+
+\begin{commentary}
+    At the time of this writing the Privileged Spec says one thing, but the
+    maintainers are planning to change it in the next revision.
```
pdonahue-ventana, line 211:
```suggestion
    Privileged spec v1.12 requires accesses to unimplemented CSRs to raise illegal instruction
    exceptions, though that requirement may be relaxed in later versions.
```

Sdtrig.tex
```diff
@@ -26,16 +27,18 @@ \chapter{Sdtrig (ISA Extension)}
 
 \section{Enumeration}
 
-\begin{steps}{Each trigger may support a variety of features. A debugger can
+\begin{steps}{Each trigger may support a variety of features.
+    If Sdtrig is implemented or unimplemented CSRs raise an exception,
```
pdonahue-ventana, line 31: Is it possible for Sdtrig to be implemented but with zero triggers?  That seems kind of silly but I think that technically somebody could claim that their zero trigger core is Sdtrig complliant.  That serves no real purpose but looks slightly more impressive in marketing Powerpoint (which is slimy but technically legit).  To avoid this, I think that we should have a sentence somewhere that says that an Sdtrig-compliant implementation has at least one trigger (unless I'm forgetting some Sdtrig feature other than triggers that somebody might want to implement without doing triggers).  That prevents any Powerpoint ethical gray areas and also makes this sentence correct.

Sdtrig.tex
```diff
@@ -357,7 +360,13 @@ \section{Trigger Registers}
...
 time than is expected.
 
-Attempts to access an unimplemented Trigger Register raise an illegal instruction
-exception.
+The behavior when accessing unimplemented Trigger Registers follows what is
+defined in the Privileged Spec.
+
+\begin{commentary}
+    At the time of this writing the Privileged Spec says one thing, but the
+    maintainers are planning to change it in the next revision.
```
pdonahue-ventana, line 368:
```suggestion
    Privileged spec v1.12 requires accesses to unimplemented CSRs to raise illegal instruction
    exceptions, though that requirement may be relaxed in later versions.
```

xml/hwbp_registers.xml
```diff
@@ -149,8 +149,8 @@
     <register name="Trigger Info" short="tinfo" address="0x7a4">
         This register is optional if no triggers are implemented, or if
         \FcsrTdataOneType is not writable and \FcsrTinfoVersion would be 0. In
-        this case the debugger can read the only supported type from
-        \RcsrTdataOne.
+        this case reading this register must return 0 or raise an exception, and
```
pdonahue-ventana, line 152:
```suggestion
        this case reading this register must return 0 or raise an illegal instruction exception, and
```

pdonahue-ventana: Also, I think that it's ambiguous whether it must (return 0 OR (exception AND debugger can ...)) vs. ((return 0 OR exception) AND debugger can ...).  The comma might be intended to help but it's not obvious to me until I think about it some more.  Maybe two sentences: It must return 0 or exception.  If either of those happen then the debugger can ...

Sdext.tex
timsifive: Done.

Sdtrig.tex
timsifive: That was explicitly allowed back when there was a single debug extension, to keep triggers optional. You're right that it doesn't make sense to allow Sdtrig to be implemented with 0 triggers. I support the idea, but I don't think that fits in this PR, and it's also technically backwards incompatible.

xml/hwbp_registers.xml
timsifive: Agreed about 2 sentences.

How strongly do you feel about "illegal instruction exception?" I was deliberately vague because the hypervisor extension mandates that CSRs that appear not to exist to a guest must raise a virtual instruction exception. But maybe we should just ignore that possibility, because I bet there are other places in the spec where we don't think about that scenario.

pdonahue-ventana: Sure, we can leave the type of exception off if you want.  That requires the reader to figure it out but it's not rocket science.  It's not like any sane reader would expect a page fault.

Sdtrig.tex
pdonahue-ventana: I don't think that it's incompatible.  Old implementations would have said that they do the debug spec but there was no "Sdtrig" until recently.  I guess that somebody in the last year or so could have claimed Sdtrig compliance without any triggers but that seems unlikely and I don't think that adding this new requirement would require them to change anything other than some Powerpoint slides.

This is along the lines of what I was talking about last week when I suggested explicitly saying that you can do external debug without Sdtrig, but you pointed out the diagram with the dotted line around the TM.

If we don't say that Sdtrig compliance => non-zero number of triggers then I worry that this recipe doesn't work for implementations that don't get exceptions on accesses to unimplemented CSRs.  (In real life, I hope that everyone gets exceptions AND has at least one trigger if they claim compliance, but in theory there's a problem.)

timsifive: I made #872 for this.

timsifive:
I'm putting this on hold, because now maybe ARC is going to allow the original phrasing where unimplemented CSRs raise an exception.

pdonahue-ventana (approved):
I am somewhat concerned that a subset of AR people think this is fine (per email discussions) but that others will disagree.  But let's move forward with this.  We're 100% compatible with the current priv spec and 1.13 is somewhat speculative.

PR #872, opened 2023-08-30
Sdtrig must implement at least one trigger.

timsifive:
Based on discussion in #870, where Paul mentioned:
> I don't think that it's incompatible. Old implementations would have
> said that they do the debug spec but there was no "Sdtrig" until
> recently. I guess that somebody in the last year or so could have
> claimed Sdtrig compliance without any triggers but that seems unlikely
> and I don't think that adding this new requirement would require them to
> change anything other than some Powerpoint slides.


pdonahue-ventana (approved):
Great.

PR #874, opened 2023-09-14
AR: Remove comment about privspec/CSR behavior

timsifive:
From Greg: "... why is this non-normative comment even necessary?  It's directly consistent with 1.12, and it's compatible with 1.13 which allows for trapping and just doesn't require it.  In other words, wrt 1.13, this trapping mandate by Debug 1.0 doesn't conflict with Priv 1.13.  And that allows Debug 1.0 to come along and tighten what is allowed by 1.13."


pdonahue-ventana (approved):
Back to the original.

PR #875, opened 2023-09-14
AR: Clarify what debuggers can assume about MRs

timsifive:


pdonahue-ventana (approved):
Yep.  I agree that debuggers must not assume that.

PR #878, opened 2023-09-18
AR: Remove Message Registers.

timsifive:
These were first introduced in #728.

Hopefully I got everything.


pdonahue-ventana (commented):
The debug_register_printers.h deletion seems to be unrelated.  Is that intentional?

timsifive:
> The debug_register_printers.h deletion seems to be unrelated. Is that intentional?

Good catch. Thanks.

pdonahue-ventana (approved):
OK

PR #879, opened 2023-09-18
AR: Clarify unimplemented Sdtrig.

timsifive:
If Sdext is implemented and Sdtrig is not implemented, then accessing any of the Sdtrig CSRs must raise an illegal instruction exception.

(This is already said in privspec 1.12, but that might change in the future.)


pdonahue-ventana (approved):
I'm glad that we could converge on a single sentence solution.

PR #881, opened 2023-09-20
AR: List which extensions were considered.

timsifive:
I added a hint as to how they were considered as LaTeX comments, because I want to keep that info but don't want to word it nicely/completely enough to actually have it as part of the spec text.

### DIFF
--- a/introduction.tex
+++ b/introduction.tex
@@ -98,8 +98,25 @@ functionality.
 
 \section{Context}
 
-This document is written to work with all RISC-V ISA extensions ratified
-through 2021.
+\begin{steps}{This specification attempts to support all RISC-V ISA extensions
+that have, roughly, been ratified through the first half of 2023.  In
+particular, though, this specification specifically addresses features in the
+following extensions:}
+\item A % Mention atomics in debug mode section
+\item C % Mentioned in single-word progbuf context
+\item D % Access register abstract register numbers.
+\item F % Access register abstract register numbers.
+\item H % Various trigger bits
+\item Sm1p13 % E.g. exceptions
+\item Ss1p13 % Mention S-Mode in various trigger stuff
+\item Smstateen % Mentioned in hcontext register description.
+\item V % Mentioned in debug mode entry section
+\item Zawrs % we refer to wrs.sto and wrs.nto
+\item Zcmp % Mention cm.push, cm.pop in Sdtrig
+\item Zicbom % Mention cbo.clean, cbo.flush, and cbo.inval in Cache Operations
+\item Zicboz % Mention cbo.zero in Cache Operations
+\item Zicbop % Mention prefetch instructions in Cache Operations
+\end{steps}
 
 \subsection{Versions}
 


timsifive:
Did I miss any extensions?

introduction.tex
```diff
@@ -98,8 +98,22 @@ \section{Terminology}
...
+that have, roughly, been ratified through the first half of 2023.  In
+particular, though, this specification specifically addresses features the
+following extensions:}
+\item A % Mention atomics in debug mode section
+\item C % Mentioned in single-word progbuf context
+\item D % Access register abstract register numbers.
+\item F % Access register abstract register numbers.
+\item H % Various trigger bits
+\item S % Mention S-Mode in various trigger stuff
+\item V % Mentioned in debug mode entry section
```
timsifive, line 111:
```suggestion
\item Smstateen % Mentioned in hcontext register description.
\item V % Mentioned in debug mode entry section
```

introduction.tex
```diff
@@ -98,8 +98,22 @@ \section{Terminology}
 
 \section{Context}
 
-This document is written to work with all RISC-V ISA extensions ratified
-through 2021.
+\begin{steps}{This specification attempts to support all RISC-V ISA extensions
+that have, roughly, been ratified through the first half of 2023.  In
+particular, though, this specification specifically addresses features the
```
gfavor, line 103: "features the"  =>  "features in the"

timsifive: Done. Thanks.

introduction.tex
```diff
@@ -98,8 +98,24 @@ \section{Terminology}
...
+that have, roughly, been ratified through the first half of 2023.  In
+particular, though, this specification specifically addresses features in the
+following extensions:}
+\item A % Mention atomics in debug mode section
+\item C % Mentioned in single-word progbuf context
+\item D % Access register abstract register numbers.
+\item F % Access register abstract register numbers.
+\item H % Various trigger bits
+\item Sm1p13 % E.g. exceptions
+\item Ss1p13 % Mention S-Mode in various trigger stuff
```
pdonahue-ventana, line 111: I would say 1.12.  Although we have early hints of the content, I haven't even seen a draft of 1.13 yet.

pdonahue-ventana:
Also Zawrs because we refer to wrs.sto and wrs.nto.  That's the only other one that seemed appropriate when I reviewed https://wiki.riscv.org/pages/viewpage.action?pageId=16458762 (recently ratified extensions).

introduction.tex
gfavor: There's nothing expected in 1.13 that would have any Debug impact.

timsifive: So saying 1.12 (which is my preference also) is OK here? I only wrote 1.13 because that was in Greg's e-mail.

timsifive:
I've added Zawrs

introduction.tex
gfavor: On the one hand one _could_ only say support up through 1.12, but that will then leave people wondering whether Debug 1.0 supports up through 1.13 (and 1.13 will be frozen well before Debug gets ratified).

pdonahue-ventana: I guess that I'm OK with either.  If I see something shocking in 1.13 then I may raise a red flag later.  But it sounds like that won't be necessary.  So let's just stick with 1.13.

timsifive: Great. We'll stick with 1.13.

pdonahue-ventana (approved):
OK.  I always have hesitation that there's something we overlooked, but I can't think of anything else.

PR #882, opened 2023-09-20
AR: Clarify mcontext/hcontext.

timsifive:
It was a bit confusing before. Hopefully now it's more clear how they relate, and why you might want each one.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -218,28 +218,15 @@
     </register>
 
     <register name="Hypervisor Context" short="hcontext" address="0x6a8">
-        This optional register is only accessible in S/HS-mode, M-mode and
-        Debug Mode.
+        This optional register may be implemented only if the H extension is
+        implemented. If it is implemented, \RcsrMcontext must also be implemented.
 
-        Accessibility of this CSR is controlled by \Rmstateenzero[57]
-        in the Smstateen extension.
+        This register is only accessible in HS-Mode, M-mode and Debug Mode. If
+        Smstateen is implemented, then accessibility of in HS-Mode is
+        controlled by \Rmstateenzero[57].
 
-        If the H extension is not implemented then this register is not
-        implemented, though the underlying state may be accessible via
-        the optional \RcsrMcontext alias.
-
-        <field name="hcontext" bits="XLEN-1:0" access="WARL" reset="0">
-            Hypervisor mode software can write a context number to this register,
-            which can be used to set triggers that only fire in that specific
-            context.
-
-            An implementation may tie any number of upper bits in this field to
-            0. If the H extension is not implemented, it's recommended to implement
-            no more than 6 bits on RV32 and 13 on RV64 (as visible through the
-            \RcsrMcontext register).  If the H extension is implemented,
-            it's recommended to implement no more than 7 bits on RV32
-            and 14 on RV64.
-        </field>
+        This register is an alias of the \RcsrMcontext register, providing
+        access to the \FcsrMcontextHcontext field from HS-Mode.
     </register>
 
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -218,15 +218,13 @@
     </register>
 
     <register name="Hypervisor Context" short="hcontext" address="0x6a8">
-        This optional register is only accessible in S/HS-mode, M-mode and
-        Debug Mode.
+        This optional register may only be implemented if the H extension is
+        implemented. If it is implemented, \RcsrMcontext must also be implemented.
```
pdonahue-ventana, line 222: "may only be implemented" seems ambiguous because "may only" is kind of like "may not" which is like "you may not have a piece of candy."

Somehow this slight tweak seems less ambiguous to me:
```suggestion
        This optional register may be implemented only if the H extension is
        implemented. If it is implemented, \RcsrMcontext must also be implemented.
```

xml/hwbp_registers.xml
```diff
@@ -218,15 +218,13 @@
...
 
-        Accessibility of this CSR is controlled by \Rmstateenzero[57]
-        in the Smstateen extension.
+        This register is only accessible in HS-Mode, M-mode and Debug Mode.
 
-        If the H extension is not implemented then this register is not
-        implemented, though the underlying state may be accessible via
-        the optional \RcsrMcontext alias.
+        if Smstateen is implemented, then accessibility of this CSR in HS-Mode
```
pdonahue-ventana, line 226:
```suggestion
        If Smstateen is implemented, then accessibility of this CSR in HS-Mode
```

xml/hwbp_registers.xml
```diff
@@ -262,8 +260,19 @@
...
 
     <register name="Machine Context" short="mcontext" address="0x7a8">
-        This optional register is an alias for \RcsrHcontext and is only
-        accessible in M-mode and Debug mode.
+        This register must be implemented if \RcsrHcontext is implemented, and
+        is optional otherwise.  It is only accessible in M-mode and Debug mode.
+
+        This register is an alias of the \RcsrHcontext register, and provides
+        access to the \FcsrHcontextHcontext field even if the \RcsrHcontext
+        register is not implemented.
```
pdonahue-ventana, line 268: This seems confusing to me.  It's an alias of a register that's not implemented.

I was going to suggest that we say: If hcontext is implemented then this is an alias of that.  If hcontext is not implemented then this is its own register.

Then it occurred to me that there's no definition of mcontext fields in the latter scenario.  Maybe we should reverse things and define the fields here.  Then mcontext doesn't need to say anything about hcontext except that it must exist if hcontext exists (which it says in the previous paragraph).  And hcontext simply says that it's an optional alias of mcontext and which modes it's accessible in.

The only complication is that the field name within this register is named hcontext.  But maybe that's OK if we explain that one primary purpose of mcontext is for hypervisors running in M mode.

xml/hwbp_registers.xml
timsifive: Done.

xml/hwbp_registers.xml
timsifive: Done.

xml/hwbp_registers.xml
timsifive: I'll flip it like you suggest, and see if we like that better.

pdonahue-ventana (approved):
I see another reason to have the field defined under mcontext.  Previously, it was defined under hcontext which only exists if H is implemented but then the field itself said "If the H extension is implemented..." which was just confusing if you looked at this register in isolation.  So this is much better.

PR #883, opened 2023-09-22
AR: Define mcontrol* triggers and multiple accesses

timsifive:
mcontrol: Behavior is undefined because it's deprecated.

mcontrol6: If a trigger reports hit=before then some memory accesses may have been performed. If it reports immediately after, then all accesses have already been performed.


xml/hwbp_registers.xml
```diff
@@ -378,6 +378,9 @@
             though the load will not update its destination register. Debuggers
             should consider this when setting such breakpoints on, for example,
             memory-mapped I/O addresses.
+
+            It is undefined what should happen when the instruction in question
```
gfavor, line 382: Saying that the behavior is "undefined" means that the possibilities are architecturally unbounded.  Any behavior would be allowed - which is both undesirable and useless to some debugging code with such instructions.

Conversely, I would expect people that care about the Zc* instructions would (rightly) object to freeze (let alone ratification) of Debug 1.0 since it would be unusuable on embedded software that makes use of these instructions.

xml/hwbp_registers.xml
```diff
@@ -746,6 +749,12 @@
             \Rxepc or \RcsrDpc (depending on \FcsrMcontrolSixAction) must be set
             to the virtual address of the instruction that matched.
 
+            If an instruction partially executed, the reported timing is still
+            before. This might mean that when the instruction is executed again
+            it will perform some of the same stores again (e.g. \tt{cbo.zero}),
+            or the instruction might have a mechanism that avoids duplicating the
+            stores (e.g. vector stores).
```
gfavor, line 756: The "vector store" reference leaves the crucial piece of information a mystery to the reader unless they are intimately familiar with RVV.  The parenthetical should say:

(e.g. vector stores and the vstart mechanism).

xml/hwbp_registers.xml
timsifive: What is the right term here? Unspecified?

`mcontrol` triggers are already deprecated, but debuggers should know about them so they're in the spec. New hardware implementations should implement mcontrol6 instead, where the behavior is specified. Is that OK, or should I invent something for people insisting on mcontrol triggers with Zc*?

xml/hwbp_registers.xml
timsifive: Done. Thanks.

xml/hwbp_registers.xml
```diff
@@ -378,6 +378,9 @@
             though the load will not update its destination register. Debuggers
             should consider this when setting such breakpoints on, for example,
             memory-mapped I/O addresses.
+
+            It is \unspecified what should happen when the instruction in question
+            performs multiple memory accesses.
```
pdonahue-ventana, line 383: What should happen is only unspecified with respect to the memory accesses, not with respect to anything else.
```suggestion
            If an instruction matches this trigger and the instruction performs
            multiple memory accesses, it is \unspecified which memory accesses
            have completed before the trigger fires.
```

xml/hwbp_registers.xml
```diff
@@ -766,6 +775,9 @@
             \Rxepc or \RcsrDpc (depending on \FcsrMcontrolSixAction) must be set
             to the virtual address of the next instruction that must be executed
             to preserve the program flow.
+
+            If the instruction performed multiple memory accesses, all of them
+            have been completed.
```
pdonahue-ventana, line 780: For completeness, should we also say the same thing on "after"?

xml/hwbp_registers.xml
```diff
@@ -746,6 +749,12 @@
             \Rxepc or \RcsrDpc (depending on \FcsrMcontrolSixAction) must be set
             to the virtual address of the instruction that matched.
 
+            If an instruction partially executed, the reported timing is still
+            before. This might mean that when the instruction is executed again
+            it will perform some of the same stores again (e.g. \tt{cbo.zero}),
+            or the instruction might have a mechanism that avoids duplicating the
+            stores (e.g. vector stores and the vstart mechanism).
```
gfavor, line 756: This focuses on stores, but what about register results of an instruction that performs multiple memory accesses?

Must any and all register results be performed only if the instruction completes, i.e. before timing disallows any register modifications (that might, for example, prevent restartability of the instruction aupon return from the debug handler), and after timing of course requires all such modifications to have been performed.

And what about loads?  Are some allowed, like stores, to have been performed under before timing?  And what if some of those loads are to non-idempotent memory?

Lastly, cbo.zero is not the best of example since it architecturally performs a single block-size memory write.  (An implementation can choose to implement that as a series of smaller stores of some size, but the expected common implementation would be one memory write.)

It seems like this paragraph needs to cover read acesses, write accesses, and register writes.  Whether the statement is common to all three, or is a separate statement for register writes and for memory accesses, or maybe different for all three, all three need to be covered explicitly in some way so that peopel know what are  the bounds on allowed and disallowed behavior for before triggers and for after triggers.

xml/hwbp_registers.xml
timsifive: I decided not to because I don't see how you can have "at least one additional instruction retired" without having all these memory accesses be complete. Is that allowed by RISC-V somewhere?

If you feel strongly about this I'll add it.

xml/hwbp_registers.xml
timsifive: The first sentence is the important one. "If an instruction partially executed, the reported timing is still before." The rest is just example. I'll tweak the language to allow for more implementations. Do you have a suggestion for an instruction that performs multiple operations but that doesn't have a restart mechanism? I can't think of any.

gfavor: There's no obvious perfect statement one coudl make.  But one needs to inform users what is and isn't possible.  Obviously users prefer a more useful statement that will allow most instructions to be restartable.

For example (and I'm not trying to say that this is what should be specified - the TG needs to consider the pros and cons of whatever it would choose to specify), one could say that some of the memory accesses (loads and stores) may have been performed but no registers will have been modified.  And one should also note that while most instructions will still always be restartable, some may not be (and accesses to non-idempotent memory may be repeated).

timsifive: I've rewritten almost all of the mcontrol6 `before` timing. Can you see if this is more clear?

xml/hwbp_registers.xml
```diff
@@ -741,15 +745,23 @@
             </value>
 
             <value v="1" name="before">
-            The trigger fired just before the instruction that triggered it was
-            retired, but after all preceding instructions are retired.
+            The trigger fired before the instruction that matched it was
+            retired, but after all preceding instructions are retired. This
+            explicitly allows for instructions partially executing.
```
gfavor, line 750: "for instructions to be partially executed."

xml/hwbp_registers.xml
```diff
@@ -741,15 +745,23 @@
...
             \Rxepc or \RcsrDpc (depending on \FcsrMcontrolSixAction) must be set
             to the virtual address of the instruction that matched.
 
-            If a load operation matched and \FcsrMcontrolSixSelect=1 then a
-            memory access has been performed (including any side effects of
-            performing such an access) even though the load has not updated its
-            destination register.
+            An instruction that caused a trigger to fire might be executed
+            partially. In that case not all memory accesses may have been
+            performed, and some registers may not have been updated. Executing
```
gfavor, line 756: Is this trying to say that ALL registers will not have been updated, or that some registers may have been updated?

xml/hwbp_registers.xml
timsifive: Done.

xml/hwbp_registers.xml
timsifive: I'm trying to say that anything partial is OK, from 0% to 100%. E.g. if we had an instruction that pops 3 registers from the stack, any of those 3 registers (or none of them) may contain the new value. The important thing for the debugger is that nothing breaks if we now resume off that instruction by letting it execute again. (Can't account for non-idempotent memory, however.)

xml/hwbp_registers.xml
pdonahue-ventana: Good point.  Never mind.

pdonahue-ventana (approved):
Technically, nobody can guarantee that executing an instruction the second time is equivalent to executing it the first time because some other hart (or DMA or whatever) could have modified memory in between.  But I think that the spirit is clear and it would just add unnecessary confusion to add any additional verbiage about other harts.  So this is fine.

xml/hwbp_registers.xml
timsifive: @gfavor Is this OK, or would you like to see the language tweaked more?

gfavor:
I assume Tim has now noticed that this PR probably wants to be reverted or replaced with a PR reflecting Oct 3rd AR feedback (email titled "Debug 1.0 AR next steps") about how best to more generally specify this issue.

Issue #857, opened 2023-07-17
(External) Debugger Implementation Read/Write Memory - transfer not specified in examples


JekkuChen:
In the Debugger Implementation, there are step by step instructions regarding how to read/write registers/memory using the Program Buffer. In all the examples, the transfer bit should be set when using the Program Buffer.

In the "Using Program Buffer" section of "Accessing Registers" it shows for example: 
![image](https://github.com/riscv/riscv-debug-spec/assets/37776102/25b1493e-85c0-46fd-801a-1064e112d161)

However, in the Memory Read/Writes, the command register - using Program Buffer - writes seem to be missing specifying transfer to be set which is required

![image](https://github.com/riscv/riscv-debug-spec/assets/37776102/1fd809e2-e19a-4912-9e1f-9a944a69cfa8)
<sub>Use Program Buffer for Read</sub>  
![image](https://github.com/riscv/riscv-debug-spec/assets/37776102/15f3db97-fa90-4ed9-a353-5c3bffe5ae88)
<sub>Use Program Buffer for Write</sub>

All memory Prog Buffer examples above don't specify transfer - explicitly -  as set, unlike the register read/write Program Buffer examples which explicitly display that transfer is set in those cases.

timsifive:
Thanks for pointing this out.

pdonahue-ventana:
This was addressed in #886.

PR #886, opened 2023-09-26
Fix debugger examples: use transfer with write

timsifive:
Addresses #857.


pdonahue-ventana (approved):
And I thought that I was so careful when I redid these values...

Issue #894, opened 2023-10-01
Does "this section" mean "this chapter"?


jhauser-us:
Section 4.1, "Debug Mode", says:
> All implemented instructions operate just as they do in M-mode, unless an exception is mentioned in this section.

Does "this section" mean just Section 4.1, as would seem to be the literal interpretation, or does it mean all of Chapter 4 (Sdext)?

timsifive:
@jhauser-us, if this change addresses your feedback, can you close this issue?

PR #895, opened 2023-10-02
AR: Clarify section/chapter in Sdext.

timsifive:
Addresses #894


pdonahue-ventana (approved):
I don't think that the section->list was strictly necessary but it's fine. But the first change was obviously necessary.

Issue #893, opened 2023-10-01
Why does an instruction count trigger also count traps?


jhauser-us:
Section 5.6.13, "Instruction Count", says that an instruction count trigger decrements its count on any trap, even if no instruction retires.  This is odd to me, given that some care is taken otherwise to decrement the count only for instructions that retire after being fetched in a matching privilege mode, and no other instructions.  At a minimum, I think a comment is warranted to explain the choice to decrement the count on traps, especially for those traps that the Unprivileged Spec categorizes as *invisible* (from Section 1.6, "Exceptions, Traps, and Interrupts").  Decrementing the count on invisible traps seems to me to flout the intention behind the word *invisible*.

timsifive:
The core reason to count all traps is that a debugger wants to retain control when an instruction is executed which causes a trap that should be resolved but isn't. E.g. a page fault, and the fault handler restarts the instruction without configuring the proper page table entry. Ignoring invisible traps sounds appealing, but I worry that it makes debugging harder exactly when it's most needed.

I didn't realize invisible traps were a thing. I think what matters in this case is whether the traps are visible to the debugger which set the trigger. Since triggers are only settable from Debug Mode or M-Mode, all traps presumably are visible. Is that right?

pdonahue-ventana:
#896 incorporates the answer into the spec.

PR #896, opened 2023-10-02
AR: Clarify why icount matches on every trap.

timsifive:
Clarifies (partly?) #893.

### DIFF
--- a/Sdext.tex
+++ b/Sdext.tex
@@ -141,7 +141,7 @@ If the instruction being stepped over would normally stall the hart, then
 instead the instruction is treated as a {\tt nop}. This includes {\tt wfi},
 {\tt wrs.sto}, and {\tt wrs.nto}.
 
-\subsection{Icount Trigger}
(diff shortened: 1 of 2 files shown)


pdonahue-ventana (commented):
I think that there's a grammar problem.  Maybe "an" illegal instruction trap?

Rather than giving specific examples without a general rule, I'd say something like:
> For use in single step, icount must match for traps where the instruction will not be reexecuted after the handler, such as illegal instructions that are emulated by privileged software and the instruction being emulated never retires.  Ideally, icount would not match for traps where the instruction will later be retried by the handler, such as page faults where privileged software modifies the page tables and returns to the faulting instruction which ultimately retires.  Trying to distinguish the two cases leads to complex rules, so instead the rule is simply that all traps match.  See also Section~\ref{stepIcount}.

(At least I think that "will the instruction ever retire" is the motivating principle.)

timsifive:
That's better. Thanks.

pdonahue-ventana (approved):
Great.

Issue #892, opened 2023-10-01
Exception priority table should be consistent with latest Priv Spec


jhauser-us:
In Section 5.3, "Priority", the Debug Spec's table of exception priorities should be made consistent with the corresponding tables in the latest Privileged Spec draft, found in that document in Section 3.1.15, "Machine Cause Register (mcause)", and (for the hypervisor extension) Section 9.6.1, "Trap Cause Codes".

PR #897, opened 2023-10-03
AR: Update priority table from latest privspec

timsifive:
Addresses #892.

Here's the privspec table:
image

And here is what ours looks like:
image

### DIFF
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -114,36 +114,36 @@ in the chain.
 
 \begin{table}[H]
 \centering
-\begin{tabular}{|l|r|l|l|}
+\begin{tabulary}{\textwidth}{|l|p{.7in}|p{2.3in}|p{2.5in}|}
   \hline
-  Priority      & Exception & Description & Trigger \\
-                &      Code &             & \\
+  Priority      & Exception Code & Description & Trigger \\
   \hline
   {\em Highest} &          3 & & etrigger \\
                 &          3 & & icount \\
                 &          3 & & itrigger \\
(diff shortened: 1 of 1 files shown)


pdonahue-ventana:
What you pasted above doesn't match the latest table 7 from https://github.com/riscv/riscv-isa-manual/blob/main/src/hypervisor.adoc#trap-cause-codes.  It's missing guest-page faults, virtual instructions, and maybe more.

timsifive:
I copied the table from https://github.com/riscv/riscv-isa-manual/blob/d924944fa58d21547ee72fd3fec6020ef2f3337c/src/machine.adoc#exception-priority

How does this work? Clearly the machine table is the base one. Then the hypervisor one is an extension of that. Is there a supervisor one that also extends it? How do we know what is the right one for Debug to extend?

I'm inclined to say that the machine mode one provides a good reference. Adding hypervisor or other extension causes is left as an exercise to the reader.

pdonahue-ventana:
The hypervisor table is a superset.  If you don't implement the H extension then you can still use the hypervisor table and just ignore the exceptions that are N/A like virtual instruction or guest-page fault.  And if you don't implement the A extension then you ignore things about AMOs.  And if you just have a M-only or M+U implementation then you ignore page faults.  etc.

timsifive:
Alright. I've updated based on the other exception priority table in the privspec. I guess I'm just grumpy that there is more than one.

Privspec table:
image

New debug spec table:
image

pdonahue-ventana (approved):
👍

PR #899, opened 2023-10-04
AR: Clarify triggers and multiple state changes

timsifive:
Incorporate more feedback from AR. Moved this into a separate Sdtrig section, since the discussion could also apply to instructions that cause triggers beside memory access ones. (E.g. an instruction that combines memory accesses with a trigger, and itrigger.)

Follow up to #883.

### DIFF
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -319,6 +319,24 @@ Cache operations encoded as HINTs do not match debug triggers.
     \end{steps}
 \end{commentary}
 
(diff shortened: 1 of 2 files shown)


pdonahue-ventana (approved):
OK.  And this is broader than just writing to memory.  It would also apply to an instruction that loads multiple registers.  Like ARM's AArch32 can load any subset of GPRs with a single instruction, but you have to make sure that if one of the destinations is also the source register (the address) that you don't update that register if you get a page fault halfway through.  That allows you to still have the original address when you reexecute.

PR #900, opened 2023-10-05
Add people with git PRs to credits.

timsifive:
Add everyone who had a PR merged that changed the spec itself.


pdonahue-ventana (approved):
Good idea.

PR #901, opened 2023-10-09
AR: Clarify itrigger behavior.

### LINKED ISSUE #890: For interrupt and exception triggers with action = 0, what value for mcause?
Section 5.6.14, "Interrupt Trigger", says:
> When the trigger fires, all CSRs are updated for the interrupt trap as defined by the Privileged Spec, and the requested action is taken just before the first instruction of the trap handler is executed.  If the trigger fires with action =0 then zero is written to the `tval` CSR on the breakpoint trap.

If a trigger fires on an interrupt trap that is taken into M mode, and `itrigger`.action = 0, what value is required to end up in the `mcause` CSR?  To indicate a breakpoint exception, it needs to be the value 3, but the quoted text suggests it will rather be the value that the interrupt handler would normally see.

Much the same question arises in the next subsection, "Exception Trigger", which has almost identical text.  If a trigger fires on an exception trap that is taken into M mode, and `etrigger`.action = 0, what value is required to be in the `mcause` CSR?  Is it 3 for the breakpoint exception, or is it the code for the exception that caused the trigger to fire?

Note, if the answer for an exception trigger is 3, then the cause code for the original triggering exception is lost, unless the trigger is configured with only one bit set in `tdata2` (or the subset of possible exception causes can be proved to be a singleton by some other means).

timsifive:
Specifically, make clear that itrigger fires *after* the original exception has finished updating all the CSRs.

Hopefully resolves #890.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -1145,11 +1145,13 @@
         interrupt. (E.g.\  it does not fire when a timer interrupt occurs but that
         interrupt is not enabled in \Rmie.)
 
-        When the trigger fires, all CSRs are updated for the interrupt trap as defined by the
-        Privileged Spec, and the requested action is taken just before the
-        first instruction of the trap handler is executed.
-        If the trigger fires with \FcsrItriggerAction=0 then zero is written to the
-        {\tt tval} CSR on the breakpoint trap (see ~\ref{sec:nativetrigger}).
+        When the trigger matches, it fires after the trap occurs, just before
+        the first instruction of the trap handler is executed.  If
+        \FcsrItriggerAction=0, the standard CSRs are updated for taking the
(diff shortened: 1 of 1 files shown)


pdonahue-ventana (commented):
I would say that the trigger fires after the trap.  This says "when the trigger fires" we first do the trap behavior.  It seems clearer to me to say that the trap happens and we do all the normal trap stuff then after that the trigger fires and we do all the normal trigger stuff.

timsifive:
Yes, I like this better. Thanks.

pdonahue-ventana (commented):
The same update should probably be made for etrigger since it had pretty much the same language.

timsifive:
Made the same change for etrigger.

pdonahue-ventana (approved):
👍

Issue #889, opened 2023-09-30
Triggers for interrupts are unclear when AIA is added


jhauser-us:
Section 5.6.14, "Interrupt Trigger", is unclear about how to respond to interrupts that may be injected into S or VS mode using new features added by extension Smaia/Ssaia, the RISC-V Advanced Interrupt Architecture (AIA).

Let's first ignore the AIA and just look at the hypervisor (H) extension, which predates the AIA.  The H extension defines interrupt numbers 2, 6, and 10 as VSSI, VSTI, and VSEI at M and S/HS levels, but when delegated down to VS mode, these get translated to interrupt numbers 1, 5, and 9 respectively, which are SSI, STI, and SEI.  If I want to have a trigger on an STI (interrupt number 5) at VS level, am I right to assume I must set bit 6 (VSTI) in `tdata2`?  Or should I set bit 5 (STI) in `tdata2`?  I think the answer is intended to be bit 6 (VSTI), but some clarification is in order.

The AIA upends the Debug spec's interrupt triggers by adding the ability to inject interrupts into lesser privilege modes that completely bypass `mip` and `mie`.  For example, using register `hvictl`, a hypervisor can inject into VS mode an interrupt number 5 that has no connection to bits 5 or 6 in `mip`/`mie`.  If I want to have a trigger for interrupt number 5 at VS level caused by `hvictl`, should I again set bit 6 (VSTI) in `tdata2`, or should I set bit 5 (STI)?

If consistency is maintained with past practice and bit 6 of `tdata2` always indicates interrupt number 5 at VS level, then what about the case when a hypervisor injects into VS mode interrupt number 6?  (This is something a hypervisor may reasonably do, if, for example, it's emulating the H extension for a nested hypervisor running as a guest in VS mode.)  To configure a trigger for interrupt number 6 at VS level, what bit should I set for that in `tdata2`, if it's been decreed that bit 6 in `tdata2`means interrupt number 5 at VS level?

On a related note, if AIA registers `mvien` and `mvip` are used to inject an arbitrary interrupt number X into S level (no involvement of the H extension here), then whether interrupt X is enabled at S level has nothing to do with bit X in `mie`.  But Section 5.6.14 says:
> The trigger only fires if the hart takes a trap because of the interrupt. (E.g. it does not fire when a timer interrupt occurs but that interrupt is not enabled in `mie`.)

[...]

PR #903, opened 2023-10-11
AR: Clarify itrigger and trigger number translation

timsifive:
itrigger numbers relate to the value written to *cause when the trap is taken.

Addresses #889.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -1132,18 +1132,18 @@
     <register name="Interrupt Trigger" short="itrigger" address="0x7a1">
         This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 4.
 
-        This trigger may fire on any of the interrupts configurable in \Rmie
-        (described in the Privileged Spec) or the NMI. The interrupts to fire on are
-        configured by setting the same bit in \RcsrTdataTwo as would be set in
-        \Rmie to enable the interrupt.
+        This trigger can fire when an interrupt trap is taken.
+
+        It can be enabled for individual interrupt numbers by setting the bit
+        corresponding to the interrupt number in \RcsrTdataTwo. The interrupt
+        number is interpreted in the mode that the trap handler executes in.
+        (E.g. virtualized interrupt numbers are not the same in every mode.)
+        In addition the trigger can be enabled for non-maskable interrupts using
+        \FcsrItriggerNmi.
 
         Hardware may only support a subset of interrupts for this trigger.  A
         debugger must read back \RcsrTdataTwo after writing it to confirm the
         requested functionality is actually supported.
-
-        The trigger only fires if the hart takes a trap because of the
-        interrupt. (E.g.\  it does not fire when a timer interrupt occurs but that
-        interrupt is not enabled in \Rmie.)
 
         When the trigger matches, it fires after the trap occurs, just before
         the first instruction of the trap handler is executed.  If


xml/hwbp_registers.xml
```diff
@@ -1132,18 +1132,20 @@
...
-        (described in the Privileged Spec) or the NMI. The interrupts to fire on are
-        configured by setting the same bit in \RcsrTdataTwo as would be set in
-        \Rmie to enable the interrupt.
+        This trigger may fire on any of the interrupts described in the
+        Privileged Spec, or the NMI. The interrupts to fire on are configured by
+        setting the bit in \RcsrTdataTwo whose index would be written to the
+        Exception Code field of the mode-relevant {\tt cause} register when the
+        trap is taken. For interrupts whose numbers are translated, that means
+        the number checked is the one in the mode that the trap handler executes
+        in.
```
pdonahue-ventana, line 1141: I don't think that this sentence will make sense to people who aren't familiar with AIA.  I would say something like:
```suggestion
        trap is taken. For virtualized interrupts that have different interrupt numbers
        when viewed from different modes, that means
        the number checked is the one in the mode that the trap handler executes
        in.
```

xml/hwbp_registers.xml
```diff
@@ -1132,18 +1132,20 @@
...
 
         Hardware may only support a subset of interrupts for this trigger.  A
         debugger must read back \RcsrTdataTwo after writing it to confirm the
         requested functionality is actually supported.
 
         The trigger only fires if the hart takes a trap because of the
-        interrupt. (E.g.\  it does not fire when a timer interrupt occurs but that
-        interrupt is not enabled in \Rmie.)
+        interrupt. (E.g.\ it does not fire for an interrupt that is masked.)
```
pdonahue-ventana, line 1148: If possible, I like to avoid "masked" because sometimes you have positive masks `(A & mask)` and sometimes you have negative masks `(A & ~mask)` so it can be ambiguous.  What you have is probably not that ambiguous but how about using something closer to the original language?  That also applies to global enable (mstatus.MIE) rather than just local enables (mie).
```suggestion
        interrupt. (E.g.\ it does not fire for an interrupt that is not enabled.)
```

xml/hwbp_registers.xml
pdonahue-ventana: Do we also need to say something about how we don't care why the interrupt is taken?  It could be that the input wire asserted, it could be that M mode wrote mie.STIE=1 in the original priv arch, it could be that the hypervisor is playing with hvip, etc.  This was one of John's concerns.  We don't have to spell out the cases but we should say something about how the trigger is based on the interrupt trap regardless of the original cause of the interrupt.

timsifive: I kind of agree with you, but I'm not sure how to say that. Aren't interrupt traps by definition taken in response to interrupts?

xml/hwbp_registers.xml
timsifive: disabled? Done.

xml/hwbp_registers.xml
pdonahue-ventana: I think that John's question was whether one category of reasons for taking an interrupt would be treated differently than another category of reasons for taking an interrupt.

Rereading what's currently there, it basically says that the trigger matches/fires only if we trap but it doesn't say that it fires every time that we trap.  Maybe it's as simple as saying "The trigger fires if and only if the hart takes a trap because of the interrupt."  That would presumably clarify that the reason for the trap doesn't matter.

timsifive:
I've rewritten this, hopefully to be more clear.

xml/hwbp_registers.xml
```diff
@@ -1132,19 +1132,20 @@
     <register name="Interrupt Trigger" short="itrigger" address="0x7a1">
         This register is accessible as \RcsrTdataOne when \FcsrTdataOneType is 4.
 
-        This trigger may fire on any of the interrupts configurable in \Rmie
-        (described in the Privileged Spec) or the NMI. The interrupts to fire on are
-        configured by setting the same bit in \RcsrTdataTwo as would be set in
-        \Rmie to enable the interrupt.
+        For interrupts where it is enabled, this trigger fires when an interrupt
+        trap is taken from a privilege mode where the trigger is enabled.
```
pdonahue-ventana, line 1136: I wasn't sure what the first "it" is.  It seems like it refers to an interrupt but singular "it" doesn't match "interrupts" plural.  Then I wondered if it would be clearer to just remove everything before the comma.

timsifive: How about:
This trigger fires when an interrupt trap is taken from a privilege mode
where the trigger is enabled and that interrupt is also enabled in the
trigger.

pdonahue-ventana: Now I understand.  There are a lot of enables here: the mode bits in tdata1, the interrupt number in tdata2, and the interrupt enables (both mie and mstatus.MIE).  I like something more like this tweak to the earlier language:

        For an interrupt where that interrupt number is enabled in tdata2, this trigger fires when an interrupt
        trap is taken from a privilege mode where the trigger is enabled.

That doesn't actually talk about the interrupt enables but that's implicit when you talk about taking a trap.

timsifive: I don't want to explicitly call out tdata2 here, because nmi is also an option. I agree there are a lot of enables. Maybe I just need to capitulate and make a numbered list.

timsifive: I made it even simpler. We don't need to talk about the mode at all in this sentence. It's already in the bits. The only real need is to indicate what the trigger is useful for. 2 paragraphs down (When the trigger matches...) already makes it clear that it only happens when a trap is actually taken. We don't need the parenthetical about interrupts being masked or whatever. The trigger fires when the trap is taken. Done.

pdonahue-ventana (approved):
That's simpler but still conveys all the necessary information.  Looks good.

PR #906, opened 2023-10-18
AR: Comment itrigger limited to 32 ints if XLEN=32

### LINKED ISSUE #904: For MXLEN=32, can't trigger on interrupts 32 or higher
The RISC-V Advanced Interrupt Architecture provides for the possibility of "major" interrupts up to number 63, even when XLEN = 32.  The AIA itself allocates interrupt numbers 35 and 43 for low-priority and high-priority RAS events, and other RISC-V extensions may define other interrupt numbers in the range of 32 to 47.

When MXLEN/DXLEN = 32, it appears that an interrupt trigger cannot select interrupt numbers 32 and higher to match on, because `tdata2` is only 32 bits wide.  This limitation probably deserves a comment at a minimum.

timsifive:
Resolves #904.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -1141,6 +1141,14 @@
         In addition the trigger can be enabled for non-maskable interrupts using
         \FcsrItriggerNmi.
 
+        \begin{commentary}
+        If XLEN is 32, then it is not possible to set a trigger for interrupts
+        with Exception Code larger than 31. A future version of the RISC-V
+        Privileged Spec will likely define interrupt Exception Codes 32 through
+        47. Some of those numbers are already being used by the RISC-V Advanced
+        Interrupt Architecture.
+        \end{commentary}
+
         Hardware may only support a subset of interrupts for this trigger.  A
         debugger must read back \RcsrTdataTwo after writing it to confirm the
         requested functionality is actually supported.


pdonahue-ventana (commented):
I don't think that the priv spec reserves 32-47.  I think that it's AIA that does so.  The priv spec says that >= 16 are designated for platform use (both in [table 3.6](https://github.com/riscv/riscv-isa-manual/blob/main/src/machine.adoc#machine-cause-register-mcause) and [table 9.6](https://github.com/riscv/riscv-isa-manual/blob/main/src/hypervisor.adoc#trap-cause-codes)).

timsifive:
You're right. I must have been looking at the exceptions when I wrote that.

So this is another in-future-versions-of-the-spec thing. Getting really tired of updating our spec to work with a document that doesn't even exist in draft form.

xml/hwbp_registers.xml
```diff
@@ -1141,6 +1141,14 @@
         In addition the trigger can be enabled for non-maskable interrupts using
         \FcsrItriggerNmi.
 
+        \begin{commentary}
+        If XLEN is 32, then it is not possible to set a trigger interrupts with
+        Exception Code larger than 31. A future version of the RISC-V Privileged
```
pdonahue-ventana, line 1146: Parser error: "a trigger interrupts" doesn't seem right to me.  Maybe this?
```suggestion
        If XLEN is 32, then it is not possible to set a trigger for interrupts with
        Exception Code larger than 31. A future version of the RISC-V Privileged
```

pdonahue-ventana (approved):

PR #908, opened 2023-10-20
The last field of a register was not printed

en-sc:
`riscv_debug_reg_fields_to_s()` exited early without processing the last field.

### DIFF
--- a/debug_reg_printer.c
+++ b/debug_reg_printer.c
@@ -61,11 +61,13 @@ static uint64_t riscv_debug_reg_field_value(riscv_debug_reg_field_info_t field, 
 }
 
 static unsigned int riscv_debug_reg_fields_to_s(char *buf, unsigned int offset,
-		riscv_debug_reg_field_list_t list, riscv_debug_reg_ctx_t context, uint64_t value)
+	struct riscv_debug_reg_field_list_t (*get_next)(riscv_debug_reg_ctx_t contex),
+	riscv_debug_reg_ctx_t context, uint64_t value)
 {
 	unsigned int curr = offset;
 	curr += get_len_or_sprintf(buf, curr, " { ");
-	for (; list.get_next; list = list.get_next(context)) {
+	for (struct riscv_debug_reg_field_list_t list; get_next; get_next = list.get_next) {
+		list = get_next(context);
 		curr += riscv_debug_reg_field_to_s(buf, curr, list.field, context,
 				riscv_debug_reg_field_value(list.field, value));
 		curr += get_len_or_sprintf(buf, curr, ", ");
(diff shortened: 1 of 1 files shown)


debug_reg_printer.c
```diff
@@ -61,11 +61,13 @@ static uint64_t riscv_debug_reg_field_value(riscv_debug_reg_field_info_t field,
...
 static unsigned int riscv_debug_reg_fields_to_s(char *buf, unsigned int offset,
-		riscv_debug_reg_field_list_t list, riscv_debug_reg_ctx_t context, uint64_t value)
+	struct riscv_debug_reg_field_list_t (*get_next)(riscv_debug_reg_ctx_t contex),
+	riscv_debug_reg_ctx_t context, uint64_t value)
 {
 	unsigned int curr = offset;
 	curr += get_len_or_sprintf(buf, curr, " { ");
-	for (; list.get_next; list = list.get_next(context)) {
+	for (struct riscv_debug_reg_field_list_t list; get_next; get_next = list.get_next) {
+		list = get_next(context);
```
timsifive, line 70: I see how the old code would omit the last field, because when list.get_next is NULL you still need to print out one more field.

debug_reg_printer.c
```diff
@@ -61,11 +61,13 @@ static uint64_t riscv_debug_reg_field_value(riscv_debug_reg_field_info_t field,
 }
 
 static unsigned int riscv_debug_reg_fields_to_s(char *buf, unsigned int offset,
-		riscv_debug_reg_field_list_t list, riscv_debug_reg_ctx_t context, uint64_t value)
+	struct riscv_debug_reg_field_list_t (*get_next)(riscv_debug_reg_ctx_t contex),
+	riscv_debug_reg_ctx_t context, uint64_t value)
```
timsifive, line 65: Why do you need to pass a function get_next instead of the list?

Also, I'm surprised that there are no callers to this function that needed to be updated.

en-sc: > Why do you need to pass a function get_next instead of the list?

IMHO it makes the loop more concise. If one were to transition to passing a list node, the loop would be required to be a `do ... while ()` since there would always be at least one node. I find `for` loops easier to read. Don't you agree?

> Also, I'm surprised that there are no callers to this function that needed to be updated.

This is a static function, only called from `riscv_debug_reg_to_s()`. The appropriate adjustment (transition from passing a list node to passing a function pointer which generates such node) is made [on line 90](https://github.com/riscv/riscv-debug-spec/pull/908/files#diff-cc411bb1fa083f40d43064106679d2a9d0ed6ac8f175866c048c0814d794497dR90).

timsifive: I missed the github info showing me that line 90 was a different function. Thanks for explaining. This is fine.

timsifive (approved):

PR #909, opened 2023-10-30
Make register dump output more concise.

timsifive:
Don't mention fields whose value is 0.
New format:
{sbautoincrement=1 sbaccess=64bit sbreadonaddr=1 }

### DIFF
--- a/debug_reg_printer.c
+++ b/debug_reg_printer.c
@@ -62,22 +62,35 @@ static uint64_t riscv_debug_reg_field_value(riscv_debug_reg_field_info_t field, 
 
 static unsigned int riscv_debug_reg_fields_to_s(char *buf, unsigned int offset,
 	struct riscv_debug_reg_field_list_t (*get_next)(riscv_debug_reg_ctx_t contex),
-	riscv_debug_reg_ctx_t context, uint64_t value)
+	riscv_debug_reg_ctx_t context, uint64_t value,
+	enum riscv_debug_reg_show show)
 {
 	unsigned int curr = offset;
-	curr += get_len_or_sprintf(buf, curr, " { ");
+	curr += get_len_or_sprintf(buf, curr, " {");
+	char *separator = "";
 	for (struct riscv_debug_reg_field_list_t list; get_next; get_next = list.get_next) {
 		list = get_next(context);
-		curr += riscv_debug_reg_field_to_s(buf, curr, list.field, context,
-				riscv_debug_reg_field_value(list.field, value));
-		curr += get_len_or_sprintf(buf, curr, ", ");
+
+		uint64_t field_value = riscv_debug_reg_field_value(list.field, value);
+
+		if ((show == RISCV_DEBUG_REG_SHOW_ALL) ||
+				(show == RISCV_DEBUG_REG_HIDE_UNNAMED_0 &&
+					(field_value != 0 ||
+						(list.field.values && list.field.values[0]))) ||
+				(show == RISCV_DEBUG_REG_HIDE_ALL_0 && field_value != 0)) {
+			curr += get_len_or_sprintf(buf, curr, separator);
+			curr += riscv_debug_reg_field_to_s(buf, curr, list.field, context,
+							field_value);
+			separator = " ";
+		}
 	}
 	curr += get_len_or_sprintf(buf, curr, "}");
 	return curr - offset;
 }
 
 unsigned int riscv_debug_reg_to_s(char *buf, enum riscv_debug_reg_ordinal reg_ordinal,
-		riscv_debug_reg_ctx_t context, uint64_t value)
+		riscv_debug_reg_ctx_t context, uint64_t value,
+		enum riscv_debug_reg_show show)
 {
 	unsigned int length = 0;
 
@@ -88,7 +101,7 @@ unsigned int riscv_debug_reg_to_s(char *buf, enum riscv_debug_reg_ordinal reg_or
 
 	if (reg.get_fields_head)
 		length += riscv_debug_reg_fields_to_s(buf, length,
-				reg.get_fields_head, context, value);
+				reg.get_fields_head, context, value, show);
 
 	if (buf)
 		buf[length] = '\0';
--- a/registers.py
+++ b/registers.py
@@ -242,7 +242,12 @@ class Field( object ):
 
     def c_values_array_def(self):
         assert len(self.values)
-        arr_elem_def = (f'[{v.value}] = "{v.name}"' for v in self.values if v.value is not None)
+
+        # Remove whitespace from the names, so when they're displayed we can use
+        # only ' ' as a separator.
+        arr_elem_def = (f'[{v.value}] = "{toCIdentifier(v.name)}"'
+                        for v in self.values if v.value is not None)
+
         #WA for *lo & *hi splitted fields
         if len(self.values) > 2**self.length():
             return f"static const char *{self.c_values_array_name()}[{2**self.length()}] = {{}};"
--- a/debug_reg_printer.h
+++ b/debug_reg_printer.h
@@ -2,6 +2,12 @@
 
 #include "debug_defines.h"
 
+enum riscv_debug_reg_show {
+	RISCV_DEBUG_REG_SHOW_ALL,
+	RISCV_DEBUG_REG_HIDE_ALL_0,
+	RISCV_DEBUG_REG_HIDE_UNNAMED_0,
+};
+
 /**
  * This function is used to fill a buffer with a decoded string representation
  * of register's value.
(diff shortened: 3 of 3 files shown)


timsifive:
@kr-sc @en-sc

pdonahue-ventana (approved):
OK

debug_reg_printer.c
```diff
@@ -46,6 +46,9 @@ static unsigned int riscv_debug_reg_field_to_s(char *buf, unsigned int offset,
 		riscv_debug_reg_field_info_t field, riscv_debug_reg_ctx_t context,
 		uint64_t field_value)
 {
+	if (field_value == 0)
```
en-sc, line 49: Wouldn't it be preferable to only skip value if it is zero and there is no name for such field value?
Like so:
```
static const char *get_field_value_name(riscv_debug_reg_field_info_t field,
                uint64_t field_value)
{
        return field.values ? field.values[field_value] : NULL;
}
```
```suggestion
	if (field_value == 0  && !get_field_value_name(list.field, field_value))
```

debug_reg_printer.c
```diff
@@ -65,12 +68,15 @@ static unsigned int riscv_debug_reg_fields_to_s(char *buf, unsigned int offset,
...
+	curr += get_len_or_sprintf(buf, curr, " {");
 	for (struct riscv_debug_reg_field_list_t list; get_next; get_next = list.get_next) {
 		list = get_next(context);
-		curr += riscv_debug_reg_field_to_s(buf, curr, list.field, context,
+		unsigned int len = riscv_debug_reg_field_to_s(buf, curr, list.field, context,
 				riscv_debug_reg_field_value(list.field, value));
-		curr += get_len_or_sprintf(buf, curr, ", ");
+		if (len) {
+			curr += len;
+			curr += get_len_or_sprintf(buf, curr, " ");
```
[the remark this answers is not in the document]
en-sc: Wouldn't you also like to skip the last separator (the one before the last field)? If so, I would suggest to move the check out of `riscv_debug_reg_field_to_s()` and rewrite the loop like so:
```
        const char *separator = "";
        for (struct riscv_debug_reg_field_list_t list; get_next; get_next = list.get_next) {
                list = get_next(context);
                if (<condition to skip the field>)
                        continue;
                curr += get_len_or_sprintf(buf, curr, separator);
                curr += riscv_debug_reg_field_to_s(buf, curr, list.field,
                                context, field_value);
                separator = ", ";
        }
```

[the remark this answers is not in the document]
timsifive: > I would suggest to leave the comma between the fields, since some value names include a space e.g. not supported.

Good point. I prefer to solve that by replacing whitespace in value names with _.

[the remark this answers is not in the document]
timsifive: > Wouldn't you also like to skip the last separator (the one before the last field)?

Yes, but I thought that was a pre-existing issue and it didn't bother me that much. I'll address it.

debug_reg_printer.c
timsifive: I don't think that's preferable. Almost everywhere the 0 value means inactive, so not interesting to list. 
E.g. dmstatus=0 would be `{version=none stickyunavail=current ndmresetpending=false confstrptrvalid=invalid authbusy=ready authenticated=false}` which is a lot to visually parse through just to see nothing is going on. And there's nothing preventing the XML to be edited in the future to be more consistent or whatever and to name every 0 field, in which case the output would be even larger.

en-sc: > I don't think that's preferable. Almost everywhere the 0 value means inactive, so not interesting to list.

The issue, in my opinion is that it is exactly **almost** everywhere. E.g. `dtmcs.version == 0 ` means `0.11` which is quite important information, considering the use of this code in OpenOCD's `riscv-013.c`.

> E.g. dmstatus=0 would be `{version=none stickyunavail=current ndmresetpending=false confstrptrvalid=invalid authbusy=ready authenticated=false}` which is a lot to visually parse through just to see nothing is going on.

IMHO this particular issue is caused by attempting to get a meaningful result from decoding a meaningless value and should be addressed in OpenOCD code. Please, take a look at [this PR](https://github.com/riscv/riscv-openocd/pull/951).

timsifive: Ideally we could mark the handful of fields where 0 is significant, but I'm not going to do that work right now. It's much better to have the output be concise than it is to be readable by someone with no experience.

I chose the 0 example because it was easy. The problem is just as bad in a realistic example:
```dmstatus=0x430382 {version=0.13 anyresumeack=1 allresumeack=1 impebreak=1 stickyunavail=current ndmresetpending=false confstrptrvalid=invalid authbusy=ready authenticated=true anyhalted=1 allhalted=1}```
vs
```dmstatus=0x430382 {version=0.13 anyresumeack=1 allresumeack=1 impebreak=1 authenticated=true anyhalted=1 allhalted=1}```

The former has a bunch of uninteresting fields that just make it harder to read the ones you usually care about.

On top of that, the XML might change in the future to add more explicit true/false encodings, which would make the problem worse. The contents of that XML are primarily for the formatting of the document, not for this automated field generation.

Comment (aap-sc) [reply debug_reg_printer.c]: @timsifive 

> Ideally we could mark the handful of fields where 0 is significant,
> It's much better to have the output be concise than it is to be readable by someone with no experience.


The former seems an unfeasible goal since it won't be easy to maintain (you've actually alluded to that already). The latter seems like an aesthetic preference  - there are people who prefer verbose printouts and there are people who hate these logs and want them to be as short as possible (me included). However, I can see the reason some may want to decode every field. I'd like to suggest a compromise: we can add an extra argument to the printing function and a corresponding riscv-specific OpenOCD command to control it. So we can enable verbose decoding of fields. I can even volunteer to provide a patch (if no objections).

File: registers.py [hunk 3]
```diff
@@ -242,7 +242,13 @@ def c_values_array_name(self):
 
     def c_values_array_def(self):
         assert len(self.values)
-        arr_elem_def = (f'[{v.value}] = "{v.name}"' for v in self.values if v.value is not None)
+
+        # Remove whitespace from the names, so when they're displayed we can use
+        # only ' ' as a separator.
+        def remove_whitespace(s):
```
en-sc, line 248: `toCIdentifier()` can be used here, if you like.

debug_reg_printer.c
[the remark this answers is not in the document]
en-sc: Thanks!

registers.py
timsifive: That works. Thanks.

debug_reg_printer.c
timsifive: OK. Done.

debug_reg_printer.c
```diff
@@ -62,22 +62,33 @@ static uint64_t riscv_debug_reg_field_value(riscv_debug_reg_field_info_t field,
...
 	for (struct riscv_debug_reg_field_list_t list; get_next; get_next = list.get_next) {
 		list = get_next(context);
-		curr += riscv_debug_reg_field_to_s(buf, curr, list.field, context,
-				riscv_debug_reg_field_value(list.field, value));
-		curr += get_len_or_sprintf(buf, curr, ", ");
+
+		uint64_t field_value = riscv_debug_reg_field_value(list.field, value);
+
+		if ((show == RISCV_DEBUG_REG_SHOW_ALL) ||
+				(show == RISCV_DEBUG_REG_HIDE_UNNAMED_0 && list.field.values) ||
```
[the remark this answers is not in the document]
timsifive: Good catch. Thanks.

debug_reg_printer.c
```diff
@@ -62,22 +62,34 @@ static uint64_t riscv_debug_reg_field_value(riscv_debug_reg_field_info_t field,
...
 		list = get_next(context);
-		curr += riscv_debug_reg_field_to_s(buf, curr, list.field, context,
-				riscv_debug_reg_field_value(list.field, value));
-		curr += get_len_or_sprintf(buf, curr, ", ");
+
+		uint64_t field_value = riscv_debug_reg_field_value(list.field, value);
+
+		if ((show == RISCV_DEBUG_REG_SHOW_ALL) ||
+				(show == RISCV_DEBUG_REG_HIDE_UNNAMED_0 &&
+					(field_value != 0 || list.field.values)) ||
```
en-sc, line 78: This will not hide an unnamed zero if there is a name for any other value (`list.field.values == {[0] = NULL, [1] = "some name"}`)

timsifive: Fixed, I think. I don't see any examples of this.

en-sc: > This will not hide an unnamed zero if there is a name for any other value (`list.field.values == {[0] = NULL, [1] = "some name"}`)

For an example, please, take a look at `csr_dcsr_cause_values` (`[0]` is zero-initialized by default).

en-sc:
LGTM.

en-sc (approved):

PR #911, opened 2023-11-03
Invalid addresses might not match.

timsifive:
See:
* https://lists.riscv.org/g/tech-debug/message/1407
* https://lists.riscv.org/g/tech-debug/files/meetings%202023/2023-01-10%20Debug%20Task%20Group%20Call%20Notes.pdf
* https://lists.riscv.org/g/tech-debug/files/meetings%202023/2023-01-10%20Debug%20Task%20Group%20Call%20Slides.pdf

### DIFF
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -319,6 +319,25 @@ Cache operations encoded as HINTs do not match debug triggers.
     \end{steps}
 \end{commentary}
 
+\subsection{Invalid Virtual Addresses}
+
+For virtual address matches without a mask, \RcsrTdataTwo must be able to hold
+all valid virtual addresses but it need not be capable of holding other values.
+Implementations may convert an invalid virtual address to a different invalid
+virtual address before comparing the address to \RcsrTdataTwo.
+
+\begin{commentary}
+    A straightforward trigger implementation would compare the effective address
+    to the contents of \RcsrTdataTwo. This is most intuitive to a user setting
+    triggers. However, the Privileged Spec makes various accommodations that
+    allow a shorter representation of invalid virtual addresses, and the Debug
+    Spec should do the same.  If an implementation converts some invalid
+    addresses to other invalid addresses, then it is impossible to compare
+    against the original effective address.  To simplify the spec and allow for
+    more implementations, we don't specify which version is compared for an
+    invalid virtual address.
+\end{commentary}
+
 \section{Multiple State Change Instructions} \label{sec:multistate}
 
 An instruction that performs multiple architectural state changes (e.g.,
--- a/Sdtrig.tex
+++ b/Sdtrig.tex
@@ -319,24 +319,34 @@ Cache operations encoded as HINTs do not match debug triggers.
     \end{steps}
 \end{commentary}
 
-\subsection{Invalid Virtual Addresses}
+\subsection{Address Matches}
 
-For virtual address matches without a mask, \RcsrTdataTwo must be able to hold
-all valid virtual addresses but it need not be capable of holding other values.
-Implementations may convert an invalid virtual address to a different invalid
-virtual address before comparing the address to \RcsrTdataTwo.
+For address matches without a mask, \RcsrTdataTwo must be able to hold all valid
+addresses in all supported translation modes. That means that after writing any
+of these valid addresses, the exact same value XLEN-wide value is read back,
+including any high bits. An implementation may be able to optimize the storage
+required, depending on the widest addresses it supports.
 
 \begin{commentary}
-    A straightforward trigger implementation would compare the effective address
-    to the contents of \RcsrTdataTwo. This is most intuitive to a user setting
-    triggers. However, the Privileged Spec makes various accommodations that
-    allow a shorter representation of invalid virtual addresses, and the Debug
-    Spec should do the same.  If an implementation converts some invalid
-    addresses to other invalid addresses, then it is impossible to compare
-    against the original effective address.  To simplify the spec and allow for
-    more implementations, we don't specify which version is compared for an
-    invalid virtual address.
+    If physical addresses are less than XLEN bits wide, they are zero-extended.
+    If virtual addresses are less than XLEN bits wide, they are sign-extended.
+    \RcsrTdataTwo must be implemented with enough bits of storage to represent
+    the full range of supported physical and virtual address values when read by
+    software and used by hardware.
 \end{commentary}
+
+\subsubsection{Invalid Addresses}
+
+If \RcsrTdataTwo can hold any invalid addresses, then writes of an
+invalid address that can not be represented as-is should be converted to
+a different invalid address that can be represented.
+
+For invalid instruction fetch addresses and load and store effective addresses,
+the compare value may be changed to a different invalid address.
+
+In addition, an implementation may choose to inhibit all trigger matching
+against invalid addresses, especially if there is no support for storage of any
+invalid address values in tdata2.
 
 \section{Multiple State Change Instructions} \label{sec:multistate}
 
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -310,12 +310,6 @@
         Table~\ref{tab:hwbp_timing}, both timings should be supported on load
         address triggers.
 
-        This trigger type may be limited to address comparisons (\FcsrMcontrolSelect is
-        always 0) only. If that is the case and masking is not supported (match
-        values 4, 5, 12, 13), then \RcsrTdataTwo must be able to
-        hold all valid virtual addresses but it need not be capable of holding
-        other values.
-
         The Privileged Spec says that breakpoint exceptions that occur on
         instruction fetches, loads, or stores update the {\tt tval} CSR
         with either zero or the faulting virtual address.  The faulting
@@ -666,12 +660,6 @@
 
         A chain of triggers must only fire if every trigger in the chain was
         matched by the same instruction.
-
-        This trigger type may be limited to address comparisons (\FcsrMcontrolSixSelect is
-        always 0) only. If that is the case and masking is not supported (match
-        values 4, 5, 12, 13), then \RcsrTdataTwo must be able to
-        hold all valid virtual addresses but it need not be capable of holding
-        other values.
 
         The Privileged Spec says that breakpoint exceptions that occur on
         instruction fetches, loads, or stores update the {\tt tval} CSR


aap-sc:
> That means it may not be possible to perform address matches on invalid virtual addresses.

**SDext** spec does not define what an `invalid virtual address` means. I guess it would not hurt to add a clarification

timsifive:
I can't find a definition of "invalid virtual address" in the privspec either. I don't want to define this in the debug spec because we might get it wrong. Better to refer people to the privspec.

timsifive:
This actually leaves out the part that invalid virtual address matches might not occur at all. That just seems to add an arbitrary exception that isn't necessary. I don't feel strongly about this because invalid virtual addresses are rare. It just feels strange to make the exception.

Sdtrig.tex
```diff
@@ -319,6 +319,22 @@ \subsection{Cache Operations}
     \end{steps}
 \end{commentary}
 
+\subsection{Invalid Virtual Addresses}
+
+For virtual address matches without a mask, \RcsrTdataTwo must be able to hold
+all valid virtual addresses but it need not be capable of holding other values.
+Implementations may convert invalid virtual addresses to a different invalid
+virtual address before comparing the address to \RcsrTdataTwo.
```
pdonahue-ventana, line 327: Two comments:
1. It's better to say it converts "an invalid address" (singular) to another (singular) invalid address, similar to the priv spec.  That avoids any implication that all invalid addresses must convert to one single value.
2. It should say that the value observed on tdata2 reads goes through this same transformation.

Sdtrig.tex
```diff
@@ -319,6 +319,22 @@ \subsection{Cache Operations}
...
+
+\begin{commentary}
+    A straightforward trigger implementation would compare the effective address
+    to the contents of \RcsrTdataTwo. This is most intuitive to a user setting
+    triggers. However, for instruction fetches the effective address might not
+    be preserved in the hart's state so it is impossible to compare against the
+    effective address. To simplify the spec and allow for more implementations,
```
pdonahue-ventana, line 334: I would say something more like "The PC might not be capable of holding all invalid addresses and some invalid addresses may map to other invalid addresses, making it impossible to compare against the original value that software attempted to write to the PC."  (I intentionally tried to avoid saying "jump" because you could jump there, you could mret/sret there, you could be right near the edge and branch there, you could sequentially execute off the end of the address space, you could be in Sv39 mode at a legal negative address and then write satp to put yourself in Bare mode where it's suddenly illegal to have ones in the top PC bits, etc.)

Sdtrig.tex
aap-sc: > It should say that the value observed on tdata2 reads goes through this same transformation.

1. Can we leave tdata2 alone? This gets extremely confusing. We have several addresing modes that can change during execution: `bare`, `sv39`, etc.. How do you expect the value of `tdata2` be changed if the execution context is not constatnt?

2.As a user of the trigger I want to be able to program the trigger with whatever value I want. Moreover, the existing trigger enumeration specification imply some `tdata2` values may not be supported. That is availability of trigger already depends on `tdata2`. I would expect that no additional changes to the value written in observed.

aap-sc:
@timsifive , @pdonahue-ventana to be fair, I don't understand the motivation behind this change. 

1. Can we revert all this and keep it simple? if the value of `tdata2`  can be written - to the trigger it is match. If some implementation does not support matching on an "invalid address' (whatever this means, we don't have it defined anywhere) should't implementation detect such case during trigger enumeration procedure?
2.  I find an argument "we don't see invalid addresses that often" not very convincing. We use debugger when something gone wrong. Registers gets corrupted, stack may be overwritten, etc. User can have some miscompilation and say experiment with some extensions which use to bits of address (similar to TBI from ARM or HW tagging). I would expect debugger to be useful in such scenarious.

I can provide these points in the thread mentioned ( https://lists.riscv.org/g/tech-debug/message/1407 ), though I find it somewhat difficult to follow.

Sdtrig.tex
timsifive: Done.

timsifive:
@aap-sc This is not intended to be a change. It's merely spelling out the result of the fact that tdata2 need only be able to represent all valid virtual addresses, and how that interacts with invalid virtual addresses (which are mentioned but not clearly defined in the privileged spec).

In what scenario does the spec prescribe different behavior now?

aap-sc:
> In what scenario does the spec prescribe different behavior now?

My bad. I've re-read the current text in this MR and it is really clear now.

pdonahue-ventana (commented):
I don't think that it says it, but this appears to allow the implementation where invalid addresses never match on match=0 or 8.  If tdata2 holds an invalid address then you can always convert that to some other invalid address that doesn't match the current access.  It probably also works for match=1 or 9 unless your maskmax6 allows the napot region to exactly match all invalid addresses.

However, the greater-than/less-than matching (match=2 or 3) still seems like it might require matching on invalid addresses.  For example, say that we're in Bare mode and match=2 (ge) and tdata2=0x1000.  All addresses from 0x000-0xfff are legal.  All illegal addresses are above 0x1000 and should match this trigger even if you remap one illegal address to another.  I don't remember the final conclusion on this point from the meeting but I wanted to point it out to @bcstrongx.

I'm generally OK with this change but want to let it sit for a day or two to gather any additional input.

timsifive:
I agree that this PR, as written currently, implies that invalid virtual addresses still match on some triggers.

IMO the decision whether to wholesale say that invalid virtual addresses might not match triggers at all is orthogonal to this PR, which is merely explaining in more detail what the existing spec leaves unsaid.

Sdtrig.tex
```diff
@@ -319,6 +319,32 @@ \subsection{Cache Operations}
     \end{steps}
 \end{commentary}
 
+\subsection{Invalid Virtual Addresses}
+
+\begin{steps}{For address matches without a mask, \RcsrTdataTwo must be able to:}
+    \item Hold all valid physical addresses (which are zero extended).
+    \item Hold all valid virtual addresses (which are sign extended).
+\end{steps}
```
pdonahue-ventana, line 327: It only needs to support these sign-extended virtual addresses if Sv* is supported.  Also, the sign extension is a RV64 thing.  An M-mode-only RV32 implementation only needs to be able to hold PAs.  A RV32 implementation with Sv32 needs 32 flops and there is no sign extension.

Could this be simplified to say that it needs to be able to hold all valid addresses in all translation modes?  Then maybe some non-normative thing that explains that this means all PAs and all VAs in all supported Sv* modes.

timsifive: I've restructured this like you suggested. Not quite sure about the commentary I wrote.

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
...
+For address matches without a mask, \RcsrTdataTwo must be able to hold all valid
+addresses in all supported translation modes. That means that after writing any
+of these valid addresses, the exact same value is read back, including any high
+bits.
+
+\begin{commentary}
+    If physical addresses are less than XLEN bits wide, they are zero-extended.
+    If virtual addresses are less than XLEN bits wide, they are sign-extended.
+    Be sure to have enough flops in \RcsrTdataTwo to hold all XLEN bits of the
+    full range of both physical and virtual addresses.
```
pdonahue-ventana, line 333: I don't think that you actually need XLEN flops.  In the example from the email thread you need 49.  And you have to sign or zero extend based on some information in those flops.  So something more like this:
```suggestion
    Be sure to have enough flops in \RcsrTdataTwo to hold the full range of
    both physical and virtual addresses with appropriate zero or sign extension
    when read by hardware or software.
```

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
     \end{steps}
 \end{commentary}
 
+\subsection{Invalid Virtual Addresses}
+
+For address matches without a mask, \RcsrTdataTwo must be able to hold all valid
+addresses in all supported translation modes. That means that after writing any
+of these valid addresses, the exact same value is read back, including any high
```
gfavor, line 326: =>  "the exact same XLEN-wide value is read back."

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
...
+    Be sure to have enough flops in \RcsrTdataTwo to hold all XLEN bits of the
+    full range of both physical and virtual addresses.
+\end{commentary}
+
+If \RcsrTdataTwo can hold any invalid addresses, then writes of an
+invalid address that can not be represented as-is should be converted to
+a different invalid address that can be represented.
+
+For invalid instruction fetch and load and store effective addresses,
+the compare value may be changed to a different invalid address.
```
gfavor, line 341: This is confusing.  It appears to say that an ifetch address or ldst effective address, if invalid, can be changed to a different invalid address value to create the compare value that is compared against te trigger address.  I assume that is not the arch intent trying to be expressed here.

Or if that is the intent, then who would actually choose to remap invalid compare addresses to different invalid values before doing a compare (i.e. no one would).

It seems like all reasonable implementations would simply compare the XLEN-wide ifetch/ldst address with the  XLEN-wide trigger address.  Is there some other implementation that the spec is trying to allow for that would justify being rather vague about exactly hwo copares are performed.

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
...
+
+\begin{commentary}
+    A straightforward trigger implementation would compare the instruction fetch
+    and load and store effective addresses to the contents of \RcsrTdataTwo.
+    This is most intuitive to a user setting triggers. Here we accommodate
+    implementations that internally shorten invalid addresses.
+\end{commentary}
+
+In addition, an implementation may choose to inhibit all trigger matching
+against invalid addresses.
```
gfavor, line 351: ... especially if there is no support for storage of any invalid address values in tdata2.

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
...
+a different invalid address that can be represented.
+
+For invalid instruction fetch and load and store effective addresses,
+the compare value may be changed to a different invalid address.
+
+\begin{commentary}
+    A straightforward trigger implementation would compare the instruction fetch
+    and load and store effective addresses to the contents of \RcsrTdataTwo.
+    This is most intuitive to a user setting triggers. Here we accommodate
+    implementations that internally shorten invalid addresses.
```
gfavor, line 347: "Internally shorten" is not a properly (i.e. explicitly) defined term in this spec; one just hopes that the reader interpets these words exactly as intended.  Plus this is an implementation-level description in what is supposed to be an architecture-level spec.

What I think what is trying to be said is that invalid addresses, when written to tdata2, may be remapped to and stored as a supported invalid address value (if there are any supported invalid address values).  Which has already been said above normatively.

Net, it seems like this note can/should be removed.

Sdtrig.tex
bcstrongx: We discussed a case in the last meeting where, in Sv48, there is a jump to an invalid address, say 0x1234aaaabbbbcccc.  That target address may be converted to another invalid address, presumably one that can be held in *tval (likely 49-bit 0x0aaaabbbbcccc), before it is fetched and compared with triggers.  Are you saying you think this should be a non-compliant implementation?

Sdtrig.tex
timsifive: Sounds good. I'm changing that.

Sdtrig.tex
timsifive: Done.

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
     \end{steps}
 \end{commentary}
 
+\subsection{Invalid Virtual Addresses}
```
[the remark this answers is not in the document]
timsifive: Good point. The original PR was for virtual addresses but then in the discussion it became clear we need it for physical addresses as well. Renamed to "Invalid Addresses".

Sdtrig.tex
timsifive: Restating @bcstrongx's example:
1. Jump to an invalid address 0x1234aaaabbbbcccc.
2. It seems likely the PC cannot represent all possible invalid addresses, so the invalid address is converted to some internal representation.
3. So the instruction fetch must happen at some invalid address, but likely not the one generated by the original instruction.

Since that is allowed, it's reasonable to allow the same for load/stores as well.

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
...
+    If virtual addresses are less than XLEN bits wide, they are sign-extended.
+    Be sure to have enough flops in \RcsrTdataTwo to hold all XLEN bits of the
+    full range of both physical and virtual addresses.
+\end{commentary}
+
+If \RcsrTdataTwo can hold any invalid addresses, then writes of an
+invalid address that can not be represented as-is should be converted to
+a different invalid address that can be represented.
+
+For invalid instruction fetch and load and store effective addresses,
```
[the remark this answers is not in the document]
timsifive: Fixed. Thanks.

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
...
+the compare value may be changed to a different invalid address.
+
+\begin{commentary}
+    A straightforward trigger implementation would compare the instruction fetch
+    and load and store effective addresses to the contents of \RcsrTdataTwo.
```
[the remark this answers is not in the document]
timsifive: Fixed also.

Sdtrig.tex
timsifive: Done.

Sdtrig.tex
gfavor: I believe I get the point.  That any "remapping" of an invalid compare address to another invalid address is more a matter of canonicity decode logic recognizing that the upper bits are non-canonical and one would, for example, just carry forward through the datapath one bit to signify valid vs. invalid address.  Then, when it gets to comparing this with an XLEN-wide tdata2 address value, the implementation could turn the one bit back into some fixed invalid pattern of high bits.

If that's the kind of idea you all have in mind, then that makes sense.  And hence the idea of invalid ifetch/ldst addresses being remappable to other invalid addresses before being compared with tdata2, seems valid and something to allow for.

Although it would seem like a "good" and cheap implementation would remap invalid address values written into tdata2 into an internal representation with an extra bit signifying valid vs. invalid address.  And invalid ifetch/ldst address values would be represented, as mentioned above, with one extra bit.  Then comparisons would indicate a match if either both addresses are valid and fully match all supported address bits, or both addresses are invalid and are then considered to match irrespective of any comparison of the lower supported address bits.  (With this, the address comparator is not XLEN-wide.  Versus converting address values to XLEN-wide values and then comparing them.  Although one can say that these are functionality equivalent implementations if both conversions of invalid addresses to XLEN-wide values do the same expansion.)

This provides a simple easy to understand model to software as to trigger matching behavior wrt invalid addresses.  Personally I would maybe mandate this behavior, but I expect others want a much looser spec that allows other behaviors that an even cheaper (?) implementation might exhibit.  (And, to be clear, an implementation would be allowed to instead never match whenever either or both addresses is/are invalid.)

timsifive: I'd be OK saying an implementation should map invalid addresses to a canonical invalid representation, that can also be represented in tdata2. But they're also allowed to have more flexible less useful behavior (for backwards compatibility) or may not match against invalid addresses at all.

Thoughts?

timsifive:
Since this is taking forever and not super useful, I created https://github.com/riscv/riscv-debug-spec/pull/919 as an alternative, which simply doesn't allow tdata2 to be smaller when only address matching is implemented.

Sdtrig.tex
timsifive: I feel we're getting off-track trying non-normatively specifying desired invalid address encodings.

Let's focus on the normative part of the spec. The language says:
> For invalid instruction fetch addresses and load and store effective addresses,
> the compare value may be changed to a different invalid address.

That's clear to me but it was not clear to @gfavor, although he now understand what we're trying to say. How should this language be changed so its meaning is clear?

Sdtrig.tex
timsifive: I'm removing this commentary.

Sdtrig.tex
pdonahue-ventana: > I'd be OK saying an implementation should map invalid addresses to a canonical invalid representation, that can also be represented in tdata2. But they're also allowed to have more flexible less useful behavior (for backwards compatibility) or may not match against invalid addresses at all.

Validity isn't known a priori.  If you support Sv48 and 48b PA (the example from the emails) then 0x0000FFFFFFFFFFFF is a valid bare mode address but not a valid Sv48 address.  So you don't really know if it's valid until the time of use.  The trigger might be enabled in all modes (M/S/U/VS/VU) so it's valid sometimes (M) and invalid other times (U if satp.MODE=Sv48).  Or satp.MODE can change dynamically which changes the way it's interpreted.

bcstrongx: It looks fine to me.  Do we want to say anything (non-normatively) about how, ideally, the conversion of invalid addresses for both tdata2 and the address to be compared are the same, though it is not guaranteed?

gfavor: I suggest taking the one extra day to properly resolve this issue and not temporarily sweeping it under the rug.  You're not actually saving time to get through the steps to Freeze approval.  It will still be a change that will need to be reviewed by AR (and in the process you wil need to specifically inform the ARC of changes and probably provide a delta spec).  So please just save everyone's time.

gfavor: I'm not taking a position on this; the rest of the TG should collectively decide.  But I'll suggest that Beeman or someone else, if they desire, create a PR for a non-normative comment.  That will concretely and expeditiously move the ball forward.  Rushing to finish things should not be the primary reason for not doing something.  Do what's right for the community and all the readers and implementers of the Debug spec that are going to come along next year and the many following years.

pdonahue-ventana: If that's going to be a separate PR then is this one ready to wrap up?

pdonahue-ventana (approved):
Modulo the non-normative comment which has been proposed to be a separate PR, I believe that we're all done with this.  Of course, we can create new PRs for any other tweaks that get identified later.

Sdtrig.tex
```diff
@@ -319,6 +319,37 @@ \subsection{Cache Operations}
...
+
+If \RcsrTdataTwo can hold any invalid addresses, then writes of an
+invalid address that can not be represented as-is should be converted to
+a different invalid address that can be represented.
+
+For invalid instruction fetch and load and store effective addresses,
+the compare value may be changed to a different invalid address.
+
+\begin{commentary}
+    A straightforward trigger implementation would compare the instruction fetch
```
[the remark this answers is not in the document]
gfavor: Has this non-normative comment been removed (about "straightforward trigger implementations")?  The file changes doesn't show this comment as having been removed, but also doesn't seem to show that it is still present.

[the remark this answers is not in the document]
pdonahue-ventana: I believe that it's removed.  I don't see that sentence in the latest: https://github.com/riscv/riscv-debug-spec/pull/911/files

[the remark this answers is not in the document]
gfavor: SGTM.  Thanks.

Issue #915, opened 2023-11-29
Clarification on updating the uncertain bit


larryrong:
Hello, 

In the debug spec - ch.5.6.12 Match Control Type 6
...
uncertain -> This bit should be updated every time that hit0 or hit1 is updated.
...

The uncertain bit will be updated when one of the hit bits is updated. In this scenario, the debug specification permits us not to implement hit bits, and, consequently, hit bits are never asserted. Does this make sense? I believe the condition for updating the uncertain bit should be 'The TM updates this field when the trigger fires.'

Best regards,
Larry

timsifive:
Thanks for pointing that out.

PR #916, opened 2023-11-29
uncertain is updated when trigger fires.

timsifive:
Don't make this contingent on hit0/hit1 being changed, since those bits are optional. (Although I discourage anyone from not implementing hit0/hit1.)

Addresses #915.


pdonahue-ventana (approved):
Yes, this is better.

PR #919, opened 2023-12-04
Don't allow tdata2 to be smaller for address matches.

timsifive:
This led to endless discussion about what happens when matching invalid addresses. This is on the critical path for Architectural Review approval, which is more important than allowing an implementation to save a few gates on triggers that match on addresses but not on data.


timsifive:
This is an alternative to #911.

pdonahue-ventana (commented):
This seems like a substantive change.  Previously, you could write 0x0123456789ABCDEF (an invalid address) and the WARL logic might modify the upper bits to some other invalid address.  This seems to say that you can't do that (or it no longer says that you can do that).

timsifive:
Yes. I just wanted to throw the option out there.

pdonahue-ventana:
It looks like #911 is moving ahead so can we just close this one?

timsifive:
Yup. Thanks for the reminder.

Issue #913, opened 2023-11-23
Behaviour of triggers after hart reset


kr-sc:
Recently I faced with an OpenOCD issue: wp/bp are not cleared after `reset` command. I tried to investigate is this behaviour correct or not. In my opinion, if hardware triggers are (not) cleared after reset, internal OpenOCD wp/bp should be (not) cleared too. I tried to find some information about it in debug-spec, but I didn't find anything. So, my question is: should hardware triggers be cleared after reset command?

kr-sc:
Note: I faced with question when I tried to investigate an OpenOCD issue: wp/bp are not cleared after reset command. I tried to investigate is this behaviour correct or not. In my opinion, if hardware triggers are (not) cleared after reset, internal OpenOCD wp/bp should be (not) cleared too.

pdonahue-ventana:
What reset are you talking about?  Did OpenOCD do hartreset?  Did OpenOCD do ndmreset?  Did someone else reset the hart?  I'm trying to understand if the debug spec needs a clarification or if this is an OpenOCD problem that should be tracked at https://github.com/riscv/riscv-openocd.

aap-sc:
@pdonahue-ventana this is about what happens with the triggers after `ndmreset` or `hartreset` . We've tried to find to find this information in `Sdtrig` and `Sdext` sectiosn of the latest spec and could not find any relevant information

aap-sc:
> That clearly includes the trigger module for each hart. If reset is applied to the TM then you'll return to the reset values specified in Sdext and Sdtrig.

The thing is that Sdtrig does not  seem to **define** the state of TM after reset. Whatever values `tinfo`, `tdataX` after reset are irrelevant since they begin to make sense only after some valid value is written to tselect. So the question boils down to: "what state does TM have after the reset?"

pdonahue-ventana:
I thought that the question was about which reset domain the TM is in.  It seems that the question is about what the reset values are.

> The thing is that Sdtrig does not seem to define the state of TM after reset.

I don't understand.  Are you looking at https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-stable.pdf?  There are "Reset" columns in the various tables and those reset values come from [the XML](https://github.com/riscv/riscv-debug-spec/blob/master/xml/hwbp_registers.xml).  Can you give a more specific example of something that you don't think is specified?

[...]

PR #920, opened 2023-12-05
CSR reset values apply to underlying triggers.

timsifive:
Make it clear that on reset, the triggers are also reset. The defined values already ensure that triggers are disabled when they have their reset values.

Last change for #913.


pdonahue-ventana (approved):
This was probably what was causing confusion with respect to the reset question.

PR #922, opened 2023-12-07
tools: Remove unnecessary parens.

timsifive:
This makes it easier to merge the code into picky code bases.


pdonahue-ventana (approved):
OK, if that's what you want.  I personally would add even more parentheses around all dissimilar operators so I don't have to waste brain power remembering if `a == b || c == d` means `(a==b)||(c==d)` vs. `(a==(b||c))==d`.

You could also remove parens around `(list.field.values && list.field.values[0])` if you want to really optimize.

timsifive:
It's just to make the OpenOCD style check happy. It seemed easier to make it conform here than it is to edit it or bypass the style check when we integrate this into OpenOCD.

PR #931, opened 2023-12-20
Also run verify in PR relative to asciidoc.

timsifive:


timsifive:
Actually, they are running already.

PR #932, opened 2023-12-20
Fix includes, and remove autogenerated files from github.

timsifive:
Includes of auto-generated files must come from the build/ directory. Otherwise there are two sets of the same files (with potential differences), leading to confusion.

The files can be generated with `make build-registers` if necessary.


wmat:
If the register files are regenerated, won't this overwrite the .adoc files requiring that they be fixed again?

timsifive:
Any fixing should be done either in the autogen code, or in the XML files that are the input to the process.

Another major advantage of the XML file that I haven't mentioned yet is that we can auto-generate C header files from the same XML files, so the C header files will always exactly match the documentation.

Issue #944, opened 2024-01-04
Contradictory requirements for dmcontrol.ndmreset=1


JanMatCodasip:
It appears that the spec contains contradictory requirements for the behavior of `dmcontrol.ndmreset` and `dmstatus.ndmresetpending` bits.

When `dmcontrol.ndmreset==1` the spec says:

![image](https://github.com/riscv/riscv-debug-spec/assets/50193733/d37c3c74-ae51-4608-a88f-1eb2bf3c9e6a)

- Left: When `ndmreset` is asserted, read of `dmstatus` is undefined.
- Right: When `ndmreset=1`, `dmstatus.ndmresetpending`, if implemented, should be equal to 1, which suggests that read of `dmstatus` is a valid operation at that time (contradiction with the above).

To remove the contradiction, we propose to change the left-hand statement to: `While ndmreset [..] is asserted, the only supported DM operations are reading/writing dmcontrol and reading dmstatus.ndmresetpending. The behavior of other accesses is undefined.`

Would that be acceptable? If so, I can create the corresponding pull request.

Thank you.

rtwfroody:
This is clearly a contradiction that should be addressed.

The left side is very old: 7 years. The right side was added in https://github.com/riscv/riscv-debug-spec/pull/594.

I believe the intent of the left side was to make it clear that you shouldn't try to access non-debug-module stuff while ndmreset is asserted. E.g. don't execute any abstract commands. @JanMatCodasip's new language sounds good to me. We'll have to run this change by architecture review, so getting it in the spec will be slower than the previous process.

JanMatCodasip:
@rtwfroody, could you please refer me to the description of the architecture review process.

What can I do to help the process?

Thank you.

rtwfroody:
I don't know of a description of the process after the original review happened. If you can make a PR, that would be helpful. @pdonahue-ventana and I will review it. Once we think it looks good, I'll email ARC and ask for their blessing.

pdonahue-ventana:
I'm OK with the proposed change.

[...]

PR #951, opened 2024-01-11
Fix contradictory req. for ndmresetpending

JanMatCodasip:
Definition of `dmstatus.ndmresetpending` bit says that this bit should be equal to 1 while `dmcontrol.ndmreset` is asserted, implying that `dmstatus.ndmresetpending` is readable even under ndmreset.

Therefore, this merge request updates the text that specifies which DM register accesses are defined under ndmreset.

The change has been discussed here: https://github.com/riscv/riscv-debug-spec/issues/944

### DIFF
--- a/debug_module.adoc
+++ b/debug_module.adoc
@@ -100,7 +100,10 @@ The Debug Module's own state and registers should only be reset at
 power-up and while {dmcontrol-dmactive} in {dm-dmcontrol} is 0. If there is another mechanism to reset the DM, this mechanism must also reset all the harts accessible to the DM.
 
 Due to clock and power domain crossing issues, it might not be possible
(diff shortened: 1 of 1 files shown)


rtwfroody:
The asciidoc merge caused conflicts. I rebased the branch onto it. @JanMatCodasip can you confirm that this is still the same change?

pdonahue-ventana:
@JanMatCodasip: Can you confirm that this PR is still what you were proposing?  If so, I think that we can merge this PR that came from the public review process.

JanMatCodasip:
@rtwfroody, @pdonahue-ventana I am sorry for my delayed reply. 

Yes, the current change is all right and corresponds to the original proposal.

PR #962, opened 2024-02-01
Increase font size in tables and register diagrams.

rtwfroody:


wmat:
Looks like the build is failing for some reason. I'm out right now but can check when I get home.

wmat:
Fixed verify by adding apt update to the runner.

PR #968, opened 2024-02-13
Dual-license XML files BSD-2-Clause and CC-BY-4.0

rtwfroody:
BSD-2-Clause is used so the C code generated from those files can be used in GPLv2 open source projects (specifically, OpenOCD) and the CC-BY-4.0 clause is the standard license for work produced by the RISC-V Foundation.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,10 @@
+All files in this repo may be used under the CC-BY-4.0 license, included below.
+Some files may also be used under a different license. Where that is the case,
+they contain a comment near the top of the file specifying the other license.
+Use under any other license is prohibited.
+
+CC-BY-4.0 license:
+
 Attribution 4.0 International
 
 =======================================================================
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -1,3 +1,15 @@
+<!--
+This file is dual-licensed. You may choose to use this file under the terms of either the:
+
+1. SPDX-License-Identifier: BSD-2-Clause
+
+OR
+
+2. SPDX-License-Identifier: CC-BY-4.0
+
+Note: This dual licensing does not apply to other files that may be part of the same project unless stated otherwise.
+-->
+
 <registers name="Abstract Commands" skip_index="1" skip_access="1"
         skip_reset="1" prefix="AC_" depth="2">
     <register name="Access Register">
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -1,3 +1,15 @@
+<!--
+This file is dual-licensed. You may choose to use this file under the terms of either the:
+
+1. SPDX-License-Identifier: BSD-2-Clause
+
+OR
+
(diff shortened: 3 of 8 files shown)


rpsene (approved):
@rtwfroody LGTM.

rtwfroody:
@borneoa does this work for you?

borneoa:
Perfect! Thanks!

rtwfroody:
> * the correct SPDX tag should be `SPDX-License-Identifier: CC-BY-4.0 OR BSD-2-Clause` in a single line, or accordingly to https://wiki.spdx.org/view/SPDX_FAQ it should be `SPDX-License-Identifier: (CC-BY-4.0 or BSD-2-Clause)`

Fixed.

> * the SPDX tag should be in the very first line of the file, to help automatic tools to build the SW BOM

Fixed.

> * the preferred comment style for the SPDX is `/* comment */` for files.h, while is `// comment` for files.c. See https://www.kernel.org/doc/html/v4.18/process/license-rules.html#license-identifier-syntax

I've left this as-is. /* */ works in every C compiler. // does not (although you might have to go back a way to find a compiler that doesn't accept //).

pdonahue-ventana (approved):
`//` comments were standardized in regular C in 1999, but using old-school comments is fine and will make C89 people happy.

I was going to advise against using Python f-strings for a similar reason since that requires Python 3.6.  However, I see that registers.py already uses f-strings extensively (which nobody seems to be complaining about) and 3.6 is now seven years old.  So that's fine.  And f-strings are so much nicer than the old clunky ways.

build/Makefile
```diff
@@ -70,14 +70,9 @@ build-registers:	$(REGISTERS_ADOC)
 	../registers.py --adoc $@ --adoc-definitions $(patsubst %.adoc,%-def.adoc,$@) $<
 
 debug_defines: debug_defines.h debug_defines.c $(patsubst %,../xml/%,$(REGISTERS_ADOC:.adoc=.xml)) ../registers.py
-	../registers.py $@ --cgetters $(filter %.xml, $^)
-
-debug_defines.%:
```
borneoa, line 75: By removing this make's target `debug_defines.%` the build `make debug_defines` fails on clean build, e.g. after a clean `git clone`
You need to drop `debug_defines.h` and `debug_defines.c` at line 74

borneoa:
If I remember correctly, the SPDX spec was requiring to use a "single line comment" so the decision was to use `// comment` style. But this was not compatible with `files.h` included by assembly files.
At the end, `// comment` was kept for `files.c` only, while the backward compatible `/* comment */` was used in `files.h`.
Anyway, not a big issue; let's keep this simple.

The result is now good for me.

There is one new issue with `build/Makefile`, commented in the patch review

build/Makefile
rtwfroody: Good point. Fixed.

borneoa:
I'm fine with it. Thanks

PR #969, opened 2024-02-14
Propagate license to generated files

borneoa:
On top of the existing pull request https://github.com/riscv/riscv-debug-spec/pull/968
I propose this additional patch to propagate the new license to the generated files by adding the proper SPDX tag.


rtwfroody:
I like propagating the licenses, but this is just hard-coding the license in yet another place. I think I'll modify registers.py to look for any SPDX tag in the source, and to put that in the generated file.

rtwfroody:
#968 was updated to include a similar feature.

Issue #966, opened 2024-02-09
asciidoc lacks lists of figures and tables


rtwfroody:
Probably fixable with https://github.com/Alwinator/asciidoctor-lists

rtwfroody:
In https://lists.riscv.org/g/tech-debug/message/1597, Bill says:
> AsciiDoc does not natively have this feature, however there is a very new extension that provides it.  I have it on my todo list to investigate using it.  I do know that it's not yet bundled as a rubygem, which would make like much easier for our AsciiDoc toolchain.

PR #971, opened 2024-02-15
Convert to asciidoc

rtwfroody:
Convert from LaTeX to asciidoc, to be consistent with other RISC-V specifications.

The only notable regression is that there is no longer a list of tables and figures. There is no good asciidoc support for this.
https://github.com/riscv/riscv-debug-spec/issues/966 tracks this.


pdonahue-ventana (commented):
Other specs changed the branch name from master to main as part of this process.  Is that something that we're supposed to also do as part of the inclusive naming policy?  Either way, we can also do what riscv-isa-manual did and also create a branch named latex which is the final latex version (for easy reference if we ever need to look at that).

@wmat probably knows something about what we're supposed to do.

rtwfroody:
Changing branch names seems like a completely separate change that has nothing to do with asciidoc.

Makes sense to make a latex branch if people find that helpful. I'll do that after this merges.

wmat:
Actually, when I "flipped the switch" to asciidoc on riscv0isa-manual I did change the master branch to main.  This makes the repo comply with inclusive language rules.

pdonahue-ventana:
I agree that the branch name is orthogonal but now is a convenient time, no?  It seems like we could effectively rename merge_asciidoc to be main and rename master to be latex.

wmat:
I'd suggest renaming master to latex, then renaming asciidoc to main.  That way you have the latex branch as a separate branch.

rtwfroody:
I definitely want a regular merge to happen in the interest of keeping the git history consistent, and to have the tools automatically enforce that we didn't miss merging something.

So:
1. Merge this change.
2. Rename master to main. 
3. Make latex branch from just before this merge.

pdonahue-ventana (approved):
Yeah, that plan is OK with me.

I can't say that I scrutinized this enormous change.  I just have faith that this stuff matches the PDF you sent out.

wmat:
That's better than my plan :)

rtwfroody:
I made the latex branch.

PR #973, opened 2024-02-23
Fix `hartset` -> {hartsel}

### LINKED ISSUE #972: Minor asciidoc conversion problem
There's a problem on this line: https://github.com/riscv/riscv-debug-spec/blob/cc2b4600b7bef2e1716b77b4f6acc66d3e83a5d2/debug_module.adoc?plain=1#L368

"hartset" should be "hartsel" (as it was in the LaTeX version).

I would just make a PR myself but I'm not familiar enough with asciidoc (yet) to know how to do the hyperlink (since it's not simply a link to a register field).

rtwfroody:
Fixes #972


pdonahue-ventana (approved):
I guess that was pretty easy.  I don't really know asciidoc so I looked around and thought that it might have to be something weird like `{dmcontrol-hartsello, hartsel}`.

PR #978, opened 2024-03-04
Clarify textra behavior when XLEN changes.

rtwfroody:


pdonahue-ventana (commented):
This seems reasonable.

>  Writes to this register will clear the high bits of the corresponding fields in {csr-textra64}.

If you do something like `csrrsi x0, textra32, 1` then it will set the LSB.  Are csrrs, csrrc, csrrsi, csrrci considered to be writes to this register?  My inclination is that they are writes because the 32b mode is basically claiming ownership and it might expect that csrrs of value X followed by csrrc of ~X should be equivalent to csrrw of X.  It would be good to explicitly say that the set/clear operations are writes.

PR #981, opened 2024-03-08
Make the data of the *context registers 32 bits wide.

rtwfroody:
Still todo, is figure out what to call XLEN there. In the introduction we mention what XLEN means for Debug and M-Mode, but some of these registers are accessible in other modes also.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -277,14 +277,15 @@ same project unless stated otherwise.
         `hstateenzero[57]` in the Smstateen extension.  Enabling {csr-scontext}
         can be a security risk in a virtualized system with a hypervisor that
         does not swap {csr-scontext}.
-        <field name="data" bits="XLEN-1:0" access="WARL" reset="0">
+        <field name="0" bits="XLEN-1:32" access="R" reset="0" />
+        <field name="data" bits="31:0" access="WARL" reset="0">
             Supervisor mode software can write a context number to this
             register, which can be used to set triggers that only fire in that
             specific context.
 
             An implementation may tie any number of high bits in this field to
-            0. It's recommended to implement no more than 16 bits on RV32, and
-            34 on RV64.
+            0. It's recommended to implement 16 bits on RV32 and 32 bits on
+            RV64.
         </field>
     </register>
 
@@ -300,17 +301,18 @@ same project unless stated otherwise.
             hypervisor directly.
         ====
 
-        <field name="hcontext" bits="XLEN-1:0" access="WARL" reset="0">
+        <field name="0" bits="XLEN-1:14" access="R" reset="0" />
+        <field name="hcontext" bits="13:0" access="WARL" reset="0">
             M-Mode or HS-Mode (using {csr-hcontext}) software can write a context
             number to this register, which can be used to set triggers that only
             fire in that specific context.
 
             An implementation may tie any number of upper bits in this field to
             0. If the H extension is not implemented, it's recommended to implement
-            no more than 6 bits on RV32 and 13 on RV64 (as visible through the
+            6 bits on RV32 and 13 bits on RV64 (as visible through the
             {csr-mcontext} register).  If the H extension is implemented,
-            it's recommended to implement no more than 7 bits on RV32
-            and 14 on RV64.
+            it's recommended to implement 7 bits on RV32
+            and 14 bits on RV64.
         </field>
     </register>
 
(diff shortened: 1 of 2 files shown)


xml/hwbp_registers.xml
```diff
@@ -13,9 +13,10 @@ same project unless stated otherwise.
     and Debug Mode to prevent untrusted user code from causing entry into Debug
     Mode without the OS's permission.
 
-    In this section XLEN means MXLEN when in M-mode, and DXLEN when in Debug
-    Mode. On systems where those values of XLEN can differ, this is handled
-    as follows.
+    In this section XLEN refers to the effective XLEN in the current execution
+    mode. This is DXLEN when in Debug Mode.
+    On systems where those values of XLEN can differ, this is handled as
+    follows.
```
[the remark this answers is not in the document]
rtwfroody: I kept DXLEN in there to just make it extra clear than when the hart is halted DXLEN applies. But you're right it's not necessary, and I guess it is odd to only mention that one and not any other modes. I'll simplify as you suggest.

xml/hwbp_registers.xml
```diff
@@ -300,7 +302,8 @@ same project unless stated otherwise.
             hypervisor directly.
         ====
 
-        <field name="hcontext" bits="XLEN-1:0" access="WARL" reset="0">
+        <field name="0" bits="XLEN-1:14" access="R" reset="0" />
+        <field name="hcontext" bits="13:0" access="WARL" reset="0">
```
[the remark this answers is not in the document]
rtwfroody: I wanted to change it to 14 bits because there is no value in implementing additional bits. None of our triggers look at bits higher than bit 13 in this register. Or did I miss something?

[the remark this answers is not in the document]
pdonahue-ventana: You're right.  I forgot that the width of mhvalue basically makes it useless to have anything wider than 14 bits.  I think that we should change the final sentence of the mcontext register's hcontext field to be:
> If the H extension is implemented, it’s recommended to implement no more than 7 bits on RV32.

Because it would now be impossible to implement more than 14 on RV64.

xml/hwbp_registers.xml
```diff
@@ -277,14 +277,15 @@ same project unless stated otherwise.
...
+        <field name="0" bits="XLEN-1:32" access="R" reset="0" />
+        <field name="data" bits="31:0" access="WARL" reset="0">
             Supervisor mode software can write a context number to this
             register, which can be used to set triggers that only fire in that
             specific context.
 
             An implementation may tie any number of high bits in this field to
             0. It's recommended to implement no more than 16 bits on RV32, and
-            34 on RV64.
+            32 on RV64.
```
pdonahue-ventana, line 288: Now I notice that this recommendation is confusing because it's not possible to implement more than 32 bits.  We can just drop everything after the comma.

rtwfroody: I had read this as:
1. It's recommended to implement no more than 16 bits on RV32.
2. It's recommended to implement 32 bits on RV64.

But I agree it's ambiguous.

Maybe we should scrap the "no more" altogether, and just recommend 16 and 32? It's already clear that fewer are allowed.

E.g.
> It's recommended to implement 16 bits on RV32, and 32 on RV64.

pdonahue-ventana: I'm OK with your proposal but I'd eliminate the comma and then maybe say "bits" to be more symmetric: "It's recommended to implement 16 bits on RV32 and 32 bits on RV64."

xml/hwbp_registers.xml
```diff
@@ -277,14 +277,15 @@ same project unless stated otherwise.
...
+        <field name="data" bits="31:0" access="WARL" reset="0">
             Supervisor mode software can write a context number to this
             register, which can be used to set triggers that only fire in that
             specific context.
 
             An implementation may tie any number of high bits in this field to
-            0. It's recommended to implement no more than 16 bits on RV32, and
-            34 on RV64.
+            0. It's recommended to implement 16 bits on RV32, and 32 bits on
+            RV64.
```
pdonahue-ventana, line 288:
```suggestion
            0. It's recommended to implement 16 bits on RV32 and 32 bits on
            RV64.
```

rtwfroody: Done.

pdonahue-ventana (approved):
Ship it!

PR #982, opened 2024-03-11
Fix asciidoc conversion for DXLEN definition.

rtwfroody:


pdonahue-ventana (approved):
I have a feeling that there will be a lot of these.

PR #984, opened 2024-03-12
Remove Section 4.9: XLEN

rtwfroody:
Section 4.1 already says "Effective XLEN is DXLEN." The statement about communicating XLEN to the user is unnecessary to the spec, and unclear, so we don't need it.


pdonahue-ventana (approved):
Sure.

Issue #985, opened 2024-03-19
Priority of mcontrol/mcontrol6 load data before trigger


en-sc:
There are two entries in table 10 (5.3. Priority)

![image](https://github.com/riscv/riscv-debug-spec/assets/109669442/dc8191e6-3988-4ce2-9d6a-1d34bbdc58f7)

and

![image](https://github.com/riscv/riscv-debug-spec/assets/109669442/a69a8af7-81ec-4c6e-b9fe-b8f0255df985)

AFAIK, both these entries specify the priority for `mcontrol/mcontrol6 load data before` (the second one explitly, the first one as one of the options).
Is this a typo? If not, what is the priority for `mcontrol/mcontrol6 load data before`?

pdonahue-ventana:
Yes, this is not accurately specified.  It has the lowest priority and shouldn't be included with the rest of them in the middle of that table.

Load/store address and store data are known before doing any memory access.  Therefore, they can have a relatively high priority (which is the first thing you show).  Load data cannot possibly be known until the load has been performed.  That must happen after address translation and PMP checks.  So load data before must have the lowest priority.

Nobody probably caught this because load data should (but not must) fire after instead of before.  See table 11.

PR #986, opened 2024-03-19
Address https://github.com/riscv/riscv-debug-spec/issues/985

pdonahue-ventana:
I believe that this bug fix should be addressed before ratification.

/cc @en-sc

### DIFF
--- a/Sdtrig.adoc
+++ b/Sdtrig.adoc
@@ -118,7 +118,7 @@ Instruction address misaligned +
 Environment call +
 Environment break +
 Load/Store/AMO address breakpoint
-.>| mcontrol/mcontrol6 load/store address/data before
+.>| mcontrol/mcontrol6 load/store address before, store data before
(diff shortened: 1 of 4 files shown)


rtwfroody:
Since the spec is frozen, I think we need to run this through architecture review. Can you rebase and submit this PR to be relative to rc2? Then we can include it in the batch of changes I was going to send over there today but haven't yet, so we can still squeeze this in.

pdonahue-ventana:
I just rebased to rc2.

pdonahue-ventana:
Well, that doesn't look right.  Did you want me to rebase to rc2 and have the PR merge into rc2?  I don't know how to change the target branch of the PR.

pdonahue-ventana:
I think that I figured that out but now there are additional changes that we already dealt with in different PRs.

pdonahue-ventana:
OK.  We can close this.

PR #987, opened 2024-03-21
Load data before trigger priority was ambiguous.

### LINKED ISSUE #985: Priority of mcontrol/mcontrol6 load data before trigger

rtwfroody:
Fixes #985

### DIFF
--- a/Sdtrig.adoc
+++ b/Sdtrig.adoc
@@ -118,7 +118,7 @@ Instruction address misaligned +
 Environment call +
 Environment break +
 Load/Store/AMO address breakpoint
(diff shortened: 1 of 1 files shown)


rtwfroody:
This replaces #986, which has some confusing git history.

Issue #993, opened 2024-03-23
Table cell justification/alignment issues


TommyMurphyTM1234:
Minor/cosmetic issue.

In several places the contents of cells in tables are not aligned/justified correctly. A few examples where the contents of cells in column 1 are affected:

![Screenshot_20240323-141535](https://github.com/riscv/riscv-debug-spec/assets/14550604/b8d33c1c-d657-4258-a516-82f585db58ea)

![Screenshot_20240323-141631](https://github.com/riscv/riscv-debug-spec/assets/14550604/7adcbecc-3068-4260-aa7a-8b8ee935bb8a)

![Screenshot_20240323-141716](https://github.com/riscv/riscv-debug-spec/assets/14550604/889e8918-f14c-4e49-bcec-78283bfe0608)

TommyMurphyTM1234:
In some cases the cell connected are split/wrapped unnecessarily:

![Screenshot_20240323-142539](https://github.com/riscv/riscv-debug-spec/assets/14550604/5c13e432-3956-427c-873a-e1243fcf9dde)

PR #996, opened 2024-03-27
Minor fixes to table layout.

rtwfroody:
Addresses #993.


pdonahue-ventana (approved):
I basically just reviewed that this doesn't appear to change anything but formatting.  I didn't look at the PDF to see that the new formatting is sensible, though I assume that you did that.

PR #988, opened 2024-03-21
1.0.0-rc2

rtwfroody:
This includes all feedback from the public review period, as well as a bugfix for #985 which was identified after the public review period ended.

This PR cannot be merged until ARC approves it.

PR #998, opened 2024-04-01
Add support for Smdbltrp.

rtwfroody:
2 changes:
1. Add extcause field, which gives is room for a double trap cause plus a few others in case of future need.
2. Add cetrig bit to let a debugger control behavior when a double trap occurs.

We expect this to be approved together with #988, but I want to keep them separate so we can easily have one approved/merged without the other.

### DIFF
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -78,7 +78,35 @@ same project unless stated otherwise.
             All values are reserved for future versions of this spec, or for use
             by other RISC-V extensions.
         </field>
-        <field name="0" bits="23:18" access="R" reset="0" />
+        <field name="0" bits="23:20" access="R" reset="0" />
+        <field name="cetrig" bits="19" access="R/W" reset="0">
+            This bit is part of ((Smdbltrp)) and only exists when that extension
+            is implemented.
+
+            <value v="0" name="disabled">
+                A hart in a critical error state does not enter Debug Mode but
+                instead asserts the critical-error signal to the platform.
+            </value>
+
+            <value v="1" name="enabled">
+                A hart in a critical error state enters Debug Mode instead of
+                asserting the critical-error signal to the platform. Upon such
+                entry into Debug Mode, the cause field is set to 7, and the
+                extcause field is set to 0, indicating a critical error
+                triggered the Debug Mode entry.  This cause has the highest
+                priority among all reasons for entering Debug Mode. Resuming
+                from Debug Mode following an entry from the critical error state
+                returns the hart to the critical error state.
+            </value>
+
+            [NOTE]
+            ====
+                When {dcsr-cetrig} is enabled, resuming from Debug Mode
+                following an entry due to a critical error will result in an
+                immediate re-entry into Debug Mode due to the critical error.
+            ====
+        </field>
+        <field name="0" bits="18" access="R" reset="0" />
         <field name="ebreakvs" bits="17" access="WARL" reset="0">
             <value v="0" name="exception">
             `ebreak` instructions in VS-mode behave as described in the
--- a/xml/core_registers.xml
+++ b/xml/core_registers.xml
@@ -70,7 +70,19 @@ same project unless stated otherwise.
             available version of this spec.
             </value>
         </field>
-        <field name="0" bits="27:18" access="R" reset="0" />
+        <field name="0" bits="27:21" access="R" reset="0" />
+        <field name="extcause" bits="20:18" access="R" reset="0">
+            When {dcsr-cause} is 7, this optional field contains the value of a
+            more specific halt reason than "other." Otherwise it contains 0.
+
+            <value v="0" name="none">
+            There is no more specific halt reason, probably because the hardware
+            does not implement this field.
+            </value>
+
+            Other values are reserved for future versions of this spec, or for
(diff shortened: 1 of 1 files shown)


rtwfroody:
We can't merge this until ARC also reviews the change.

pdonahue-ventana (approved):
I went back and forth about whether we should proactively reserve more bits.  In the end, I think that this is fine for the foreseeable future and any Debug 1.0 follow-on extensions like double trap or secure debug.  Debug 2.0 (which would have a different tinfo.version) can consider the two cause fields to be a unified 6-bit field with 64 encodings rather than the current 7+8 encodings.

pdonahue-ventana (commented):
I just saw your latest change.  Why 26:24 rather than left-justifying or right-justifying in that reserved area?

rtwfroody:
Ved requested those specific bits, to match something he's already written.
I don't really care which bits so I went with it.

pdonahue-ventana:
@ved-rivos: Why bit 19 instead of 18?

ved-rivos:
@pdonahue-ventana - bit 18 has been allocated for previous ELP - `pelp` - please see https://github.com/riscv/riscv-debug-spec/pull/855

pdonahue-ventana:
OK.  That's from a while ago so I forgot about it.

pdonahue-ventana (approved):
I previously approved this.  But I'll explicitly do it again now that it also contains the extcause change.

xml/core_registers.xml
```diff
@@ -70,7 +70,48 @@ same project unless stated otherwise.
...
             </value>
         </field>
-        <field name="0" bits="27:18" access="R" reset="0" />
+        <field name="0" bits="27" access="R" reset="0" />
+        <field name="extcause" bits="26:24" access="R" reset="0">
+            When {dcsr-cause} is 7, this optional field contains the value of a
+            more specific halt reason than "other." Otherwise it contains 0.
+
+            <value v="0" name="dbltrp">
+                An exception or interrupt occurred while the trap handler was in
```
ved-rivos, line 79:
```suggestion
                The hart entered a critical error state.  Part of the ((Smdbltrp)) extension.
```
Reworded to:
1. Use the critical error term to tie it back to the `cetrig`
2. Such an exception/interrupt may not always lead to a critical error (e.g., at S-mode or VS-mode it does not) and in M-mode it does not if it was not the RNMI handler and RNMI was implemented. So this would avoid needing to put all conditions in here.

xml/core_registers.xml
```diff
@@ -70,7 +70,48 @@ same project unless stated otherwise.
             available version of this spec.
             </value>
         </field>
-        <field name="0" bits="27:18" access="R" reset="0" />
+        <field name="0" bits="27" access="R" reset="0" />
+        <field name="extcause" bits="26:24" access="R" reset="0">
+            When {dcsr-cause} is 7, this optional field contains the value of a
+            more specific halt reason than "other." Otherwise it contains 0.
+
+            <value v="0" name="dbltrp">
```
ved-rivos, line 78: Could we call it `critical error`.

ved-rivos (approved):
Looks good. I left a couple of comments.

xml/core_registers.xml
rtwfroody: OK.

xml/core_registers.xml
rtwfroody: Done.

ved-rivos:
@rtwfroody Thanks. LGTM.

xml/core_registers.xml
```diff
@@ -70,7 +70,48 @@ same project unless stated otherwise.
...
             </value>
         </field>
-        <field name="0" bits="27:18" access="R" reset="0" />
+        <field name="0" bits="27" access="R" reset="0" />
+        <field name="extcause" bits="26:24" access="R" reset="0">
+            When {dcsr-cause} is 7, this optional field contains the value of a
+            more specific halt reason than "other." Otherwise it contains 0.
+
+            <value v="0" name="critical error">
+                The hart entered a critical error state. Part of the
```
ved-rivos, line 79:
```suggestion
                The hart entered a critical error state, as defined in the Smdbltrp extension.
```

xml/core_registers.xml
```diff
@@ -70,7 +70,48 @@ same project unless stated otherwise.
...
+                priority among all reasons for entering Debug Mode. Resuming
+                from Debug Mode following an entry from the critical error state
+                returns the hart to the critical error state.
+            </value>
+
+            [NOTE]
+            ====
+                When {dcsr-cetrig} is enabled, resuming from Debug Mode
+                following an entry due to a critical error will result in an
+                immediate re-entry into Debug Mode due to the critical error.
```
ved-rivos, line 111: The debugger may resume with cetrig set to 0 to allow the platform defined actions on critical-error signal to occur. Other possible actions include initiating a hart or platform reset using the Debug Module reset control.

rtwfroody: Adding this.

ved-rivos: Thanks!

rtwfroody:
Per https://lists.riscv.org/g/tech-chairs/message/1898 this was approved by ARC.

PR #1005, opened 2024-04-15
Add dcsr.cetrig for Sddbltrp

rtwfroody:
Private email from Ved:
> We discussed this PR in the ARC this week and the suggestion is to include all
> of the Sddbltrp into this PR. Here is additional inclusion suggested:
> 
>     dcsr - bit 19, cetrig, WARL
>     Smdbltrp extension adds the cetrig (critical-error-trigger) field to dcsr.
>     When cetrig is set to 1, a hart in a critical error state enters Debug Mode
>     instead of asserting the critical-error signal to the platform. Upon such
>     entry into Debug Mode, the cause field is set to 7, and the extcause field
>     is set to 0, indicating a critical error triggered the Debug Mode entry.
>     This cause has the highest priority among all reasons for entering Debug
>     Mode. Resuming from Debug Mode following an entry from the critical error
>     state returns the hart to the critical error state.
> 
> Note:
>     The implication of this specification is that resuming from Debug Mode,
>     following an entry due to a critical error, will either result in an
>     immediate re-entry into Debug Mode due to the critical error if cetrig is
>     set to 1, or it will cause the critical-error signal to be asserted to the
>     platform.


pdonahue-ventana (commented):
Why bit 19 instead of bit 18?

rtwfroody:
Closing in favor of #998.

PR #1006, opened 2024-04-15
Asciidoc conversion fixes for Sdext

### LINKED ISSUE #1000: Typo/mistake in "Debug Mode" section/mprven
The current spec says:

```
All operations are executed with machine mode privilege, except that
additional Debug Mode CSRs are accessible and in may be ignored
according to {dcsr-mprven}
```

"and in may be ignored according to " does not sound right. Looks like we have a misplaced text here. `mprven` is about overriding priveledge level of memory references from the Debug Mode.

pdonahue-ventana:


rtwfroody (changes requested):
Can you give this a title that is at least minimally descriptive?

pdonahue-ventana:
Done

PR #1007, opened 2024-04-15
Asciidoc conversion fixes for appendix A

### LINKED ISSUE #1004: Clarification of register needed
Hi, 

Apologies if this has been fixed already, I think there is either a typo or some context missing in Appendix 2, A.2. Execution Based, in the second paragraph:

"When taking this jump, is saved to dpc and cause is updated in dcsr. This jump is similar to a trap but it is not architecturally considered a trap, so for instance doesn’t count as a trap for trigger behavior."

Should it be something like:

"When taking this jump, **_register x_** is saved to dpc and cause is updated in dcsr. This jump is similar to a trap but it is not architecturally considered a trap, so for instance doesn’t count as a trap for trigger behavior." 

Not sure which register should be "register x" here but it just reads off to me.

Thanks, 
Ciaran

pdonahue-ventana:


rtwfroody (changes requested):
Looks good, but can you make the title more descriptive?

pdonahue-ventana:
Done.

rtwfroody (approved):

Issue #995, opened 2024-03-27
Q about Table 9 (action encoding)


martinberger:
I have a question about Table 9 (action encoding) in Section 5.2 on Page 59 in 1.0.0-rc2. In the row for the action with encoding 1 we have the sentence

>    *"hardware should clear the action field whenever the action field is  1, the new value of dmode would be 0, and the new value of he action    field would be 1."*

It is unclear because the sentences asks to *"clear"* the action field but also asks to set it to 1. This clearing to 1 should only happen if the action field is 1.

(There is also a typo: *"**he** action field"* should the *"**the** action field"*.)

pdonahue-ventana:
IF the current action field is 1 and hardware is doing a write that would write (1) dmode=0 and (2) action=1 THEN hardware should write action=0 (rather than allowing item 2).

tomaird:
From your answers to previous issues (https://github.com/riscv/riscv-debug-spec/issues/746 and https://github.com/riscv/riscv-debug-spec/issues/749), it seems the intention is that it's impossible to get in the state where `dmode=0` and `action=1`, in which case I think the spec wording is unclear.

I propose this sentence: 

> Since the tdata registers are WARL, hardware should clear the action field whenever the action field is 1, the new value of dmode would be 0, and the new value of he action field would be 1.

Should be changed to this:

> Since the tdata registers are WARL, hardware should clear the action field whenever the new value of dmode would be 0, and the new value of the action field would be 1.

Since it doesn't actually matter what the _current_ value of `action` is, it only matters what the _resulting_ combination of `action` and `dmode` are

pdonahue-ventana:
Yes, it's illegal to have dmode=0 and action=1 and the WARL logic must prevent that situation.  The spec provides a recommendation for one specific situation (see https://github.com/riscv/riscv-debug-spec/pull/501 for some history).

A debugger can attempt to write dmode=0 and action=1 in multiple scenarios:
 1. The previous state had action=1 and dmode=1.
     - The current spec directly addresses this.
     - This might be a buggy external debugger (but not a buggy native debugger).
     - The WARL hardware could either:
       - write dmode=1 and action=1 (dropping the write)
       - write dmode=0 and something fixed like action=0 (which is the current spec's recommendation)
 2. The previous state had action!=1 and dmode=1.
     - This might be a buggy external debugger (but not a buggy native debugger).
     - The WARL hardware could either:
       - write dmode=1 and action=1 (dropping the write to dmode)
[...]

PR #1009, opened 2024-04-24
Simplify WARL description for dmode=0 and action=1.

rtwfroody:
Instead of describing a recommended implementation (which didn't cover all cases), explicitly state the value that is illegal and let hardware choose how to do that.

Addresses #995.

### DIFF
--- a/Sdtrig.adoc
+++ b/Sdtrig.adoc
@@ -57,7 +57,9 @@ Triggers can be configured to take one of several actions when they fire. <<tab:
 ^| Value | Description
 |0 | Raise a breakpoint exception. (Used when software wants to use the trigger module without an external debugger attached.) `xepc` must contain the virtual address of the next instruction that must be executed to preserve the program flow.
 |1 | Enter Debug Mode. {csr-dpc} must contain the virtual address of the next instruction that must be executed to preserve the program flow. +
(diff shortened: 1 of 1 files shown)


Sdtrig.adoc
```diff
@@ -57,7 +57,9 @@ Triggers can be configured to take one of several actions when they fire. <<tab:
 ^| Value | Description
 |0 | Raise a breakpoint exception. (Used when software wants to use the trigger module without an external debugger attached.) `xepc` must contain the virtual address of the next instruction that must be executed to preserve the program flow.
 |1 | Enter Debug Mode. {csr-dpc} must contain the virtual address of the next instruction that must be executed to preserve the program flow. +
-This action is only legal when the trigger's {mcontrol-dmode} is 1. Since the `tdata` registers are *WARL*, hardware should clear the action field whenever the action field is 1, the new value of {mcontrol-dmode} would be 0, and the new value of  he action field would be 1. +
+This action is only legal when the trigger's {mcontrol-dmode} is 1.  Since the
+{csr-tdata1} is WARL, hardware must prevent it from containing {tdata1-dmode}=0
+and action=1. +
```
pdonahue-ventana, line 62:
```suggestion
This action is only legal when the trigger's {mcontrol-dmode} is 1.  Since
{csr-tdata1} is WARL, hardware must prevent it from containing {tdata1-dmode}=0
and action=1. +
```
Or "the tdata1 register" would also be OK.

rtwfroody: Fixed. Thanks.

pdonahue-ventana (approved):

rtwfroody:
This introduces a new "must", but that is OK because it's simply restating something about a field that is only legal in certain cases.

PR #1022, opened 2024-04-26
Restore references to `mhartid` in `debug_module.adoc`

en-sc:
Seems like these references were lost durining convertion from LaTeX. See https://github.com/riscv/riscv-debug-spec/blob/f510a7dd33317d0eee0f26b4fa082cd43a5ac7ea/debug_module.tex#L125-L126


pdonahue-ventana (approved):
Looks good.

Issue #1011, opened 2024-04-24
Timing requirements for a chain of `mcontrol` load address triggers


en-sc:
[5.7.11. Match Control (mcontrol, at 0x7a1)] states:

> A chain of triggers that don’t all have the same timing value will never fire. That means to implement
the suggestions in Table 11, both timings should be supported on load address triggers.

Is this correct? Seems like it's enough to support `timing == before (0)` since [Table 11. Suggested Trigger Timings] recommends `timing == before` for Load Address triggers.

rtwfroody:
To implement the suggested timings, load address triggers need:
1. timing=before, to implement a simple load address before
2. timing=after, to implement load address + data after

Which means that load address triggers need both before and after timing support. Does that make sense?

en-sc:
Oh, I see, I have misunderstood.
So `That means to implement the suggestions in Table 11...` implies **all** suggestions from the table.
Maybe something like:
> That means to implement the suggestions in Table 11, both timings should be supported on a load address trigger if it can be chained with a load data trigger.

would be more clear?

PR #1023, opened 2024-04-26
Clarify timing support suggestion for a chain of mcontrol load triggers

en-sc:
Related to #1011

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -333,7 +333,7 @@ same project unless stated otherwise.
         A chain of triggers that don't all have the same {mcontrol-timing}
         value will never fire. That means to implement the suggestions in
         &lt;&lt;tab:hwbp_timing&gt;&gt;, both timings should be supported on load
-        address triggers.
+        address triggers that can be chained with a load data trigger.
 
         The Privileged Spec says that breakpoint exceptions that occur on
         instruction fetches, loads, or stores update the `tval` CSR


xml/hwbp_registers.xml
```diff
@@ -332,8 +332,8 @@ same project unless stated otherwise.
 
         A chain of triggers that don't all have the same {mcontrol-timing}
         value will never fire. That means to implement the suggestions in
-        &lt;&lt;tab:hwbp_timing&gt;&gt;, both timings should be supported on load
-        address triggers.
+        &lt;&lt;tab:hwbp_timing&gt;&gt;, both timings should be supported on a load
+        address trigge if it can be chained with a load data trigger.
```
rtwfroody, line 336:
```suggestion
        &lt;&lt;tab:hwbp_timing&gt;&gt;, both timings should be supported on load
        address triggers that can be chained with a load data trigger.
```

rtwfroody (changes requested):

xml/hwbp_registers.xml
en-sc: Thanks for the catch! Addressed.

pdonahue-ventana (approved):
Sounds right to me.  If anybody asks: Any hardware that complied with the previous statement will also comply with the new statement, so there's no compatibility problem (and it's only a "should" anyway).

rtwfroody (approved):

PR #1024, opened 2024-04-26
Eliminate redundant statement about AMO's stored value availability

en-sc:
It is covered by
> Whether data store triggers match on AMOs is UNSPECIFIED.

### DIFF
--- a/Sdtrig.adoc
+++ b/Sdtrig.adoc
@@ -228,12 +228,9 @@ them as follows:
 The address is always available to trigger on, although the value loaded
 might not be, depending on the hardware implementation.
 . Each AMO instruction is a store for the write portion of the operation.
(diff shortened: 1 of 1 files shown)


rtwfroody (approved):
This is more concise and just as clear.

pdonahue-ventana (approved):
This looks good.  There's no normative change but it's clearer and the formatting is more consistent (rather than having some things in bullets and some not in bullets).

rtwfroody:
Now that we're frozen, we are still allowed to make "minimal, non-substantive" changes (https://docs.google.com/document/d/1Au3veNdNJQKPq-oiQRKTzdgmM72FDaqZOKeH7sOnG04/edit#heading=h.fjjoh42d729w). This qualifies.

PR #1027, opened 2024-05-02
Don't trademark the trigger module

gameboo:
`(TM)`  seems to show as a trademark in asciidoc (at least in the github renderer). Escaping the opening parenthesis `\(TM)` appears to fix that.


pdonahue-ventana (approved):

rtwfroody:
Thank you!

PR #1032, opened 2024-05-14
Fix a typo in [3.12. Security]

en-sc:


pdonahue-ventana (approved):
To be or not to be?  Not to be in this case.

PR #1034, opened 2024-05-14
[3.14.6. Abstract Control and Status] make `busy` a reference

en-sc:
This makes it clear that the statatement refers to a field of a register, not a name of a value of `abstractcs.cmderr` field.

Moreover, references are used throughout the spec in similar cases.


pdonahue-ventana (approved):
I agree that this minor formatting change does help provide context.

PR #1035, opened 2024-05-14
Fix formatting for `mconfigptr`

en-sc:
Names of registers not defined in this specification are highlighted using backticks throughout the specification.

`mconfigptr` used to be an exception.


pdonahue-ventana (approved):
Thanks.

PR #1036, opened 2024-05-14
Fix references to `sbcs.sbaccess64/128`

en-sc:
LaTeX refernces didn't use numbers and replaced them with text.

Seems like this caused an error during convertion and references to `sbcs.sbaccess64/128` started to refer to `sbcs.sbaccess` field with an addition of `Sixtyfour` and `OneTwentyeight` respectively.


pdonahue-ventana (approved):
Thanks for all these formatting improvements.

PR #1037, opened 2024-05-14
Restore reference to `time`

en-sc:
Seems like these references were lost durining convertion from LaTeX.

See https://github.com/riscv/riscv-debug-spec/blob/f510a7dd33317d0eee0f26b4fa082cd43a5ac7ea/Sdext.tex#L47-L49


pdonahue-ventana (approved):
Now it actually makes sense.  Thanks.

PR #1038, opened 2024-05-14
Replace single quotes with backticks in `mstatus` reference

en-sc:


pdonahue-ventana (approved):
This would match the LaTeX formatting.  Thanks.

PR #1039, opened 2024-05-14
Replace "count" with a reference to 'icount.count'

en-sc:
Without the additional formatting the
> but count must be able to count

part seems a bit strange.


pdonahue-ventana (approved):
This wasn't a link in the LaTeX version but it probably should have been.  I think that this is an improvement.

PR #1040, opened 2024-05-17
Clarify the restriction on `dmcontrol` writes  while `abstractcs.busy` is high

### LINKED ISSUE #1021: How to reset the DM on the first connection on both 0.13 and 1.0 spec versions
* Assume an external debugger (ED) is connecting to a hart for the first time.
* ED wants to reset the DM in order to get it into a predictable state.
* To reset the DM the ED needs to set `dmcontrol.dmactive = 0`.
* However, the spec states [3.7. Abstract Commands]:
> While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}...

The ED does not know whether an abstract command is executing, so it needs to read `dmcontrol` to know the value of `dmcontrol.hartsel` before writing `dmcontrol = dmactive=0 | <the old hartsel value>` (requesting DM reset without changing hartsel).
* Assume the read of `dmcontrol` results in an error.
* Since the ED does not know anything about the state of the DM yet, there are two possibilities:
  1. `dmcontrol.dmactive` is low and [3.14.2. Debug Module Control ]
   > Any accesses to the module may fail.
  2. `dmcontrol.dmactive` is high and it's a legitimate DMI error (e.g. some issue with the connection).
* There are two possible routes:
  1. ED assumes `dmcontrol.dmactive` is low and writes `dmcontrol = <dmcontrol.dmactive=1> | <hartsel=0>` -- may result in the change of `hartel`, so ED can't do that.
  2. ED reads `dmcontrol` again.

Now let's take into account RISC-V Debug Spec v0.13, since the ED does not yet know which version it is (`dtmcs` only distinguishes between 0.11 and (0.13 or 1.0)).
In versions less then 0.13.3, according to [1.2.1.2. Incompatible Changes from 0.13 to 1.0]:
> 3. Bump version to 3. #512 , Require debugger to poll dmactive after lowering it. #566

The HW is not required to eventually return something meaningful when `dmcontroll.dmactive` was set to low. So the ED may wait forever in the loop, reading `dmcontrol` and getting an error on RISC-V Debug Spec v0.13 - compliant HW.

Can this somehow be fixed?

en-sc:
With the clarification, it's obvious that the DM can be reset without knowing the `hartsel`.

Fixes #1021

### DIFF
--- a/debug_module.adoc
+++ b/debug_module.adoc
@@ -369,6 +369,8 @@ of {dm-data0}) or hart (e.g. contents of a register modified by a Program Buffe
 Before starting an abstract command, a debugger must ensure that {dmcontrol-haltreq}, {dmcontrol-resumereq}, and {dmcontrol-ackhavereset} are all 0.
 
 While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq}.
+The exception to this rule is when 0 is written to {dmcontrol-dmactive}.
+In that case the DM is reset, and the other values written to {dm-dmcontrol} are ignored.
 
 If an abstract command does not complete in the expected time and
 appears to be hung, the debugger can try to reset the hart (using {dmcontrol-hartreset} or {dmcontrol-ndmreset}).


debug_module.adoc
```diff
@@ -368,7 +368,8 @@ of {dm-data0}) or hart (e.g. contents of a register modified by a Program Buffe
 
 Before starting an abstract command, a debugger must ensure that {dmcontrol-haltreq}, {dmcontrol-resumereq}, and {dmcontrol-ackhavereset} are all 0.
 
-While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq}.
+While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq},
+unless the DM is reset by the same write to {dm-dmcontrol} ({dmcontrol-dmactive} is lowered).
```
rtwfroody, line 372: This sentence is getting really long.
```suggestion
While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq}.
The exception to this rule is when 0 is written to {dmcontrol-dmactive}. In that case the DM is reset, and the other values written to {dm-dmcontrol} are ignored.
```

rtwfroody (commented):
I like the spirit of this change. I wonder if it also needs a line in the definition of dmactive that makes it clear that when 0 is written to dmactive that all other dmcontrol bits are ignored.

debug_module.adoc
en-sc: Addressed.

rtwfroody:
This is not backwards compatible, and the issue is not severe enough to warrant a backwards-incompatible change.

PR #1046, opened 2024-06-13
DM should ignore some bits while abstract command is executing

### LINKED ISSUE #1021: How to reset the DM on the first connection on both 0.13 and 1.0 spec versions

rtwfroody:
This makes things work better in case the debugger wants to reset but doesn't know hartsel, as mentioned in #1021.

Also the debugger should write dmactive=0 to perform a reset, instead of "may".

### DIFF
--- a/debug_module.adoc
+++ b/debug_module.adoc
@@ -369,11 +369,11 @@ of {dm-data0}) or hart (e.g. contents of a register modified by a Program Buffe
 Before starting an abstract command, a debugger must ensure that {dmcontrol-haltreq}, {dmcontrol-resumereq}, and {dmcontrol-ackhavereset} are all 0.
 
 While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq}.
-The exception to this rule is when the debugger writes 0 to
-{dmcontrol-dmactive}. In that case {hartsel} may be written with a new value
-(which the hardware should ignore).
 The hardware should not rely on this debugger behavior, but should enforce it by
 ignoring writes to these bits while {abstractcs-busy} is high.
+The exception to this rule is when the debugger writes 0 to
+{dmcontrol-dmactive}. In that case the hardware must be unaffected by the new
+value of {hartsel}.
 
 If an abstract command does not complete in the expected time and
 appears to be hung, the debugger can try to reset the hart (using {dmcontrol-hartreset} or {dmcontrol-ndmreset}).
--- a/debug_module.adoc
+++ b/debug_module.adoc
@@ -369,8 +369,11 @@ of {dm-data0}) or hart (e.g. contents of a register modified by a Program Buffe
 Before starting an abstract command, a debugger must ensure that {dmcontrol-haltreq}, {dmcontrol-resumereq}, and {dmcontrol-ackhavereset} are all 0.
 
 While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq}.
-The exception to this rule is when 0 is written to {dmcontrol-dmactive}, in
-which case the DM may ignore all bits other than {dmcontrol-dmactive}.
+The exception to this rule is when the debugger writes 0 to
+{dmcontrol-dmactive}. In that case {hartsel} may be written with a new value
+(which the hardware should ignore).
+The hardware should not rely on this debugger behavior, but enforce it by
+ignoring writes to these bits while {abstractcs-busy} is high.
 
 If an abstract command does not complete in the expected time and
 appears to be hung, the debugger can try to reset the hart (using {dmcontrol-hartreset} or {dmcontrol-ndmreset}).
--- a/debug_module.adoc
+++ b/debug_module.adoc
@@ -372,7 +372,7 @@ While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is 
 The exception to this rule is when the debugger writes 0 to
 {dmcontrol-dmactive}. In that case {hartsel} may be written with a new value
 (which the hardware should ignore).
-The hardware should not rely on this debugger behavior, but enforce it by
+The hardware should not rely on this debugger behavior, but should enforce it by
 ignoring writes to these bits while {abstractcs-busy} is high.
 
 If an abstract command does not complete in the expected time and
--- a/debug_module.adoc
+++ b/debug_module.adoc
@@ -371,9 +371,6 @@ Before starting an abstract command, a debugger must ensure that {dmcontrol-halt
 While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq}.
 The hardware should not rely on this debugger behavior, but should enforce it by
 ignoring writes to these bits while {abstractcs-busy} is high.
-The exception to this rule is when the debugger writes 0 to
-{dmcontrol-dmactive}. In that case the hardware must be unaffected by the new
-value of {hartsel}.
 
 If an abstract command does not complete in the expected time and
 appears to be hung, the debugger can try to reset the hart (using {dmcontrol-hartreset} or {dmcontrol-ndmreset}).
--- a/debug_module.adoc
+++ b/debug_module.adoc
@@ -369,6 +369,8 @@ of {dm-data0}) or hart (e.g. contents of a register modified by a Program Buffe
 Before starting an abstract command, a debugger must ensure that {dmcontrol-haltreq}, {dmcontrol-resumereq}, and {dmcontrol-ackhavereset} are all 0.
 
 While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq}.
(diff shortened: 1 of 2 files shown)


rtwfroody:
@en-sc, can you take a look at this?

xml/dm_registers.xml
```diff
@@ -340,13 +340,17 @@ same project unless stated otherwise.
...
+            During this time, the DM may ignore any register writes.
 
             <value v="0" name="inactive">
             The module's state, including authentication mechanism,
             takes its reset values (the {dmcontrol-dmactive} bit is the only bit which can
             be written to something other than its reset value). Any accesses
             to the module may fail. Specifically, {dmstatus-version} might not return
             correct data.
+
+            When this value is written, the DM may ignore any other bits written
```
en-sc, line 352:
```suggestion
            When this value is written, the DM should ignore any other bits written
```
AFAIU, this should be required from the DM. Otherwise it's unclear how this is an exception to: https://github.com/riscv/riscv-debug-spec/blob/ca36be78dd0ddb0dafa974ce554aff42b796d679/debug_module.adoc?plain=1#L371-L373

rtwfroody: I think this is already consistent. The original text said:
> Don't change hartsel when an abstract command is executing.
This allows a hardware implementation to not have to deal with that confusing case.

The new text says:
> Don't change hartsel when an abstract command is executing, unless dmcontrol is written to 0.
The case for the hardware still isn't confusing because dmactive=0 will cancel the abstract command somehow. Or the hardware may choose to wait until the command completes, in which case it can easily do so by ignoring the change to hartsel.

All this accommodation is to give the hardware implementation options, while allowing the debugger to reset without reading hartsel first.

rtwfroody: But thinking about this more, we need a statement that says it's OK to change hartsel (and maybe relax other constraints) when writing dmcontrol=0. Maybe saying should like you suggested is the simplest way to do that.

rtwfroody:
Rather than allowing bits to be ignored depending on dmactive=0, does it make sense to ignore writes to hartsel, haltreq, resumereq, etc. whenever abstractcs.busy=1? That seems cleaner and more robust overall.

However, if we require this behavior it's not backwards compatible. We can say the hardware should do that, and debuggers should preserve the values, and add a note that when resetting it's sometimes unavoidable for the debugger to preserve them.

Thoughts?

en-sc:
Sorry for taking so long to reply.

> Rather than allowing bits to be ignored depending on dmactive=0, does it make sense to ignore writes to hartsel, haltreq, resumereq, etc. whenever abstractcs.busy=1? That seems cleaner and more robust overall.

This is cleaner.

However, AFAIU, this complicates the HW and I'm not sure it is useful for the SW.

The thing is, after the External Debugger has connected to the DM, it knows the state of `abstractcs.busy`.

If for whatever reason it looses track of that state (`abstractcs.busy` is left high after some operation) (e.g. OpenOCD times out waiting for Abstract Command to complete), currently it should read `abstractcs` to check `busy` before attempting a change of `hartsel` and so on.

After the proposed change, the External Debugger would instead be required to read back the written value (in case of `hartsel`) or wait for the effect (`haltreq`, `resumereq`).

I think it will complicate error reporting . E.g.:

* `abstractcs.busy` is high (e.g. OpenOCD timed out waiting for Abstract Command to complete).
* `haltreq` was written and after a while the effect (`dmstatus.[all|any]halted`) has not been observed.
* The debugger can only report something like "halt request failed, maybe an abstract command was not finished".

Currently the similar situation is processed like so:
* Wait for `abstractcs.busy` to get low. If it does not, a clear and precise error message can be reported.
* Wait for `dmstatus.[all|any]halted`, being sure it can not fail due to `abstractcs.busy`.

rtwfroody:
@pdonahue-ventana can you comment here?

pdonahue-ventana:
What about the dmactive recipe?  You write 0 to dmactive and then poll until you see dmactive=0 on a read.  What happens in between the write and the read with respect to any outstanding abstract command and the hartsel field?

There are two types of implementations:
1. those that cannot tolerate hartsel changing during an abstract command
2. those that can tolerate hartsel changing during an abstract command

If busy=1 prevents hartsel from changing then I guess that things will be fine with either implementation and then once dmactive actually goes 0 then hartsel will get reset to 0.  If busy=1 does not prevent hartsel from changing then type 1 implementations will break, though type 2 implementations will be fine.  So maybe we should at least allow implementations (particularly type 1 implementations) to ignore updates to certain fields when busy=1 (where "certain fields" doesn't include ndmreset).

pdonahue-ventana:
On the other hand, if we ignore hartsel writes when busy=1 then what if I want to reset hart N (where N is not the hart where we're currently doing an abstract command)?  I might want to write hartsel=N, hartreset=1 rather than doing the heavy ndmreset.  Or do we have to wait for the command to complete before doing that?  And if the command doesn't complete then presumably we want to hartreset the hart that was doing the command rather than some other hart (or we could ndmreset).

rtwfroody:
> what if I want to reset hart N (where N is not the hart where we're currently doing an abstract command)? I might want to write hartsel=N, hartreset=1 rather than doing the heavy ndmreset. 

That's a good use case for changing hartsel while an abstract command is executing. Sounds like my "simple" proposal of just ignoring hartsel writes when busy=1 doesn't work.

rtwfroody:
> > what if I want to reset hart N (where N is not the hart where we're currently doing an abstract command)? I might want to write hartsel=N, hartreset=1 rather than doing the heavy ndmreset.
> 
> That's a good use case for changing hartsel while an abstract command is executing. Sounds like my "simple" proposal of just ignoring hartsel writes when busy=1 doesn't work.

As discussed in the meeting, this use case is already disallowed. The spec says:
> While an abstract command is executing ..., a debugger must not change {hartsel}, ...

rtwfroody:
> After the proposed change, the External Debugger would instead be required to read back the written value (in case of `hartsel`) or wait for the effect (`haltreq`, `resumereq`).

I don't think it's required, given that abstractcs.busy cannot spontaneously go high. It only goes high in response to a debugger request.

> I think it will complicate error reporting . E.g.:
> 
> * `abstractcs.busy` is high (e.g. OpenOCD timed out waiting for Abstract Command to complete).
> * `haltreq` was written and after a while the effect (`dmstatus.[all|any]halted`) has not been observed.
> * The debugger can only report something like "halt request failed, maybe an abstract command was not finished".
> 
> Currently the similar situation is processed like so:
> 
> * Wait for `abstractcs.busy` to get low. If it does not, a clear and precise error message can be reported.
> * Wait for `dmstatus.[all|any]halted`, being sure it can not fail due to `abstractcs.busy`.

That will still work. The problem you describe only happens if the debugger asserts haltreq without first confirming that `abstractcs.busy` went low. In fact I would expect debuggers to confirm an abstract command has completed before trying anything different.

pdonahue-ventana:
Since a debugger is not allowed to change hartsel (or haltreq, resumereq, ackhavereset, setresethaltreq, clrresethaltreq) while a command is busy, if we say that hardware may/should ignore changes to these fields while it's busy then there's no backward compatibility problem and no loss of functionality.  We can say "may" along with a non-normative note that it's really a "should" for hartsel on implementations where unexpected hartsel changes will break a currently executing abstract command.  Or we could simply say "should" for all of the fields and all implementations.

(I can imagine that unexpectedly changing hartsel is a problem on some implementations but I can't imagine how setting/clearing resethaltreq would be a problem.  This is just an observation, not a proposal to treat that differently.)

xml/dm_registers.xml
rtwfroody: I think this is resolved more cleanly by saying hardware should ignore changes to hartsel while an abstract command is being executed.

debug_module.adoc
```diff
@@ -369,6 +369,11 @@ of {dm-data0}) or hart (e.g. contents of a register modified by a Program Buffe
 Before starting an abstract command, a debugger must ensure that {dmcontrol-haltreq}, {dmcontrol-resumereq}, and {dmcontrol-ackhavereset} are all 0.
 
 While an abstract command is executing ({abstractcs-busy} in {dm-abstractcs} is high), a debugger must not change {hartsel}, and must not write 1 to {dmcontrol-haltreq}, {dmcontrol-resumereq}, {dmcontrol-ackhavereset}, {dmcontrol-setresethaltreq}, or {dmcontrol-clrresethaltreq}.
+The exception to this rule is when the debugger writes 0 to
+{dmcontrol-dmactive}. In that case {hartsel} may be written with a new value
+(which the hardware should ignore).
+The hardware should not rely on this debugger behavior, but enforce it by
+ignoring writes to these bits while {abstractcs-busy} is high.
```
pdonahue-ventana, line 376:
```suggestion
The hardware should not rely on this debugger behavior, but should enforce it by
ignoring writes to these bits while {abstractcs-busy} is high.
```

pdonahue-ventana (approved):
I'm OK with this clarification/recommendation (which is not a normative change that would delay anything).

rtwfroody:
This PR has morphed into recommending some more robust behavior while abstract commands are executing.

pdonahue-ventana (approved):
This is good.  Non-normative suggestions that are backward-compatible.

PR #1048, opened 2024-06-18
Drop `dmcontrol` check before abstract command

en-sc:
The check was introduced in #324.

AFAIU, after #361 and #383 were merged made the check became obsolete.

IMHO, the check is confusing and redundant, since all the fields in question are `W1`/`WARZ`, so it can be dropped.

### DIFF
--- a/debug_module.adoc
+++ b/debug_module.adoc
@@ -366,8 +366,6 @@ failed, the debugger has to look at the state of the DM (e.g. contents
 of {dm-data0}) or hart (e.g. contents of a register modified by a Program Buffer program) to determine which one failed.
 ====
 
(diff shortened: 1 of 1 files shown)


rtwfroody (approved):
I agree.

haltreq is WARZ so always reads 0. resumereq and ackhavereset are both W1 which should always read 0.

So there's no point in checking that those bits are 0.

pdonahue-ventana (approved):
I also agree.  This is just a clarification and not a normative change.  There is no impact on hardware.  Debuggers that do such a check will continue to work but we shouldn't require debuggers to do this pointless check.

PR #1051, opened 2024-07-03
Update resume from Debug Mode behavior with Smdbltrp

ved-rivos:


pdonahue-ventana (approved):
I see.  MRET/SRET clear MDT so that you won't get a critical error on the next trap from U mode (or whatever).  Resuming should clear it for the same reason.

This is very slightly different than the MRET/SRET requirements in the priv spec because you are able to go back to M mode with MDT=1 (whereas MRET to M would clear MDT).  But we don't want to clear MDT in that case because the debugger needs to be allowed to resume to the critical part of the M mode handler where another trap should be a critical error.  So this all makes sense.

ved-rivos:
Additionally, debug mode can resume to S with `sstatus.SDT=1`  and to VS with `vsstatus.SDT=1`  to allow debugging M, S, and VS mode without trashing their state.

pdonahue-ventana:
It's not entirely clear that the third sentence has the same Ssdbltrp condition as the second sentence.  It really depends on both H and Ssdbltrp because (for obvious reasons) Ssdbltrp only adds the vsstatus bit if H is implemented.

ved-rivos:
I placed the second and third sentence in a separate list item to avoid confusion that sentence 3 is not related to Ssdbltrp. I dont say "if H is implemented" as it is implied by VU mode which is provided only when H is implemented.

pdonahue-ventana:
A separate bullet is perfect.

Sdext.adoc
```diff
@@ -192,6 +192,11 @@ When a hart resumes:
 . The current privilege mode and virtualization mode are changed to that
 specified by {dcsr-prv} and {dcsr-v}.
 . If the new privilege mode is less privileged than M-mode, `MPRV` in `mstatus` is cleared.
+. If the Smdbltrp extension is implemented and the new privilege mode is not M,
+  then the `MDT` bit is set to 0.
+. If the Ssdbltrp extension is implemented and the new privilege mode is U, VS,
+  or VU, then `sstatus.SDT` is also set to 0. Additionally, if it is VU, then
+  `vsstatus.SDT` is also set to 0.
```
rtwfroody, line 199:
```suggestion
. If the Smdbltrp extension is implemented:
.. If the new privilege mode is not M, then the `MDT` bit is set to 0.
.. If the new privilege mode is U, VS, or VU, then `sstatus.SDT` is set to 0.
.. If the new privilege mode is VU, then `vsstatus.SDT` is set to 0.
```

I think this is a little more clear.

rtwfroody (changes requested):

Sdext.adoc
pdonahue-ventana: That's not the same because it confuses Smdbltrp and Ssdbltrp.

rtwfroody: My fault for not reading more closely. Thanks.

Sdext.adoc
```diff
@@ -192,6 +192,11 @@ When a hart resumes:
 . The current privilege mode and virtualization mode are changed to that
 specified by {dcsr-prv} and {dcsr-v}.
 . If the new privilege mode is less privileged than M-mode, `MPRV` in `mstatus` is cleared.
+. If the Smdbltrp extension is implemented and the new privilege mode is not M,
+  then the `MDT` bit is set to 0.
+. If the Ssdbltrp extension is implemented and the new privilege mode is U, VS,
+  or VU, then `sstatus.SDT` is also set to 0. Additionally, if it is VU, then
```
rtwfroody, line 198:
```suggestion
  or VU, then `sstatus.SDT` is set to 0. Additionally, if it is VU, then
```

rtwfroody (changes requested):

Sdext.adoc
ved-rivos: Updated.

ved-rivos:
@pdonahue-ventana @rtwfroody 
I noticed that `cetrig` was marked R/W but should have been WARL. Please take a look.

pdonahue-ventana:
The cetrig change looks good.

rtwfroody (approved):

PR #1055, opened 2024-07-12
Clarify that DMI nop does not affect busy/error.

rtwfroody:
The confusion was raised in the 2024-07-08 meeting.


rtwfroody:
I considered changing "should" to "must" but does that then make it backwards incompatible? I really hope there are no implementations that have "nop" affect these status bits, though. It would break all kinds of things. So maybe should->must would be a bug fix?

pdonahue-ventana (approved):
What you have is good.  I think that "must" is probably better but I worry about going through reviews, delaying ratification, etc.  This isn't a ["stop the presses"](https://en.wikipedia.org/wiki/Stop_press) moment so I'm inclined to go with what's here.

PR #1059, opened 2024-08-13
Fix `sw` instruction assembly in Appendix A.2

tienntr:
`sw 0x400(zero), <gpr>` has incorrect assembly syntax, source register should come first.


pdonahue-ventana (approved):
Good catch.

PR #1061, opened 2024-08-14
`dtmcs.abits` description should reference to `dmi.address`

en-sc:


pdonahue-ventana (approved):
It was also this way in the LaTeX, but I think that your change is correct.

Issue #1062, opened 2024-08-17
Suggested improvements in text of spec


rsnikhil:
Here are some suggestions to improve the text in:

    Version 1.0.0-rc3, Revised 2024-05-15: Frozen
    https://github.com/riscv/riscv-debug-spec/releases/download/1.0.0-rc3/riscv-debug-specification.pdf

* Sec 1.2 Context: Should 'Zicsr' be included in this list?

* Sec 3.7.1.3 Access Memory:
      "... with the exact same memory view and permissions as the selected hart has."

      But a hart often has two views into memory: Instruction view and
      Load/Store/AMO view, which may not be exactly in sync (needing
      FENCE.I).  Suggest that the text clarifies which view it refers to.

* Sec 4.7 Halt:
        "2. prv and v are set to reflect the current privilege mode"

    Suggest change to:
        "2. prv and v are set to reflect the current privilege and virtualization mode"
    (This makes the language symmetric with 4.8,
        "2. The current privilege mode and virtualization mode are
            changed to that specified by prv and v.")

* Apx B.2.6: The examples use the Program Buffer to read MSTATUS/F1
    into s0, then finally use an Abstract command to read s0.

    Why would one do this 2-step process, since abstract commands are
    cabable of reading MSTATUS/F1 directly?

    Also, these examples overwrite s0. Would it make more sense to use
    dscratch0?  Would be good to clarify.

* Apx B.2.8:
        "(depending on and other system configuration")
        Dodgy English?

rsnikhil:
Question re. Sec 3.14.2 Debug Module Control (dmcontrol, at 0x10)

>  On any given write, a debugger may only write 1 to at most one of
>  the following bits: resumereq, hartreset, ackhavereset,
>  setresethaltreq, and clrresethaltreq. The others must be written 0."

What happens if the write violates this?  For abstract commands there
is an abstractcs[cmderr] for reporting errors, but what happens for
this dmcontrol error?

rtwfroody:
> * Sec 1.2 Context: Should 'Zicsr' be included in this list?

Yes. The spec includes features specific to accessing CSRs, which means it's addressing Zicsr. #1064 

> * Sec 3.7.1.3 Access Memory:
>   "... with the exact same memory view and permissions as the selected hart has."
>   ```
>   But a hart often has two views into memory: Instruction view and
>   Load/Store/AMO view, which may not be exactly in sync (needing
[...]

PR #1064, opened 2024-08-21
Mention that the spec addresses Zicsr

rtwfroody:
Also sort the list of extensions alphabetically.

Follow up to #1062

### DIFF
--- a/introduction.adoc
+++ b/introduction.adoc
@@ -94,14 +94,15 @@ in the following extensions:
 . F
 . H
 . Sm1p13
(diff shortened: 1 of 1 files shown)


pdonahue-ventana (approved):
Unlike a bunch of other extensions that debug "supports" but that it doesn't really care about one way or the other (e.g. M - multiply/divide), Zicsr is crucial to being able to access the triggers so it does make sense to list that one.  And sorting alphabetically makes sense.

PR #1065, opened 2024-08-21
Clarify Memory Access acts like data access.

rtwfroody:
I'm not sure if this is necessary. Does RISC-V allow data loads to differ from instruction fetches? For a long time any mention of caches was avoided in all specs.

Inspired by #1062.

### DIFF
--- a/xml/abstract_commands.xml
+++ b/xml/abstract_commands.xml
@@ -179,10 +179,11 @@ same project unless stated otherwise.
 
     <register name="Access Memory">
         This command lets the debugger perform memory accesses,
-            with the exact same memory view and permissions as the selected
-            hart has. This includes access to hart-local memory-mapped
-            registers, etc. The command performs the following sequence of
-            operations:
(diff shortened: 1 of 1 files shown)


pdonahue-ventana:
> Does RISC-V allow data loads to differ from instruction fetches?

Yes.  The Zifencei chapter of the unpriv spec says: "RISC-V does not guarantee that stores to instruction memory will be made visible to instruction fetches on a RISC-V hart until that hart executes a FENCE.I instruction."

pdonahue-ventana (commented):
I would say something more like "This command lets the debugger perform memory accesses, with the exact same memory view and permissions as performing a load/store on the selected hart."  Then that leads nicely into the next sentence about memory-mapped registers.

rtwfroody:
@pdonahue-ventana sounds good. Changed.

pdonahue-ventana (approved):
Great

PR #1067, opened 2024-08-23
Clarify why to use progbuf to access FPRs/CSRs

rtwfroody:
More from #1062


pdonahue-ventana (approved):
This seems very clear now.

PR #1075, opened 2024-10-09
Readme: Fix build instructions

JanMatCodasip:
The Readme still contained old (pre-adoc) build instructions. Fix it.

Mention also the use of submodules which are needed for the documentation to build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,21 @@ You may be looking for one of the following pre-built PDFs:
 Build Instructions
 ------------------
 
-```
-sudo apt-get install git make python3 python3-sympy graphviz texlive-full
+```bash
+# Install docker and python3-sympy, if not installed already.
+
+# Pull the latest RISC-V Docs container image:
+docker pull riscvintl/riscv-docs-base-container-image:latest
+
+git clone https://github.com/riscv/riscv-debug-spec.git
+cd riscv-debug-spec
+
+# Optionally, check out a specific revision:
+# git checkout <rev>
+
+git submodule update --init --recursive
+
+cd build
 make
 ```
 


pdonahue-ventana (commented):
This seems right at first glance.  However, it's different than .github/workflows/verify.yml (which I'm not an expert on).  Any thoughts on that?  And while I'm looking in the .github directory, does .github/workflows/publish.yml need to be updated?  It still has the texlive stuff in it.

JanMatCodasip:
Hello @pdonahue-ventana -

I have checked `.github/workflows/verify.yml` and it is correct and consistent with the updated Readme. Still, I have slightly expanded the build instructions in Readme to provide a little more guidance to the user.

I have also looked at `.github/workflows/publish.yml` and you are correct, this job is definitely outdated. This job [has not been executed](https://github.com/riscv/riscv-debug-spec/actions/workflows/publish.yml) for quite some time because it is only active for git tags, and the current tags do not match the [wildcard in the job definition](https://github.com/riscv/riscv-debug-spec/blob/bcd05b798f3584cc89aebb5c9e30012e65c711a1/.github/workflows/publish.yml#L5). I propose to remove it (in another merge request: https://github.com/riscv/riscv-debug-spec/actions/workflows/publish.yml).

rtwfroody (approved):

JanMatCodasip:
@rtwfroody @pdonahue-ventana Could this PR be merged, please? 

Thank you.

JanMatCodasip:
@pdonahue-ventana Gentle ping, could this PR be merged, please?

rtwfroody (approved):
Thank you!

PR #1085, opened 2024-11-20
Better description of how etrigger matches.

### LINKED ISSUE #1084: etrigger priority
Must all etrigger-firing exceptions have the highest priority, or can a subset have lower priorities? For example, a load page fault etrigger having lower priority than an instruction address mcontrol6 trigger.

pdonahue-ventana:
The previous language talked about "When the trigger matches" but it never described matching.  It said that it could fire on mcause codes but did that mean a CSR write instruction with a particular mcause value could make it fire?  I think that this is better.

Fixes https://github.com/riscv/riscv-debug-spec/issues/1084.

### DIFF
--- a/xml/hwbp_registers.xml
+++ b/xml/hwbp_registers.xml
@@ -1252,11 +1252,13 @@ same project unless stated otherwise.
 
         This register is accessible as {csr-tdata1} when {tdata1-type} is 5.
 
-        This trigger may fire on up to XLEN of the Exception Codes defined in
-        `mcause` (described in the Privileged Spec, with Interrupt=0). Those
-        causes are configured by writing the corresponding bit in {csr-tdata2}.
-        (E.g.  to trap on an illegal instruction, the debugger sets bit 2 in
-        {csr-tdata2}.)
+        This trigger can match when a trap is taken due to an exception.
+
+        The trigger may be configured to match on up to XLEN of the Exception
+        Codes defined in `mcause` (described in the Privileged Spec, with
+        Interrupt=0). Those causes are configured by writing the corresponding
+        bit in {csr-tdata2}. (e.g. to have etrigger match on an illegal
+        instruction trap, the debugger sets bit 2 in {csr-tdata2}.)
 
         [NOTE]
         ====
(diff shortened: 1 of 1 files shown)


xml/hwbp_registers.xml
```diff
@@ -1252,11 +1252,13 @@ same project unless stated otherwise.
 
         This register is accessible as {csr-tdata1} when {tdata1-type} is 5.
 
-        This trigger may fire on up to XLEN of the Exception Codes defined in
-        `mcause` (described in the Privileged Spec, with Interrupt=0). Those
-        causes are configured by writing the corresponding bit in {csr-tdata2}.
-        (E.g.  to trap on an illegal instruction, the debugger sets bit 2 in
-        {csr-tdata2}.)
+        This trigger can match when a trap is taken due to an exception.
```
rtwfroody, line 1255: This feels like a very wish-washy statement.

How about:
This trigger matches when an exception is taken and the bit corresponding to the exception is set in tdata2.

Probably requires rewriting the paragraph below for it all to make sense.

pdonahue-ventana: > This feels like a very wish-washy statement.

I agree.  I didn't like the phrasing but I was trying to match the existing itrigger description which was [updated last year](https://github.com/riscv/riscv-debug-spec/commit/92835a85352d2b5e087ca953c164181367078fe8).  I was trying to limit the scope of this change but maybe we should update itrigger, too.

xml/hwbp_registers.xml
```diff
@@ -1252,11 +1252,13 @@ same project unless stated otherwise.
...
+
+        The trigger may be configured to match on up to XLEN of the Exception
+        Codes defined in `mcause` (described in the Privileged Spec, with
+        Interrupt=0). Those causes are configured by writing the corresponding
+        bit in {csr-tdata2}. (e.g. to have etrigger match on an illegal
+        instruction trap, the debugger sets bit 2 in {csr-tdata2}.)
```
rtwfroody, line 1261: This still doesn't say that the trigger matches when an exception is taken, which is what your change description says you wanted to accomplish.

rtwfroody (changes requested):

pdonahue-ventana:
Now that this is ratified, I think that any changes need to be reviewed by Ved.  The two of us can continue to polish this.  Then I think that we can gather up any pending clarifications over some period of time and send them all in a batch to him.

PR #1092, opened 2024-12-10
add dependabot to create PR to update the submodules to latest

kbroch-rivosinc:
relates to https://github.com/riscv-admin/docs-sig/issues/63


rtwfroody:
What does dependabot do? Why would we want it?

kbroch-rivosinc:
> What does dependabot do? Why would we want it?

It runs daily and checks if submodule dependencies in this repo could be updated.  If they can, it provides a PR.

rtwfroody:
> It runs daily and checks if submodule dependencies in this repo could be updated. If they can, it provides a PR.

Sounds useful. Thanks.

rtwfroody (approved):

PR #1099, opened 2025-01-08
registers.py: Fix subtractions in generated C code

JanMatCodasip:
The registers.py generator outputs expressions with subtractions into the C code of debug_defines.{h,c} in this form:

```
((XLEN) + -6ULL)
```

Such code may look awkward to code readers and also triggers GCC's `-Wconversion` warnings in the projects that will include these generated C files.

Fix the generator to produce expressions with subtraction in a natural form:

```
((XLEN) - 6ULL)
```

### DIFF
--- a/registers.py
+++ b/registers.py
@@ -472,7 +472,22 @@ def sympy_to_c(expression, sym_to_c = lambda s: f"({s})", unsigned=True):
     elif isinstance(expression, sympy.Symbol):
         return sym_to_c(expression)
     elif isinstance(expression, sympy.Add):
-        return "(" + " + ".join(stc(t) for t in reversed(expression.args)) + ")"
+        args = list(reversed(expression.args))
+        result = ""
+        for i in range(len(args)):
+            arg = args[i]
+            is_first = (i == 0)
+            if is_first:
+                result += stc(arg)
+            else:
+                # Simplify additions of negative constants:
+                # Use (a - 1) instead of (a + -1)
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
For an easier review, attached are the generated files before and after this change:

[generated_debug_defines.zip](https://github.com/user-attachments/files/18352602/generated_debug_defines.zip)

JanMatCodasip:
@en-sc Would you please have a moment for this review? Thank you.

registers.py
```diff
@@ -472,7 +472,22 @@ def sympy_to_c(expression, sym_to_c = lambda s: f"({s})", unsigned=True):
...
+            if is_first:
+                result += stc(arg)
+            else:
+                # Simplify additions of negative constants:
+                # Use (a - 1) instead of (a + -1)
+                negative_constant = arg.is_constant() and (arg < 0)
+                result += " - " if negative_constant else " + "
+                if negative_constant:
+                    arg = -arg  # flip to positive
+                result += stc(arg)
```
rtwfroody, line 489: This seems simpler to read:
```suggestion
                if arg.is_constant() and (arg < 0):
                    result += " - %s" % stc(-arg)
                else:
                    result += " + %s" % stc(arg)
```

rtwfroody (changes requested):

registers.py
JanMatCodasip: I agree, that is much cleaner. I have adopted the suggestion. Thanks.

PR #1100, opened 2025-01-20
Fix the link to `dmstatus.version` in "Minor Changes from 0.13 to 1.0"

en-sc:
The link got broken during LaTeX to AsciiDoc conversion. Link: https://github.com/riscv/riscv-debug-spec/blob/f510a7dd33317d0eee0f26b4fa082cd43a5ac7ea/introduction.tex#L172


pdonahue-ventana (approved):
Thanks.  This was an asciidoc conversion problem.  This PR makes it match the original latex version.

PR #1103, opened 2025-01-29
`dmactive` description should allow system-wide reset

en-sc:
[3.2. Reset Control] allows for a system-wide reset to reset the Debug Module.
Link: https://github.com/riscv/riscv-debug-spec/blob/a4678108534bef5b4fbf51e92acacd08b28c17af/debug_module.adoc?plain=1#L100

However, `dmactive` field description used to prohibit **any** other DM reset mechanism.

The commit fixes this inconsistency.

### DIFF
--- a/xml/dm_registers.xml
+++ b/xml/dm_registers.xml
@@ -374,7 +374,8 @@ same project unless stated otherwise.
             </value>
 
             No other mechanism should exist that may result in resetting the
-            Debug Module after power up.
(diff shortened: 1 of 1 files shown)


rtwfroody:
I don't think this is inconsistent. There *should* not exist a separate mechanism that resets the DM, but if there is then it must also reset all the harts.

It might be worth mentioning that caveat about resetting all the harts in the dmactive description, but we shouldn't change the language away from "this should not exist."

PR #1122, opened 2025-05-07
Remove typedef in generated 'debug_defines.h'

borneoa:
The pull request #858 introduces four typedef for some struct in the generated 'debug_defines.h'
Using typedef is not welcome by every project that should include these generated files.
For example, OpenOCD coding style requires not adding new typedef.

This series:
- replaces the typedefs in 'debug_defines.h' with simple 'struct struct_name';
- still defines the old typedef, so existing code will still compile without changes;
- modify the example code in 'debug_reg_printer.c' to use the new data types;
- drop one unused parameter in 'debug_reg_printer.c' that triggers a compile warning.

I still have doubts if I should keep the backward compatibility. Please let me know!

### DIFF
--- a/registers.py
+++ b/registers.py
@@ -104,10 +104,14 @@ class Register( object ):
 
     @staticmethod
     def c_field_list_type():
-        return ("typedef struct riscv_debug_reg_field_list_t {\n" +
-                "\triscv_debug_reg_field_info_t field;\n" +
-                "\tstruct riscv_debug_reg_field_list_t (*get_next)(riscv_debug_reg_ctx_t context);\n" +
-                "} riscv_debug_reg_field_list_t;\n")
+        return ("struct riscv_debug_reg_field_list {\n" +
+                "\tstruct riscv_debug_reg_field_info field;\n" +
+                "\tstruct riscv_debug_reg_field_list (*get_next)(struct riscv_debug_reg_ctx context);\n" +
+                "};\n" +
+                "/* deprecated, prefer 'struct riscv_debug_reg_field_list' to 'struct riscv_debug_reg_field_list_t' */\n" +
+                "#define riscv_debug_reg_field_list_t riscv_debug_reg_field_list\n" +
+                "/* deprecated, prefer 'struct riscv_debug_reg_field_list' to 'riscv_debug_reg_field_list_t' */\n" +
+                "typedef struct riscv_debug_reg_field_list riscv_debug_reg_field_list_t;\n\n")
     def c_field_getter_names(self):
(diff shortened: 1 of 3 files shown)


registers.py
```diff
@@ -104,10 +104,14 @@ def sorted_fields( self ):
...
-        return ("typedef struct riscv_debug_reg_field_list_t {\n" +
-                "\triscv_debug_reg_field_info_t field;\n" +
-                "\tstruct riscv_debug_reg_field_list_t (*get_next)(riscv_debug_reg_ctx_t context);\n" +
-                "} riscv_debug_reg_field_list_t;\n")
+        return ("struct riscv_debug_reg_field_list {\n" +
+                "\tstruct riscv_debug_reg_field_info field;\n" +
+                "\tstruct riscv_debug_reg_field_list (*get_next)(struct riscv_debug_reg_ctx context);\n" +
+                "};\n" +
+                "/* deprecated, prefer 'struct riscv_debug_reg_field_list' to 'struct riscv_debug_reg_field_list_t' */\n" +
+                "#define riscv_debug_reg_field_list_t riscv_debug_reg_field_list\n" +
```
en-sc, line 112: I believe it's okay to break this interface from version to version since it's not the part of the spec itself.
@rtwfroody, @pdonahue-ventana, could you please share your opinion?

en-sc (approved):
@borneoa, thank you for the change!

One more question: should this stileguide be documented somewhere?

en-sc:
@pdonahue-ventana, this change does not affect the specification, so perhaps it can be merged? What is the appropriate process for such changes?

pdonahue-ventana:
@ved-rivos: What's the process for handling this PR which affects only autogenerated .h files but not the spec itself?

Does anyone besides OpenOCD rely on these files?

I wonder if this script should be moved out to another project which then refers to this one (as a submodule or via some other mechanism).  Then the script can change at any time without the need to make changes here.

ved-rivos:
@pdonahue-ventana if the .h file and/or the script needs an update to match the ratified spec then we should update to fix that issue.

pdonahue-ventana:
The .h file changes have more to do with C syntax than with the actual debug spec content.  This comment basically summarizes it: `prefer 'struct riscv_debug_reg_info' to 'riscv_debug_reg_info_t'` (i.e. use this new type rather than the old type, though both contain the same fields)

The change does, for example, have `typedef struct riscv_debug_reg_field_info riscv_debug_reg_field_info_t` so it's presumably backward-compatible with code that uses the old name.

PR #1174, opened 2025-11-16
Fix typo in "Reset" section (hasresethaltreq -> halt-on-reset bits)

dokunira:
This PR corrects a typo in the section titled "Reset".

As discussed in the issue, the text incorrectly referred to hasresethaltreq (the feature support bit) instead of the functional "halt-on-reset bits" which actually cause the halt.

This change makes this section consistent with the terminology used in Section titled "Run Control".


dokunira:
Sorry for the noise. Closing this to re-submit from a proper feature branch.

PR #1220, opened 2026-06-05
Norm rule tags for Sdtrig

pclarkhmc:
Added normative rule tags to Sdtrig.adoc and hwbp_registers.adoc


davidharrishmc (commented):
Looks good to me.  No deletions made from the ratified specification.  Normative rule tags look reasonable.

In my opinion, this is safe to merge into the debug spec the same way normative rules were merged into riscv-isa-manual.  When the debug spec moves into riscv-isa-manual, the rules will be ready to go in the normal flow.

